PR #6, opened 2016-07-20
CRAFTERCMS-2001 - Group checkboxes does not render saved values if content-type is Component

vestrada:


components/cstudio-forms/controls/checkbox-group.js
```diff
@@ -212,7 +212,7 @@ YAHOO.extend(CStudioForms.Controls.CheckBoxGroup, CStudioForms.CStudioFormField,
                     groupEl.appendChild(rowEl);
                 }
                 _self.value = newValue;
-                _self.form.updateModel(_self.id, _self.getValue());
+                //_self.form.updateModel(_self.id, _self.getValue());
```
russdanner, line 215: Has this been fully tested.  It looks weird.  Why does this solve the problem?
Also, don't comment out code.  remove it.

components/cstudio-forms/controls/rte.js
```diff
@@ -546,7 +546,7 @@ YAHOO.extend(CStudioForms.Controls.RTE, CStudioForms.CStudioFormField, {
 						ed.contextControl._onChange(null, ed.contextControl);				
 			        });
 			        ed.onChange.add(function(ed, l) {
-			        	ed.contextControl.resizeEditor(ed);
+			        	//ed.contextControl.resizeEditor(ed);
```
russdanner: This looks like it will cause a bug
Also why is it commented out rather than removed?

components/cstudio-forms/controls/checkbox-group.js
vestrada, line 215: Yes, this change was tested by Uber, He asked for the change, so I asked him to check it.
The problem was that this function was called in the init function, so it was changing the correct value to empty value. 
I'm going to remove it.

components/cstudio-forms/controls/rte.js
vestrada: It's ok, This line was causing the issue, it was unnecessary. I test it in all browsers. 
I'm going to remove it.

PR #44, opened 2016-09-22
CRAFTERCMS-2048 - make real dialog with user friendly message when no…

vestrada:
… content types available


components/cstudio-common/common-api.js
```diff
@@ -1443,7 +1443,20 @@ var nodeOpen = false;
                         }
 
                         if (contentTypes.length == 0) {
-                            alert("no content types available for [" + site + ":" + path + "]");
+                            var dialogEl = document.getElementById("errMissingRequirements");
+                            if(!dialogEl){
+                                var dialog = new YAHOO.widget.SimpleDialog("errMissingRequirements",
+                                    { width: "400px",fixedcenter: true, visible: false, draggable: false, close: false, modal: true,
+                                        text: "No content types available for [" + site + ":" + path + "]", icon: YAHOO.widget.SimpleDialog.ICON_BLOCK,
+                                        constraintoviewport: true,
```
russdanner, line 1451: The message should be in a resource file and the text of the message should not look technical  [SITE : PATH] 
Use: "No content types available for content at path 'PATH'

PR #63, opened 2016-10-24
CRAFTERCMS-2168 - In Content Type Page when the user save a content type that doesn't contain a template, the user should be available of continue or cancel the save function.

vestrada:


russdanner (approved):
Reviewed

PR #90, opened 2016-11-02
CRAFTERCMS-2130 - UI includes of resources (js,css & angular templates) should have a cache param

jvega190:
CRAFTERCMS-2130 - UI includes of resources (js,css & angular templates) should have a cache param


russdanner (changes requested):
@jose.vega i dont get the last commit

[4:01]  
+CStudioAuthoring.Utils.addCss("/static-assets/components/cstudio-admin/mods/bulkoperations.css?version=" + CStudioAuthoring.UIBuildId);

[4:02]  
we should not udpate every call to CStudioAuthoring.Utils.addCss

[4:02]  
just go in the method and add the cache buster there

[4:02]  
same with JS

[4:02]  
You are making the developer work rather than letting the API take care of the work

PR #98, opened 2016-11-14
issues/263, issues/217

asotog:
issues/217: flexible rte width, so we can avoid overflow
issues/263: last slash removed from dashboards widgets preview urls

### DIFF
--- a/components/cstudio-forms/controls/rte.js
+++ b/components/cstudio-forms/controls/rte.js
@@ -465,7 +465,7 @@ YAHOO.extend(CStudioForms.Controls.RTE, CStudioForms.CStudioFormField, {
 	        editor_selector : rteUniqueInitClass,
 	        theme : "advanced",
 	        skin : "cstudio-rte",
-	        width : width,
+			width: '100%', // Why? Field width should be flexible, because of responsiveness, should resize based on the screen size, this works together with max width set on post render event below on setup: method 
 	        height: height,
 	        encoding : "xml",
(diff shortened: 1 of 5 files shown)


components/cstudio-forms/controls/rte.js
```diff
@@ -465,7 +465,7 @@ YAHOO.extend(CStudioForms.Controls.RTE, CStudioForms.CStudioFormField, {
 	        editor_selector : rteUniqueInitClass,
 	        theme : "advanced",
 	        skin : "cstudio-rte",
-	        width : width,
+			width: '100%', // Why? Field width should be flexible, because of responsiveness, should resize based on the screen size, this works together with max width set on post render event below on setup: method
```
russdanner, line 468: We cannot do this... width is handled by the form def.... ideally it matches the size of the region so that the wrapping of text is correct

russdanner:
RTE fix is invalid, cannot accept pull request.  See comment on RTE.

components/cstudio-forms/controls/rte.js
asotog: well, width is defined by form definition, but if you said for example 800px, if you resize the window, the field will go out of the form, because width is overflowing the available container space, so fix was to make it 100% but use the definition width for max width

russdanner: OK I see the max width.   I will re-open and merge

russdanner:
Re-opening because I see how the solution works now

PR #108, opened 2016-11-18
https://github.com/craftercms/craftercms/issues/280

asotog:
Date picker fix, so when date is pre populated, it runs the validation, and mark the field as valid from the beginning


cortiz:
@asotog Merging this But I'll need a PR for Master!

PR #173, opened 2017-01-11
Studio: Preview is not reloaded after using the "Change Template" option. #329

vestrada:


russdanner:
See comments in review

PR #208, opened 2017-01-26
UI Bugfixes

asotog:
https://github.com/craftercms/craftercms/issues/426

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -667,6 +667,7 @@ var nodeOpen = false;
 
                         this.on("submitComplete", function(evt, args){
                             //window.location.reload();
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -667,6 +667,7 @@ var nodeOpen = false;
 
                         this.on("submitComplete", function(evt, args){
                             //window.location.reload();
+                            CStudioAuthoring.SelectedContent.selectedContent = []; // clear selected contents after publish
```
russdanner: This really should be common method clear or clean up stuff after actions.
Also CStudioAuthoring.SelectedContent should have a clear method... this code operates directly on the datastructure

asotog: Ready, please check latest added commit

PR #228, opened 2017-02-03
[studio-ui] Add button to sync from repository #438

jvega190:


static-assets/components/cstudio-common/common-api.js
```diff
@@ -3379,6 +3380,27 @@ var nodeOpen = false;
...
+
+            syncFromRepo: function(site, callback){
+                var serviceUrl = this.syncRepoServiceUrl;
+                var postData = "site_id=" + site;
+
+                var serviceCallback = {
+                    success: function(jsonResponse) {
+                        var results = {};
+                        if(jsonResponse.responseText != "") {
+                            results = eval("(" + jsonResponse.responseText + ")");
```
russdanner, line 3392: Is it possible to use JSON.parse?  I see that in some places and eval in others.

jvega190: Done!

PR #234, opened 2017-02-06
https://github.com/craftercms/craftercms/issues/287 - using uri property with some validations instead of property

vestrada:


static-assets/components/cstudio-contextual-nav/wcm_content.js
```diff
@@ -628,7 +628,7 @@ CStudioAuthoring.ContextualNav.WcmActiveContentMod = CStudioAuthoring.Contextual
                                     }
                                     if(CStudioAuthoringContext.isPreview){
                                         try{
-                                            var currentContentTO = (pageParameter.replace("html","xml") == contentTO.item.browserUri) ? null : contentTO.item;
+                                            var currentContentTO = (pageParameter.replace("html","xml") == contentTO.item.uri.replace('/site/website', '').replace("/index.xml", "")) ? null : contentTO.item;
```
russdanner: I understand what this does and I can merge this but please re-submit this this code the following fixes:
- Break this ternary operation down in to more simple statements. It's very hard to understand what is going on.  Instead of simple VALUE = CONDITION ? A : B you have VALUE = (VALUEA.MANIPULATION.MANIPULATION=VALUEB.MANIPULATION.MANIPULATION) ? null : SOMEVALUE.

That's just too difficult to read AND more importantly, I dont understand (as the reader WHY all these things are happening.  Assign the manipulations to names that tell me what the manipulated values are. Then use those values.

PR #312, opened 2017-03-27
Validating site Id on create dialog

vestrada:


sumerjabri:
Please change the acceptance criteria to be:

- lowercase only
- accept '-' and '_'
- accept numbers

(right now it seems you're only accepting mixed-case alpha numeric, I just want to add '-' and '_' and enforce lowercase.)

PR #328, opened 2017-03-31
[studio-ui] Support externally managed users #655

jvega190:


sumerjabri:
Will hold off on accepting this PR until the backend is there too. Keep it open.

PR #535, opened 2017-08-01
[studio-ui] Clicking the back button in studio logs user out #1161

jvega190:


avasquez614:
@jvega-crafter please check why Codacy failed and let me know if you can fix it

PR #540, opened 2017-08-08
[studio-ui] When you click on a component item from search results, it tries to render it #1177

jvega190:


jvega-crafter:
https://github.com/craftercms/studio2-ui/pull/540

PR #558, opened 2017-08-22
Config properties added

avasquez614:
* Configuration properties added for Engine related configurations.


sumerjabri:
@avasquez614 please rename to "Crafter Engine Site Configuration" instead of "Engine Site Configuration"

PR #610, opened 2017-09-27
[studio] Improve CMIS data source error message when cannot connect #1300

jvega190:


russdanner:
* Update link styles and 
* Use glyph for the expand collapse  + / - may work well instead of the  ascii > and < chars
https://www.useloom.com/share/ceebeed11ae643c5a41c48caf2261acf

PR #665, opened 2017-10-30
#1503

vestrada:
https://github.com/craftercms/craftercms/issues/1503 - [studio-ui] its no longer possible to create a template for a new content type #1503

### DIFF
--- a/static-assets/components/cstudio-dialogs/new-template.js
+++ b/static-assets/components/cstudio-dialogs/new-template.js
@@ -41,7 +41,7 @@ CStudioAuthoring.Dialogs.NewTemplate = CStudioAuthoring.Dialogs.NewTemplate || {
 	 */
 	createDialog: function(path) {
 		var CMgs = CStudioAuthoring.Messages;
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-dialogs/new-template.js
```diff
@@ -41,7 +41,7 @@ CStudioAuthoring.Dialogs.NewTemplate = CStudioAuthoring.Dialogs.NewTemplate || {
 	 */
 	createDialog: function(path) {
 		var CMgs = CStudioAuthoring.Messages;
-		var previewLangBundle = CMgs.getBundle("siteDropdown", CStudioAuthoringContext.lang);
+        var siteDropdownLangBundle = CMgs.getBundle("siteDropdown", CStudioAuthoringContext.lang);
```
russdanner, line 44: Why is the bundle variable called siteDropdownLangBundle.  This code has nothing to do with siteDropdown.  That's copy paste and it makes the code hard to read

russdanner: Maybe this label needs to moved to a more generic bundle.  Assuming the var name applies to the bundle. So the real issue is this resource is in the wrong bundle.

PR #730, opened 2017-12-20
Custom controls to be dependent one from another

Lpereira00:
Controls required to be dependent one from another, is checkbox group is empty, input will be required, if checkbox group is not empty, input will be optional


sumerjabri:
Per our chat, we can't accept this PR as it's a point solution, but we do understand the need and will support via future enhancements and via the new marketplace.

PR #768, opened 2018-02-08
#1848

vestrada:
https://github.com/craftercms/craftercms/issues/1848 - [studio-ui] Static assets folders are not refreshed after delete an image item when file name contains dots and spaces #1848


sumerjabri:
@vestrada please fix conflict and resubmit

vestrada:
@sumerjabri Done

PR #783, opened 2018-02-28
[studio-ui] Pasting a large tree takes a while without an indication what is happening #1902

jvega190:
https://github.com/craftercms/craftercms/issues/1902

### DIFF
--- a/static-assets/scripts/main.js
+++ b/static-assets/scripts/main.js
@@ -758,7 +758,7 @@
             }
 
             function authLoop() {
-                $scope.authDelay = 1000;
+                $scope.authDelay = 60000;
                 $scope.reLoginError = null;
                 var isInIframe = (window.location != window.parent.location) ? true : false;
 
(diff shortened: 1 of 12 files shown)


static-assets/scripts/main.js
```diff
@@ -758,7 +758,7 @@
             }
 
             function authLoop() {
-                $scope.authDelay = 1000;
+                $scope.authDelay = 60000;
```
sumerjabri, line 761: What's this about?

ui/scss/temp.scss
```diff
@@ -92,7 +92,7 @@ body, var {
   top: 0;
   left: 0;
   width: 100%;
-  z-index: 100;
+  z-index: 99;
```
sumerjabri, line 95: Please make sure the session times out, and we gray out the screen, everything is grayed out except the session timeout dialog.

sumerjabri (changes requested):

ui/scss/temp.scss
jvega190: Yes, actually this change fixes it. Also should have been a separate commit.

static-assets/scripts/main.js
jvega190: This is a timer for token validation, set to the same time as yui code, should have been a separate commit

sumerjabri (approved):

PR #846, opened 2018-06-06
#2153

vestrada:
[studio-ui] image picker fails update due to slow preview deployment #2153


russdanner:
Looks ok to me.  I'll test it.

PR #848, opened 2018-06-07
#2091

vestrada:
https://github.com/craftercms/craftercms/issues/2091 - [studio-ui] Setting Populate Expression property on date-time form engine control has no effect #2091


russdanner:
Operators are not respected.
White space causes functionality to fail

https://www.useloom.com/share/6e94614ca6d94f7aa86f2ddc88bdf216

PR #896, opened 2018-09-05
#2290, 2407

vestrada:
https://github.com/craftercms/craftercms/issues/2290 - [studio-ui] Update UI to add a global menu #2290


static-assets/ng-views/adminConsole.html
```diff
@@ -0,0 +1,25 @@
+<div class="admin-console-view col-lg-10 col-lg-offset-1">
+
+    <header class="screen-title">
+        <h1>Admin Console</h1>
```
sumerjabri: Can we rename this file and these headings to Global Menu instead?

sumerjabri (changes requested):

static-assets/ng-views/adminConsole.html
vestrada: Done

sumerjabri (changes requested):
Can we also rename the file adminConsole.html to global menu?

static-assets/ng-views/adminConsole.html
vestrada: Done

sumerjabri (approved):

PR #905, opened 2018-09-12
#2454

vestrada:
https://github.com/craftercms/craftercms/issues/2454 - [studio-ui] Switch to use /api/2/user API calls #2454


static-assets/components/cstudio-common/common-api.js
```diff
@@ -4074,7 +4074,7 @@ var nodeOpen = false,
              */
             getUserInfo: function(callback) {
                 var serviceUrl = this.getUserInfoServiceURL;
-                serviceUrl += CStudioAuthoringContext.user;
+                //serviceUrl += CStudioAuthoringContext.user;
```
sumerjabri: Can we just delete this line?

PR #1125, opened 2019-04-16
#2869

vestrada:
https://github.com/craftercms/craftercms/issues/2869 - [studio-ui] Create a global audit viewing tool #2869 - Adding System opt

### DIFF
--- a/static-assets/ng-views/audit.html
+++ b/static-assets/ng-views/audit.html
@@ -26,14 +26,17 @@
             <span class="pr20" ng-show="!audit.sites">{{ audit.site }}</span>
             <div class="btn-group" uib-dropdown is-open="sitesDropdown.isopen" style="margin-right: 5px;" ng-show="audit.sites">
                 <button type="button" class="btn btn-default dropdown-toggle text-capitalize" uib-dropdown-toggle title="{{audit.site || 'admin.audit.ALL_SITES' | translate}}" >
-                    {{audit.site || 'admin.audit.ALL_SITES' | translate }}  <span class="caret"></span>
+                    {{audit.siteLabel || audit.site || 'admin.audit.ALL_SITES' | translate }}  <span class="caret"></span>
                 </button>
                 <ul class="dropdown-menu" role="menu">
                     <li>
                         <a ng-click="audit.updateSite()">{{ 'admin.audit.ALL_SITES' | translate }}</a>
                     </li>
+                    <li>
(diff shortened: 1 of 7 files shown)


static-assets/ng-views/audit.html
```diff
@@ -26,14 +26,17 @@ <h1>{{ 'admin.audit.AUDIT' | translate }}</h1>
             <span class="pr20" ng-show="!audit.sites">{{ audit.site }}</span>
             <div class="btn-group" uib-dropdown is-open="sitesDropdown.isopen" style="margin-right: 5px;" ng-show="audit.sites">
                 <button type="button" class="btn btn-default dropdown-toggle text-capitalize" uib-dropdown-toggle title="{{audit.site || 'admin.audit.ALL_SITES' | translate}}" >
-                    {{audit.site || 'admin.audit.ALL_SITES' | translate }}  <span class="caret"></span>
+                    {{audit.siteLabel || audit.site || 'admin.audit.ALL_SITES' | translate }}  <span class="caret"></span>
```
rart, line 29: Why are there so many different possibilities of what label to display? Is this an API inconsistency? Lets log a backend ticket if that’s the case.

static-assets/scripts/admin.js
```diff
@@ -549,16 +549,22 @@
                 }, audit.defaultDelay);
             };
 
-            audit.updateSite = function(site){
+            audit.updateSite = function(site, isNoTranslate){
```
rart: Rename `isNoTranslate` to better indicate what it is. The logic now does the opposite of what I thought it’d do based on the name of the var.

rart (changes requested):

static-assets/ng-views/audit.html
vestrada: At this point we have 3 options.
1. All sites: This happen when we send site param empty to the api. We need to translate it.
2. Root Studio: Sumer wants to show to the user "System" label, instead of "Root Studio". So We need an extra label. Also we need to translate it
3. Sites (Other options): Each specific site selected by the user. 

So I'm trying to manage all those scenarios for it.
This only show what user selected.

rart: Why is the value under two different properties, though? site or siteLabel

vestrada: Site is siteID from sites API. It is displayed through an angular repeat. Site label is a static value, in this case "System" that needs to be translated.

static-assets/scripts/admin.js
vestrada: Done

rart (approved):

PR #1137, opened 2019-04-24
#3100

vestrada:
https://github.com/craftercms/craftercms/issues/3100 - [studio-ui] Missing validation msg in Push to Remote Git Repo and Use a remote Git repository instead #3100


rart (approved):
Code ok.
@yacdaniel to validate that the changed logic complies with correct validation flow.

PR #1142, opened 2019-04-25
 [studio-ui] When editing a content type and cancelling changes, the screen becomes unresponsive #3111

jvega190:


jvega190:
@sumerjabri the reason why were changing `hide` calls on dialogs to `destroy` is because when hidding the dialogs and opening new ones the library has conflicts with z-indexes and that's the reason why the dialogs show under the overlay. By destroy the dialogs, the markup gets cleaned and this way we avoid conflicts.

PR #1146, opened 2019-04-29
#3120

vestrada:
https://github.com/craftercms/craftercms/issues/3120 - [studio-ui] Preview Assets in Browse page for s3, cmis webdav is not working #3120

### DIFF
--- a/static-assets/components/cstudio-browse/browseWebDAV.js
+++ b/static-assets/components/cstudio-browse/browseWebDAV.js
@@ -222,7 +222,7 @@
         $('.cstudio-wcm-result .results').delegate( ".magnify-icon", "click", function() {
             var path = $(this).attr('data-source');
             var type = $(this).attr('data-type');
-            me.magnify(path, type);
+            CStudioAuthoring.Utils.previewAssetDialog(path, type);
         });
 
         var pathURL = CStudioAuthoring.Utils.getQueryParameterByName("path");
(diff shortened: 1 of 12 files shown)


rart (commented):
@vestrada please confirm: 

The FTLs contain markup for what looks to be a pop up (modal dialog). Are these related to the recent hide/destroy issue?

Since these dialogs are on the FTL destroying them might cause issues reopening them without refreshing the page. 

Do all of the hide/destroy fixes work when being opened various times without refreshing the page?

Why are these pop ups on the ftl but are basically blank?

vestrada:
No, It is not related. This is a custom pop up created for browse page, to preview a specific asset, I was just changing it because seems like there was a change when the search page was created. In this case there is just one dialog that will preview different assets. And it isn't yui dialog.

static-assets/components/cstudio-browse/browseWebDAV.js
```diff
@@ -222,7 +222,7 @@
         $('.cstudio-wcm-result .results').delegate( ".magnify-icon", "click", function() {
             var path = $(this).attr('data-source');
             var type = $(this).attr('data-type');
-            me.magnify(path, type);
+            CStudioAuthoring.Utils.previewAssetDialog(path, type);
```
rart, line 225: Will `CStudioAuthoring.Utils.previewAssetDialog` make use of the markup on the FTL to create the dialog? There's no FreeMarker output or unique differences between the different browse FTLs; it looks like duplicated HTML. 

Refactor for the pop up template to be on the JavaScript `previewAssetDialog` function and remove the HTML from the FTLs.

rart (changes requested):

vestrada:
Done

rart (approved):

PR #1150, opened 2019-05-02
#3128

vestrada:
https://github.com/craftercms/craftercms/issues/3128 - [studio-ui] Dependencies are shown incorrectly #3128


sumerjabri:
@vestrada why `master` and not `develop`?

PR #1153, opened 2019-05-03
#2812

vestrada:
https://github.com/craftercms/craftercms/issues/2812 - [studio-ui] Make CMIS datasources consistent with WebDav and S3 datasource #2812


rart (changes requested):
See changes on the other PR

rart (changes requested):
Re-submit including other PR's fixes

rart (approved):

PR #1154, opened 2019-05-03
#2812

vestrada:
https://github.com/craftercms/craftercms/issues/2812 - [studio-ui] Mke CMIS datasources consistent with WebDav and S3 datasource

### DIFF
--- a/static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
+++ b/static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
@@ -66,6 +66,7 @@ CStudioAuthoring.Dialogs.UploadCMISDialog = CStudioAuthoring.Dialogs.UploadCMISD
      */
     closeDialog:function() {
         this.dialog.destroy();
+        $("#cstudio-wcm-popup-div").off("keyup");
     },
 
     /**
@@ -104,9 +105,6 @@ CStudioAuthoring.Dialogs.UploadCMISDialog = CStudioAuthoring.Dialogs.UploadCMISD
             '</div> ' +
             '</div>';
 
-        document.getElementById("upload-popup-inner").style.width = "350px";
-        document.getElementById("upload-popup-inner").style.height = "180px";
-
         // Instantiate the Dialog
         upload_dialog = new YAHOO.widget.Dialog("cstudio-wcm-popup-div",
             { width : "360px",
@@ -139,7 +137,6 @@ CStudioAuthoring.Dialogs.UploadCMISDialog = CStudioAuthoring.Dialogs.UploadCMISD
         $("body").on("keyup", "#cstudio-wcm-popup-div", function(e) {
             if (e.keyCode === 27) {	// esc
                 me.closeDialog();
-                $("#cstudio-wcm-popup-div").off("keyup");
             }
         });
 
--- a/static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
+++ b/static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
@@ -1,0 +1,290 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
```diff
@@ -0,0 +1,290 @@
...
+        this.site = site;
+        this.path = path;
+        this.profile = repositoryId;
+        this.asPopup = true;
+        this.serviceUri = serviceUri;
+        this.callback = callback;
+        this.dialog.show();
+        document.getElementById("cstudio-wcm-popup-div_h").style.display = "none";
+
+        if(window.frameElement){
```
rart, line 52: Is this dialog always going to be on the same domain? 
Accessing parent/child windows is usually calling for trouble.

static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
```diff
@@ -0,0 +1,290 @@
...
+            '</div>' +
+            '<div class="contentTypePopupBtn"> ' +
+            '<input type="button" class="btn btn-primary cstudio-xform-button ok" id="uploadButton" value="Upload" disabled />' +
+            '<input type="button" class="btn btn-default cstudio-xform-button" id="uploadCancelButton" value="Cancel"  /></div>' +
+            '</form></div>' +
+            '<div><div  style="visibility:hidden; margin-bottom:1.5em;" id="indicator">Uploading...</div>' +
+            '</div> ' +
+            '</div>';
+
+        document.getElementById("upload-popup-inner").style.width = "350px";
```
rart: Since these are hard coded, why not set them directly on the HTML. But better yet, if they're not calculated or dynamic set these styles on a stylesheet to the class.

static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
```diff
@@ -0,0 +1,290 @@
...
+        YAHOO.util.Event.addListener(filenameInput, "change", this.uploadFileEvent);
+
+        var eventParams = {
+            self: this
+        };
+
+        YAHOO.util.Event.addListener("uploadButton", "click", this.uploadPopupSubmit, eventParams);
+        YAHOO.util.Event.addListener("uploadCancelButton", "click", this.uploadPopupCancel);
+
+        $("body").on("keyup", "#cstudio-wcm-popup-div", function(e) {
```
rart, line 137: This will persist past the dialog getting closed if not closed using the escape key. You should detach this event when the dialog is disposed from any means.

static-assets/components/cstudio-forms/data-sources/img-CMIS-upload.js
```diff
@@ -0,0 +1,116 @@
...
+		}
+   	} 
+	
+	return this;
+}
+
+YAHOO.extend(CStudioForms.Datasources.ImgCMISUpload, CStudioForms.CStudioFormDatasource, {
+	itemsAreContentReferences: true,
+
+    decreaseFormDialog: function(){
```
rart: Function is duplicate across various dialogs, share it. Same for other instances of this situation.

rart (changes requested):

static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
vestrada: Done

static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
vestrada: Done

static-assets/components/cstudio-forms/data-sources/img-CMIS-upload.js
vestrada: Done

static-assets/components/cstudio-dialogs/uploadCMIS-dialog.js
vestrada: FrameElement is getting the iframe which the element is. I don't think the domain is going to change because it is a dialog that it is open inside from engine.

rart (approved):

sumerjabri (changes requested):
Please add DE translation.

vestrada:
Sumer, 3.0.x doesn't has DE translation only 3.1.1

sumerjabri (approved):

PR #1158, opened 2019-05-13

jvega190:

### DIFF
--- a/static-assets/components/cstudio-admin/mods/workflow-states.js
+++ b/static-assets/components/cstudio-admin/mods/workflow-states.js
@@ -132,29 +132,26 @@ YAHOO.extend(CStudioAdminConsole.Tool.WorkflowStates, CStudioAdminConsole.Tool, 
 	    		"<option value='EXISTING_DELETED'>EXISTING_DELETED</option>" +
 	    	"</select><br/>" +
             CMgs.format(formsLangBundle, "setStatedDialogSystemProcessing")+": <input id='setProcessing' type='checkbox' value='false'/>" +
-	    "</div>";
+		"</div>";
 		
 		var handleSet = function() {
 			var state = document.getElementById('setState').value;
 			var processing = document.getElementById('setProcessing').checked;
-            var maxList = list.length - 1;
 			
 			for(var i=0;  i< list.length; i++) {
 				var item = list[i];
 				var path = item.path;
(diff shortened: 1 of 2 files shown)


static-assets/ng-views/settings.html
```diff
@@ -26,7 +26,7 @@ <h1>{{ 'settings.HEADER_LABEL' | translate }}</h1>
             <h2>{{ 'settings.PROFILE' | translate }}</h2>
             <div class="well">
                 <div class="user-display">
-                    <div class="name">{{user.name}} {{user.surname}}</div>
+                    <div class="name">{{user.name || user.first_name}} {{user.surname || user.last_name}}</div>
```
rart, line 29: Are both property name alternatives in use? If so, when and why two different property names for same thing?

static-assets/components/cstudio-admin/mods/workflow-states.js
```diff
@@ -137,21 +137,24 @@ YAHOO.extend(CStudioAdminConsole.Tool.WorkflowStates, CStudioAdminConsole.Tool,
...
-				cb = { 
-						success:function() {
-							CStudioAdminConsole.Tool.WorkflowStates.prototype.renderStatesTable();
-						}, 
-						failure: function() {} 
-				};
-
-				YConnect.initHeader(CStudioAuthoringContext.xsrfHeaderName, CStudioAuthoringContext.xsrfToken);                                        
-				YConnect.asyncRequest("POST", CStudioAuthoring.Service.createServiceUri(serviceUri), cb);
+				if(maxList <= i){
```
rart: Looks like in the truthy condition, the dialog is destroyed while the request is in flight. Is there any sort of progress indicator for while the request is inflight? Should the dialog be closed until the request is done? Does it depend at all on the result of the request?

rart (changes requested):

static-assets/ng-views/settings.html
jvega190: What happens is that user data is being set in two scenarios.

1.  No user logged in studio. When there's no user you need to login, and login callback has user data (in this case the prop is `user.name`, for example)

2.  User already logged in. When user is already logged in into studio the user data comes from model, and is set on a variable from the freemarker template, parsed and it's retrieved as the user data (in this scenario the props have different names, like `user.first_name`).

PR #1159, opened 2019-05-13

jvega190:


rart (changes requested):
See comments on other PR.

The other PR had another file changed as well. Shouldn't those changes come here too?

rart (approved):

PR #1160, opened 2019-05-13

jvega190:


jvega190:
@rart  What happens is that user data is being set in two scenarios.

1.  No user logged in studio. When there's no user you need to login, and login callback has user data (in this case the prop is user.name, for example)

2.  User already logged in. When user is already logged in into studio the user data comes from model, and is set on a variable from the freemarker template, parsed and it's retrieved as the user data (in this scenario the props have different names, like user.first_name).

rart (changes requested):
Can we change the FreeMarker prop names to match the API prop names? 
If so, please do.
If not, why not?

jvega190:
@rart yes it was possible, PR updated

rart (approved):

rart (changes requested):
Please camelcase the attributes

rart (approved):

PR #1163, opened 2019-05-16
 [studio-ui] add ability to default time in the date time picker form control #2648

jvega190:
https://github.com/craftercms/craftercms/issues/2648


static-assets/components/cstudio-forms/controls/date-time.js
```diff
@@ -1210,15 +1210,18 @@ YAHOO.extend(CStudioForms.Controls.DateTime, CStudioForms.CStudioFormField, {
 				currentDate.setDate(currentDate.getDate()+(modifier*value*daysInWeek))
 			}else if(type=="days") {
 				currentDate.setDate(currentDate.getDate()+(modifier*value));
+			}else if(type=="hours") {
+				currentDate.setTime(currentDate.getTime() + (modifier*(value*60*60*1000)));
```
rart, line 1214: Change all these to triple equals

Multiply against the static resulting value (3,600,000) instead of performing the open every time

rart (changes requested):

rart (changes requested):
Files from a different PR sneaked on to this one: admin.js, settings.html

PR #1173, opened 2019-05-28
…

vestrada:
https://github.com/craftercms/craftercms/issues/3166 - [studio-ui] let user know they don't have access to any sites when sites dashboard is empty #3166


rart (changes requested):
Same as analog PR

vestrada:
Done

rart (changes requested):
Why is the copyright included multiple times on `main.css`?

vestrada:
Seems like all scss files have included the copyright. There are some files that are imported to main files, that is why you are seeing it multiple times. If we want to remove it we should remove the copyright from imported files.

vestrada:
Done

rart (approved):

PR #1175, opened 2019-05-29
Adding a sass sheet build script

rart:
#3118 - https://github.com/craftercms/craftercms/issues/3118

Pending to add all sheets & possibly configure output directory


ui/scss/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "studio-scss",
+  "version": "1.0.0",
+  "description": "SASS for studio",
+  "private": true,
+  "main": "index.js",
```
jdrossl, line 6: Just a question, how do I tell yarn to run this file?

jdrossl (approved):

ui/scss/package.json
rart: You may create a scripts section on the `package.json` like so:
```
  "scripts": {
    "build": "node index.js"
  },
```
Then you'd do `yarn build` to build it through yarn.

You may also directly call `node index.js` which is what I did to run/test during my testing.

jdrossl: I have to check to see if that is possible with the maven plugin, because it uses a custom node binary

PR #1179, opened 2019-06-03
#2808

vestrada:
https://github.com/craftercms/craftercms/issues/2808 - studio-ui] child-content data source should allow user to decide if add or browse options are enabled #2808

### DIFF
--- a/static-assets/components/cstudio-forms/data-sources/child-content.js
+++ b/static-assets/components/cstudio-forms/data-sources/child-content.js
@@ -25,6 +25,7 @@ function(id, form, properties, constraints)  {
 	this.type = "";
     this.defaultEnableCreateNew = true;
     this.defaultEnableBrowseExisting = true;
+    this.countOptions = 0;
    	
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/data-sources/child-content.js
```diff
@@ -41,19 +42,94 @@ function(id, form, properties, constraints)  {
         if(properties[i].name === "enableCreateNew"){
             this.enableCreateNew = properties[i].value == "true" ? true : false;
             this.defaultEnableCreateNew = false;
+            properties[i].value == "true" ? this.countOptions ++ : null;
```
rart: `===`

rart (changes requested):

static-assets/components/cstudio-forms/data-sources/child-content.js
vestrada: Done

rart (approved):

PR #1183, opened 2019-06-04
#2588

vestrada:
https://github.com/craftercms/craftercms/issues/2588 - [studio-ui] Add the name of the environment to the Configuration section of Site Config #2588


static-assets/components/cstudio-admin/mods/admin-configurations.js
```diff
@@ -110,9 +110,10 @@ CStudioAuthoring.Module.requireModule(
 
                 var callback = {
                     success: function (data) {
-                        if (data.environment){
+                        var environment = YAHOO.lang.JSON.parse(data.responseText).environment;
```
rart, line 113: use `JSON.parse` instead - generally speaking

PR #1184, opened 2019-06-04
[studio-ui] Add highlight to Sidebar #2901 - Bugfixes

jvega190:
https://github.com/craftercms/craftercms/issues/2901

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -385,10 +385,6 @@ var nodeOpen = false,
 
                 var event = new CustomEvent("setContentDone");
                 document.dispatchEvent(event);
-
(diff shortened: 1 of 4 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -7780,14 +7803,14 @@ var nodeOpen = false,
                         "</tr>"].join("");
                 }
 
-                if(lockOwner && lockOwner != null && lockOwner.trim() != "") {                    
+                if(lockOwner && lockOwner != null && lockOwner.trim() != "") {
```
rart, line 7806: `lockOwner && lockOwner != null` are redundant conditions.
avoid potential type coercion, use `!==`.

Looks like `if ((lockOwner || "").trim() !== "")` would suffice :)

PR #1193, opened 2019-06-12
Missing asset, react system adjustments, ajax lib

rart:


ui/app/src/utils/ajax.ts
```diff
@@ -0,0 +1,41 @@
+import { ajax } from 'rxjs/ajax';
```
sumerjabri, line 18: Missing copyright?

sumerjabri (changes requested):

ui/app/src/utils/ajax.ts
rart: Oopsie! Fixed.

sumerjabri (approved):

PR #1196, opened 2019-06-14
User dashboard optimizations: fixing redundant checks and additional ng renders

rart:
* Removing redundant user-session check.
* Defaulting login screen logo URL to Crafter CMS logo.


rart:
Note: `/get-ui-resource-override?resource=logo.jpg` request always results in a 404. 

Was this meant for customers to be able to put their logos on the login screen? If so, perhaps that service should always return the logo URL to be shown - defaulting to Crafter CMS logo and overridable by configurable by org admins.

Is that in use at all? If so, how is this configured?

If this is not in use, I can update the PR to remove that call all together. This would save a network request as well as an error on both browser and server consoles :)

PS. This error also occurs in studio once logged in.

sumerjabri:
@rart Remove logo override.

sumerjabri (changes requested):
@rart Remove logo override.

rart:
> @rart Remove logo override.

@sumerjabri done, updated. Will create a separate PR for the intra studio piece as that one is more delicate.

rart:
Done completely removing the logo override calls all over the place.

PR #1201, opened 2019-06-19
#3221

vestrada:
https://github.com/craftercms/craftercms/issues/3221 - [studio-ui] Please change top nav labels to tooltips.


rart (changes requested):
Please add title attribute to all tooltipped items.

The bootstrap ones should work with simply `title` instead of `data-title`.

Add the title attribute to the ones on the angular side.

(Make it accessible)

vestrada:
Done

rart (approved):

PR #1208, opened 2019-06-25
[studio-ui] Change the "Content Type" and "Mime Type" filter selection to checkboxes in Studio search #3135

jvega190:
[studio-ui] Change the "Content Type" and "Mime Type" filter selection to checkboxes in Studio search #3135
[studio-ui] Search Results page: Default the cursor to the search box #3158

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -6274,26 +6274,29 @@ var nodeOpen = false,
                     pl     = /\+/g,  // Regex for replacing addition symbol with a space
                     search = /([^&=]+)=?([^&]*)/g,
                     decode = function (s) { return decodeURIComponent(s.replace(pl, " ")); },
-                    query  = window.location.search.substring(1);
+                    query  = window.location.search.substring(1),
+                    key,
+                    value;
 
                 urlParams = {};
-                while (match = search.exec(query)){
-                    // urlParams[decode(match[1])] = decode(match[2]);
+                while (match = search.exec(query)) {
+                  key = decode(match[1]),
+                  value = decode(match[2]);
 
-                    // if urlPamars at position already exists
-                    if (urlParams[decode(match[1])]) {
-                      // if current value is string - create array and add both
-                      if (typeof urlParams[decode(match[1])] === 'string') {
-                        var valuesArray = [];
-                        valuesArray.push(urlParams[decode(match[1])]);
-                        valuesArray.push(decode(match[2]));
-                        urlParams[decode(match[1])] = valuesArray;
-                      }else{
-                        urlParams[decode(match[1])].push(decode(match[2]));
-                      }
-                    } else {
-                      urlParams[decode(match[1])] = decode(match[2]);
+                  // if urlPamars at position already exists
+                  if (urlParams[key]) {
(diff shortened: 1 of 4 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -6278,7 +6278,22 @@ var nodeOpen = false,
 
                 urlParams = {};
                 while (match = search.exec(query)){
-                    urlParams[decode(match[1])] = decode(match[2]);
+                    // urlParams[decode(match[1])] = decode(match[2]);
+
+                    // if urlPamars at position already exists
+                    if (urlParams[decode(match[1])]) {
```
rart: Store decode product in var instead of multiple time decode call

static-assets/components/cstudio-search/search.js
```diff
@@ -201,6 +204,28 @@
...
+        // Selecting a multiple value filter
+        $('.cstudio-search').on('change', '.filter-item input[type="checkbox"]', function(){
+          var searchFilters = CStudioSearch.searchContext.filters,
+              filterName = $(this).attr('name'),
+              filterValues = [];
+
+          //timeout to wait for other checkboxes selections
+          clearTimeout(filterTimeout);
+          filterTimeout = setTimeout(function(){
+            $.each($(".filter[name='" + filterName + "']:checked"), function(){
```
rart: `$(".filter[name='" + filterName + "']:checked").each(function`

templates/web/search.ftl
```diff
@@ -42,13 +42,12 @@
...
 
     <script src="/studio/static-assets/scripts/crafter.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
-    <script src="/studio/static-assets/scripts/animator.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>    
+    <script src="/studio/static-assets/scripts/animator.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
     <link rel="stylesheet" type="text/css" href="/studio/static-assets/yui/assets/skins/sam/calendar.css?version=${UIBuildId!.now?string('Mddyyyy')}"/>
     <link rel="stylesheet" type="text/css" href="/studio/static-assets/styles/search.css?version=${UIBuildId!.now?string('Mddyyyy')}"/>
     <script src="/studio/static-assets/libs/js-cache/cache.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
     <script src="/studio/static-assets/libs/amplify/lib/amplify.core.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
 
-    <script src="/studio/static-assets/libs/bootstrap/js/bootstrap.min.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
```
rart, line 51: Tooltips for the context nav use bootstrap so this will likely break them. Please check.

rart (changes requested):

templates/web/search.ftl
jvega190: Bootstrap was previously added into head.ftl (common file included in this template) and it caused bootstrap to be imported twice, and causing problems. That's why it was removed from here. (Head is included some lines before, so we do have bootstrap)

rart (approved):

PR #1213, opened 2019-06-27
[studio-ui] Remove XSRF cookie property from authoring context #3253

jvega190:
https://github.com/craftercms/craftercms/issues/3253

### DIFF
--- a/ui/app/src/utils/storage.ts
+++ b/ui/app/src/utils/storage.ts
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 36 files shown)


ui/app/src/utils/storage.ts
```diff
@@ -0,0 +1,26 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Cookies from 'js-cookie';
+
+export function getRequestForgeryToken() {
+  return Cookies.get('XSRF-TOKEN');
```
rart, line 21: We currently have a constant for the cookie name. At least in Studio authoring context but this one is not available in the ng app. Where do we get if from there? If in the ng app is hard coded is safe to assume we don’t change the name across customers in which case let’s just drop the constant (for now). Otherwise, please use constant here. PM me to discuss.

rart (changes requested):
Looks like we missed cleaning up`studio-context.ftl`.

rart (approved):

PR #1231, opened 2019-07-09
Support customization of component icons/styles in Studio

zac-poe:
This change allows you to leverage the existing mime-type.xml configuration to also define styles for content types.

If you create a custom content type **video**, the following configuration will allow you stylize your components in the Crafter Studio sidebar:
`    <mime-type>
        <type>/component/video</type>
        <icon>
            <class>fa-video-camera</class>
        </icon>
    </mime-type>
` 

image
\* _This sample screenshot also uses a custom sidebar.xml folder, but with the logic in this PR, any content type from any wcm-root-folder can be customized._

This PR also provides support for defining a single mime-type value in mime-type.xml, which previously had a logic issue between the SiteServiceImpl parsing for single items and the looping logic in the Javascript code.

If this PR is accepted, I would recommend also updating the comment description in mime-type.xml to something like: 
> \- type: The mime type for assets, or content type for components. This is the target type that will be affected by the new icon/styles defined on the configuration

This change was not included here since those files come from the separate studio repository - please let me know if an additional PR is desired for this.

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -7331,7 +7331,10 @@ var nodeOpen = false,
                     }
 
                 }else{
-                    if(treeNodeTO.isComponent){     //isLevelDescriptor - also component
+                    if(customIcons[treeNodeTO.contentType]){ //support any local overrides for content type styling
+                        mainIconClass = customIcons[treeNodeTO.contentType].class;
+                        customStyle = customIcons[treeNodeTO.contentType].style;
+                    }else if(treeNodeTO.isComponent){     //isLevelDescriptor - also component
                         mainIconClass = defaultIcons.component.class;
                     }else if(treeNodeTO.isPage){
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -9098,8 +9101,13 @@ CStudioAuthoring.FilesDiff = {
                 key;
 
             if(data && data["mime-type"]){
-                for( var i = 0; i < data["mime-type"].length; i++){
-                    confMimeType = data["mime-type"][i];
+                var mimeTypes = data["mime-type"];
+                if(!Array.isArray(mimeTypes)) {
+                    //support single values coming from SiteServiceImpl#createMap
+                    mimeTypes = [mimeTypes];
```
zac-poe, line 9107: This change is not required to support customization of component icons, but without it, you must add 2 or more mime-type configurations for the mime-type.xml file to be processed

rart (approved):

PR #1237, opened 2019-07-12
Fix #3281: [studio-ui] Cannot add users to group when there are more …

suelje:
Fix https://github.com/craftercms/craftercms/issues/3281: [studio-ui] Cannot add users to group when there are more than 10 users available

Overrule default limit parameter when requesting Studios getAllUsers() API endpoint


sumerjabri:
Longer term, we need to follow: https://github.com/craftercms/craftercms/issues/3283

rart (approved):

PR #1240, opened 2019-07-18
[studio-ui] WebDav Image Browse datasource not working #3299

jvega190:
https://github.com/craftercms/craftercms/issues/3299


rart (approved):
@jvega190 please double check that `openWebDAVBrowse` is not invoked from anywhere else still using the removed param.

PR #1264, opened 2019-08-19
Postfix errors popup broken #3372

aachavarria:


rart (commented):
Please check with different content lengths and on windows/linux chrome

rart (approved):

PR #1267, opened 2019-08-22
German language updates for CrafterCMS v3.1.0

suelje:


static-assets/components/cstudio-common/resources/de/base.js
```diff
@@ -238,12 +238,12 @@ CStudioAuthoring.Messages.registerBundle("previewTools", "de", {
     sites: "Projekte",
     help: "Hilfe",
     documentation: "Dokumentation",
-    about: "Über",
+    about: "Über CrafterCMS",
```
rart: Please spell as “Crafter CMS” (space missing)

static-assets/scripts/resources/locale-de.json
```diff
@@ -272,32 +273,32 @@
...
         "SUCCESSFULLY_PULLED": "Pull war erfolgreich",
         "REPO_DELETED": "Repository wurde gelöscht",
-        "MERGE_STRATEGY_LABEL": "Strategie zusammenführen",
+        "MERGE_STRATEGY_LABEL": "Merge-Strategie",
         "MERGE_STRATEGIES": {
-          "none": "Keiner",
-          "ours": "Akzeptiere unsere",
-          "theirs": "Akzeptiere ihre"
+          "none": "Keine",
+          "ours": "OURS",
```
rart: Is it generally used and better understood as just “ours” / “theirs” in German instead of accept ours / accept theirs? Is the all capitals intentional?

static-assets/scripts/resources/locale-de.json
```diff
@@ -365,7 +363,7 @@
     "OK":"OK",
     "HELP": "Hilfe",
     "DOCUMENTATION": "Dokumentation",
-    "ABOUT": "Über",
+    "ABOUT": "Über CrafterCMS",
```
rart: “Crafter CMS” please

static-assets/scripts/resources/locale-de.json
```diff
@@ -330,28 +330,26 @@
...
+      "EXCEPTION": "Exception"
     }
   },
   "common": {
     "CANCEL_LABEL": "Abbrechen",
     "SEARCH_LABEL": "Suchen",
     "CLOSE_LABEL": "Schließen",
     "ADD_LABEL": "Hinzufügen",
-    "CREATE": "Erstellen",
+    "CREATE": "create",
```
rart: Please confirm this is the correct capitalization. Same for the “edit” below.

rart (changes requested):
Thanks for the language contributions! Please review comments.

static-assets/scripts/resources/locale-de.json
suelje: Hmm, there is no good translation into german. When it comes to git-related terms, I would keep them in english.

static-assets/scripts/resources/locale-de.json
suelje: These are the "technical names" of the dialogs. They need to be consistent through all languages, e.g. `<div class="form-group" ng-if="!(dialogMode == 'create')"> `

(I think they should not be part of the resource bundles)

static-assets/scripts/resources/locale-de.json
rart: Makes sense.

Did you write ”OURS” and “THEIRS” all in capitals on purpose? Instead of regular capitalization (i.e. Ours, Theirs)

suelje: I've changed them to lowercase

static-assets/scripts/resources/locale-de.json
rart: Could you point me to where in the code have you seen a message bundle item used to be compared as a constant? Indeed that shouldn’t happen. 

However, I’d think these items (Create and Edit) are rather used to title a dialog (e.g. in the header of a pop up) or something of sorts. Actual text on the UI, in which case they should keep the capitalisation and should be translated.

suelje: https://github.com/craftercms/studio-ui/blob/master/static-assets/ng-views/admin-users.html#L95
`dialogMode` in here is taken from the resource bundle. So in this case, any other value than `create` would break the user creation dialog, see:

dialogMode

rart (approved):
I've gone ahead and made a couple of tweaks to capitalization on top of your PR and consolidated those keys that should be translations but were split in _translation vs constant_ usage. 

You may want to pull to get those updates
resources/de/base.js @ 543 & 544
resources/locale-de.json @ 278 & 279
resources/locale-de.json @ 349 - 351

There'll be a PR fixing the dialog mechanics and using the consolidated keys the right way. Thanks for making us aware of this.
https://github.com/craftercms/craftercms/issues/3388

And again, thanks so much for helping in making better German translations!

static-assets/scripts/resources/locale-de.json
rart: This is a bug! Thanks for pointing it out.
https://github.com/craftercms/craftercms/issues/3388

rart:
FYI: https://github.com/craftercms/studio-ui/pull/1269

PR #1286, opened 2019-09-04
[studio-ui] Change Publish status "Idle" to "Ready" #3351

aachavarria:
Added missing translations.
Added translations support to the publishing status on admin.js.
Created a if conditional to support the new key ready.

### DIFF
--- a/static-assets/components/cstudio-contextual-nav/status.js
+++ b/static-assets/components/cstudio-contextual-nav/status.js
@@ -175,12 +175,12 @@ CStudioAuthoring.ContextualNav.StatusNavMod = CStudioAuthoring.ContextualNav.Sta
 
             CStudioAuthoring.Service.getPublishStatus(CStudioAuthoringContext.site, {
               success: function (response) {
-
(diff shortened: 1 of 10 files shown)


static-assets/components/cstudio-contextual-nav/status.js
```diff
@@ -175,12 +175,12 @@ CStudioAuthoring.ContextualNav.StatusNavMod = CStudioAuthoring.ContextualNav.Sta
 
             CStudioAuthoring.Service.getPublishStatus(CStudioAuthoringContext.site, {
               success: function (response) {
-
+                let message = (response.status.toLowerCase() === 'idle')? CMgs.format(contextNavLangBundle, 'ready') : CMgs.format(contextNavLangBundle, response.status.toLowerCase());
```
rart: A bit less repeated code this way:
```js
let message = CMgs.format(
  contextNavLangBundle, 
  (response.status.toLowerCase() === 'idle') ? 'ready' : response.status.toLowerCase()
);
```

rart (approved):

static-assets/components/cstudio-contextual-nav/status.js
aachavarria: updated, also I added a new fn because the same message is called in other places.

PR #1289, opened 2019-09-05
[studio-ui] Language settings are persisted per browser and not per user #3290

jvega190:
https://github.com/craftercms/craftercms/issues/3290

### DIFF
--- a/templates/web/common/page-fragments/studio-context.ftl
+++ b/templates/web/common/page-fragments/studio-context.ftl
@@ -51,7 +51,7 @@
    	}
 
   var lang = localStorage.getItem(CStudioAuthoringContext.user + '_crafterStudioLanguage');
-  CStudioAuthoringContext.lang = lang;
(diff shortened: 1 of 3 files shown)


rart (changes requested):
Let's change all the mechanics to work with local storage instead of cookies as the setting is of no use to the back end.

templates/web/common/page-fragments/studio-context.ftl
```diff
@@ -50,6 +50,8 @@
    		document.location = CStudioAuthoringContext.baseUri;
    	}
 
+  var lang = localStorage.getItem(CStudioAuthoringContext.user + '_crafterStudioLanguage');
```
rart, line 53: This could be null. Need fallback.

rart (changes requested):
Change the semantic references to “cookies”. Event though it’s now local storage it still references cookies in comments and var names,

rart (approved):

PR #1284, opened 2019-09-02
Feature/2884

aachavarria:
https://github.com/craftercms/craftercms/issues/2884

PR #1290, opened 2019-09-05
Rollback PR #1287 and PR #1284

aachavarria:
This rollback is because using the new name "repositoryPath" it is only going to work with new dataSources, the blueprints were created using dataSources with "repoPath".


rart (changes requested):
Missing message bundle translation update to address the root cause of this whole chain

aachavarria:
Form Definition xml of the Editorial Blueprint

Screen Shot 2019-09-05 at 10 43 35 AM

rart (approved):

PR #1295, opened 2019-09-06
[studio-ui] Some fields should be excluded from the postfix requirement #3432

jvega190:
https://github.com/craftercms/craftercms/issues/3432

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-types.js
+++ b/static-assets/components/cstudio-admin/mods/content-types.js
@@ -50,6 +50,7 @@ CStudioAuthoring.Module.requireModule(
             {
                 success: function(config) {
                     CStudioAdminConsole.isPostfixAvailable = config["form-engine"] && config["form-engine"]["field-name-postfix"] === "true" ? true : false;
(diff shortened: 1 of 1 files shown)


rart (approved):

static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -96,19 +97,19 @@ CStudioAuthoring.Module.requireModule(
...
-                            for (var k = 0; k < postfixes.length; k++) {
-                                if (currentField.id.indexOf(postfixes[k]) > -1) {
-                                    postfixesFlag = true;
-                                    break;
-                                }
-                            }
-                            if(!postfixesFlag && postfixes.length > 0){
-                                postfixError.push({"title" : currentField.title, "type" : currentField.type});
+                        if( (currentField.id || currentField.id !== '') && (currentField.title && currentField.title !== '')
+                          && ( CStudioAdminConsole.ignorePostfixFields.indexOf(currentField.id) < 0 ) ) {
```
rart, line 101: use includes instead of index of

rart (changes requested):

PR #1297, opened 2019-09-10
[studio-ui] Template Code Examples #3430

jvega190:
https://github.com/craftercms/craftercms/issues/3430


rart (changes requested):
Pending discussed changes including changing `values` to keys and holding actual snippets elsewhere in the code (not in the DOM).

rart (approved):

PR #1302, opened 2019-09-13
[studio-ui] Wrong array initialization in forms-engine #3455

jvega190:
https://github.com/craftercms/craftercms/issues/3455

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -2783,7 +2783,7 @@ var CStudioForms = CStudioForms || function() {
       for (var key in formModel) {
 
         var
-          attributes = [],
+          attributes = [' '],
           fieldInstruction = fieldInstructions[key],
           fieldList = fieldLists[key],
           invalidFields = [],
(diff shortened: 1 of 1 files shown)


rart (changes requested):
This was done this way for a reason. What seems to be the problem?

Removing the space may cause issues writing the xml as the first attribute and the tag name would get concatenated. For example: `<form name=“”>` would result in `<formname=“”>`.
There’s also logic below - in line 2815 - that depends on this initialisation.

jvega190:
There's already an space added in line 2868 to avoid the tag name and attribute to be concatenated. The problem is that the array will be like `[" ", attribute1, ...]` and then they'll be added comma separated, causing the xml to be like `<form ,name="">...` (adds a comma that causes the xml to be incorrect). Line 2815 needs to be removed. Please let me know what you think

rart:
Why are they comma separated? Those should be space-separated. There should be a join invocation somewhere and it should be using space as the separator, not comma.

jvega190:
Oh, yeah.. Actually that is the issue, let me see why is that comma being added, I'll be sending the fix

jvega190:
@rart the problem was that in an else there was no `join` being used to add the attributes. PR updated

rart (approved):
:)

PR #1304, opened 2019-09-16
#3285

vestrada:
https://github.com/craftercms/craftercms/issues/3285 - [studio-ui] on delete warn user if there are existing dependencies that depend on the object being deleted #3285


rart (changes requested):
Need to refactor the react commonalities of this dialog with the publishing one. Part of it is moving material components to their own file containing the common styles. 

Need to incorporate new i18n (new mechanics still pending merge, though).

These two could possibly get exported out of the same file.

rart (changes requested):
Conflicts

rart (approved):

PR #1305, opened 2019-09-17
[studio-ui] History: Revert of item removes the ability to revert until screen refresh #3423

jvega190:
https://github.com/craftercms/craftercms/issues/3423

### DIFF
--- a/static-assets/components/cstudio-admin/mods/admin-configurations.js
+++ b/static-assets/components/cstudio-admin/mods/admin-configurations.js
@@ -16,468 +16,475 @@
  */
 
 CStudioAuthoring.Module.requireModule(
-	'ace',
-	'/static-assets/components/cstudio-common/ace/ace.js', {}, {
-	moduleLoaded: function() {
+  'ace',
+  '/static-assets/components/cstudio-common/ace/ace.js', {}, {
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-admin/mods/admin-configurations.js
```diff
@@ -168,105 +182,115 @@ CStudioAuthoring.Module.requireModule(
...
+          self.loadSelectedConfig();
+				}; // end of change
+      },
+
+      loadSelectedConfig: function() {
+        const self = this,
+              itemSelectEl = this.configInfo.itemSelectEl,
+              editAreaEl = this.configInfo.editAreaEl,
+              editor = this.configInfo.editor,
+              sampleEditor = this.configInfo.sampleEditor;
```
rart: Where are these new vars getting used?

rart (changes requested):

static-assets/components/cstudio-admin/mods/admin-configurations.js
jvega190: They used to be sent as parameters to the function (those vars are being used), now they are global properties of the object

rart (approved):

PR #1309, opened 2019-09-19
[studio-ui] Cleanse/escape the logs before displaying them in the site log viewer #2871

aachavarria:
[studio-ui] Cleanse/escape the logs before displaying them in the site log viewer #2871


rart (changes requested):
use code or pre tags instead of blockquote for the exception

check rendering when exception is blank or null. in fact, check for all.

rart (approved):
Please submit the other locales (besides English)

aachavarria:
Updated.

rart (approved):

PR #1320, opened 2019-09-24
[studio-ui] Clicking the label on dependency selection dialogs does not toggle the checkboxes #3466

jvega190:
https://github.com/craftercms/craftercms/issues/3466


rart (approved):
1. Extract function and use in both click events.
```js
{
  items.map((item) => {
    const click = () => ...
    return {
      <div className="dependency-selection--section-dependencies" key={item.uri}>
```

2. Apply fix for the child items as well.

PR #1323, opened 2019-09-30
Additional option to search for child-content

suelje:
Old feature request: https://github.com/craftercms/craftercms/issues/256

We extended the data-source child-content for an option to use search instead of browse.

**Detailed description**
Basically, in our crafter site we have page assets with slots to contain components. Components are reused on different pages and different tenants. A "slot" is managed by a datasource of type child-content. We already have a lot of components, which caused problems because all components were created in a couple of folders only. That’s why we introduced [macros](https://docs.craftercms.org/en/3.1/developers/content-modeling.html#macros-for-data-sources) for automated organizing components in folders like ‘/site/components/{yyyy}/{mm}/{dd}’
By doing so, folders are “small” enough, but now content editors had problems trying to find existing components in the child-content datasource. In order to use the browse functionality, they would need to know the exact day, when a component has been created... and content editors never know :slightly_smiling_face:
That’s why we have added an option to use search functionality in addition or instead of “Browse for existing”.
child-content

We've extended Search in Crafter by a _select_ mode and use 'browsePath' property as search folder:
search

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -1128,6 +1128,14 @@ var nodeOpen = false,
                     searchUrl += "&page=1";
                 }
 
+                if (!CStudioAuthoring.Utils.isEmpty(searchContext.mode)) {
+                  searchUrl += "&mode=" + searchContext.mode;
+                }
+
+                if (!CStudioAuthoring.Utils.isEmpty(searchContext.itemsPerPage)) {
+                  searchUrl += "&ipp=" + searchContext.itemsPerPage;
+                }
+
                 var childSearch = null;
 
                 if (!searchId || searchId == null || searchId == "undefined"
@@ -1135,6 +1143,7 @@ var nodeOpen = false,
                     childSearch = CStudioAuthoring.ChildSearchManager.createChildSearchConfig();
                     childSearch.openInSameWindow = openInSameWindow;
                     searchId = CStudioAuthoring.Utils.generateUUID();
+                    searchUrl += "&searchId=" + searchId;
 
                     childSearch.searchId = searchId;
                     childSearch.searchUrl = searchUrl;
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -1136,6 +1136,10 @@ var nodeOpen = false,
                   searchUrl += "&ipp=" + searchContext.itemsPerPage;
                 }
 
+                if (!CStudioAuthoring.Utils.isEmpty(searchContext.query)) {
+                    searchUrl += "&query=" + searchContext.query;
+                }
+
                 var childSearch = null;
 
                 if (!searchId || searchId == null || searchId == "undefined"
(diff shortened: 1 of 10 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -2034,7 +2047,7 @@ var nodeOpen = false,
                 });
 
                 $modal.find('.bd').append(template).end().appendTo(parentEl);
-                $modal.find('.studio-ice-container').css('z-index', 100525);
+                $modal.find('.studio-ice-container').css('z-index', 1035);
```
rart, line 2050: Did you have issues with the z-index? What compelled you to change?

sumerjabri:
@suelje We just completed the "Shared Content" and "Embedded Content" data sources, and have deprecated "Child Content."

Please update your PR to be for the new data sources, and feel free to ping us on Slack or here if you need any help/support.

Thank you for contributing to Crafter CMS :)

static-assets/components/cstudio-forms/data-sources/child-content.js
```diff
@@ -24,6 +24,7 @@ CStudioForms.Datasources.ChildContent = function (id, form, properties, constrai
   this.type = '';
```
rart, line 24: Please move changes done over this file to data-sources/shared-content.js

templates/web/search.ftl
```diff
@@ -39,7 +39,6 @@
     <script src="${path}momentjs/moment-timezone-with-data-2012-2022.min.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
 
     <#include "/templates/web/common/page-fragments/studio-context.ftl" />
-    <#include "/templates/web/common/page-fragments/context-nav.ftl" />
```
rart, line 42: Removing this include will break regular studio search. 

You could potentially use a query string param to conditionally include this on the FTL (e.g. the _searchMode_ or so)

rart (changes requested):
Please functionality to `data-sources/shared-content.js`

Made some notes around the code. Feel free to ping me with any question :)

Thanks

static-assets/components/cstudio-common/common-api.js
suelje: Yeah, I had issues with z-index; e.g when you edit a page and you use "open search for select". That search layer must be above (black, semitransparent) overlay, but with this commit, we give also the option to open a new layer for editing an asset listed in search results. That specific z-index was the only one, which worked for us.
If we omit this commit, we won't be able to delete or to edit assets when in search select mode. You will only have the option to preview media assets.

Bildschirmfoto 2019-10-19 um 08 43 32

templates/web/search.ftl
suelje: I removed the static include from the template because I don't want the sidebar when in search-select mode. When in regular studio search, the sidebar is loaded here: https://github.com/eXa-online/studio-ui/blob/feature/search-childcontent/static-assets/components/cstudio-search/search.js#L89-L93
But you are right. A query string parameter might be easier. I will give it a try

suelje:
Updated PR to be for the new data source : https://github.com/craftercms/studio-ui/pull/1353
This one can be closed

rart:
Close in favour of https://github.com/craftercms/studio-ui/pull/1353

PR #1327, opened 2019-10-04
[studio] Create APIs to support "Publishing status dashboard" #3429

aachavarria:
https://github.com/craftercms/craftercms/issues/3429
https://github.com/craftercms/craftercms/issues/3352


ui/app/src/components/PublishingPackage.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+interface PublishingPackage {
+  package: Package
+}
+
+export default function PublishingPackage(props: PublishingPackage) {
+  const classes = useStyles({});
+  const {formatMessage} = useIntl();
+
+  const {id, approver, schedule, state, comment, environment} = props.package;
```
rart: If this component represents the package, perhaps we can pass the package props straight as direct props; meaning not under a `package` prop.

ui/app/src/services/publishing.tsx
```diff
@@ -0,0 +1,5 @@
+import { get } from "../utils/ajax";
+
+export function fetchPackages (siteId: string) {
+  return get(`/studio/api/2/publish/packages?siteId=${siteId}`, {'X-XSRF-TOKEN': '060f063c-7812-4426-abfa-a1169d1e300c'})
```
rart: Move the mockup XSRF token header to the App.tsx since that file is pretty much for that sort of stuff the way we use it.

rart (changes requested):
- Use double quotes for HTML-attributes on JSX
- Is select button the confirmation dropdown? Let's call it more semantically. For example ConfirmDropdown :)

ui/app/src/services/publishing.tsx
aachavarria: using setRequestForgeryToken();

rart (approved):

PR #1329, opened 2019-10-07
Additional option to search for images in data-source `img-repository-upload`

suelje:
Old feature request: craftercms/craftercms#256

We extended the data-source img-repository-upload with an option to use search instead of browse for images.

**Detailed description**
Similar to https://github.com/craftercms/studio-ui/pull/1323, we have added an option to use search functionality instead of “Browse for uploaded image”.
img-repository-upload
Search screen is extended by a select mode:
Bildschirmfoto 2019-10-07 um 21 39 12

PR #1353, opened 2019-10-20
Extend data-sources shared-content and img-repository-upload with an option to use search

suelje:
Old feature request: https://github.com/craftercms/craftercms/issues/256
This combines https://github.com/craftercms/studio-ui/pull/1323 and https://github.com/craftercms/studio-ui/pull/1329

We extended the data-sources `shared-content` and `img-repository-upload` with an option to use search instead of browse.

**Detailed description**
Basically, in our crafter site we have page assets with slots to contain components. Components are reused on different pages and different tenants. A "slot" is managed by a datasource of type child-content. We already have a lot of components, which caused problems because all components were created in a couple of folders only. That’s why we introduced [macros](https://docs.craftercms.org/en/3.1/developers/content-modeling.html#macros-for-data-sources) for automated organizing components in folders like ‘/site/components/{yyyy}/{mm}/{dd}’
By doing so, folders are “small” enough, but now content editors had problems trying to find existing components in the child-content datasource. In order to use the browse functionality, they would need to know the exact day, when a component has been created... and content editors never know :slightly_smiling_face:
That’s why we have added an option to use search functionality in addition or instead of “Browse for existing”.
child-content

We've extended Search in Crafter by a _select_ mode and use 'browsePath' property as search folder:
search

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -1994,7 +1994,7 @@ var nodeOpen = false,
                 });
 
                 $modal.find('.bd').append(template).end().appendTo(parentEl);
-                $modal.find('.studio-ice-container').css('z-index', 100525);
+                $modal.find('.studio-ice-container').css('z-index', 1035);
 
                 parent.iframeOpener = window;
                 window.open(url, name);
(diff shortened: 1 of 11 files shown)


rart (changes requested):
As discussed, let's please rollback the _codacy_ related changes over the message bundles leaving the PR with only the _meat_ of the changes 🙂 Thanks!

rart (approved):
👍

PR #1354, opened 2019-10-21
[studio-ui] History dialog "Revert" action doesn't ask for confirmation before executing #3512

jvega190:
https://github.com/craftercms/craftercms/issues/3512

### DIFF
--- a/ui/app/src/utils/i18n-legacy.ts
+++ b/ui/app/src/utils/i18n-legacy.ts
@@ -135,13 +135,17 @@ const contentTypesMessages = defineMessages({
 
 });
 
-const commonMessages = defineMessages({
+const words = defineMessages({
+  notification: {
+    id: 'words.notification',
+    defaultMessage: 'Notification'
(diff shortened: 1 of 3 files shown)


ui/app/src/utils/i18n-legacy.ts
```diff
@@ -135,6 +135,17 @@ const contentTypesMessages = defineMessages({
...
+  confirm: {
+    id: 'commonMessages.confirm',
+    defaultMessage: 'Confirm'
+  },
+  cancel: {
+    id: 'commonMessages.cancel',
+    defaultMessage: 'Cancel'
+  }
+});
+
```
rart: Rename message bundle to words as per code autocompletion PR. 

You may equalize these two PR messages to avoid conflicts (i.e. include the strings you're using in both PRs in both PRs)

rart (changes requested):

rart (approved):

PR #1365, opened 2019-10-23
[3571] Update create site dialog to use the new API for marketplace

aachavarria:
https://github.com/craftercms/craftercms/issues/3571


avasquez614:
@rart this is blocking me so I'll merge it. if anything needs to be changed then @yureii00 can send another PR later.

PR #1369, opened 2019-10-24
[3268] Date-time-picker inside the targeting dialog.

jvega190:
https://github.com/craftercms/craftercms/issues/3268

- Added timezone selector
- Set clear on date-time selector
- Close dialog on click targeting dialog (when opened)
- Minor fixes

### DIFF
--- a/static-assets/components/cstudio-contextual-nav/targeting.js
+++ b/static-assets/components/cstudio-contextual-nav/targeting.js
@@ -328,12 +328,10 @@ CStudioAuthoring.ContextualNav.TargetingMod = CStudioAuthoring.ContextualNav.Tar
 
                       if (currentProp.default_value === '') {
                         $dateTimePicker.val(currentProp.default_value);
-                        $timeZonePicker.hide();
                       }else{
                         $dateTimePicker.datetimepicker({
                           value: new Date(currentProp.default_value)
                         });
-                        $timeZonePicker.show();
                       }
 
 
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-contextual-nav/targeting.js
```diff
@@ -15,12 +15,17 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+CStudioAuthoring.Utils.addJavascript("/static-assets/jquery/timezones/timezones.full.js");
```
rart: add conditionally, only if plugin is not already loaded

static-assets/components/cstudio-contextual-nav/targeting.js
```diff
@@ -308,14 +321,17 @@ CStudioAuthoring.ContextualNav.TargetingMod = CStudioAuthoring.ContextualNav.Tar
                       break;
                     case "datetime":
                       var $controlEl = $(controlEl),
-                          $dateTimePicker = $controlEl.find('.date-picker');
+                          $dateTimePicker = $controlEl.find('.date-picker'),
+                          $timeZonePicker = $controlEl.siblings('.timezone-container').find('.zone-picker');
 
                       if (currentProp.default_value === '') {
                         $dateTimePicker.val(currentProp.default_value);
+                        $timeZonePicker.hide();
```
rart: why hidden here? what’s the logic?

rart (changes requested):

static-assets/components/cstudio-contextual-nav/targeting.js
jvega190: When datetime selector is cleared (input empty) it hides the timezone selector as it is not needed (you can save empty value on the datetime selector)

rart: Let's leave it visible - just like the timepicker is all the time.

rart (changes requested):

rart (approved):

PR #1373, opened 2019-10-26
[3572] Update new API parameter names

aachavarria:
https://github.com/craftercms/craftercms/issues/3572

### DIFF
--- a/ui/app/src/components/CreateSiteDialog.tsx
+++ b/ui/app/src/components/CreateSiteDialog.tsx
@@ -520,12 +520,7 @@ function CreateSiteDialog() {
     };
     if (site.sandboxBranch) params.sandboxBranch = site.sandboxBranch;
     if (site.blueprintFields) params.siteParams = site.blueprintFields;
-    //TODO# remove this when change to Api2
-    let _params: any = {};
-    Object.keys(params).forEach(key => {
-      _params[underscore(key)] = params[key];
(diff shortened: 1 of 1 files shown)


ui/app/src/components/CreateSiteDialog.tsx
```diff
@@ -520,12 +520,7 @@ function CreateSiteDialog() {
     };
     if (site.sandboxBranch) params.sandboxBranch = site.sandboxBranch;
     if (site.blueprintFields) params.siteParams = site.blueprintFields;
-    //TODO# remove this when change to Api2
```
rart, line 550: What about API1 call which is still done for built in blueprints?

rart (changes requested):

ui/app/src/components/CreateSiteDialog.tsx
aachavarria: Built in blueprints are still using api1 create site

aachavarria: built in blueprint uses fn "createParams" and marketplace uses fn "createMarketplaceParams"

rart (changes requested):
Conflicts

aachavarria:
conflicts fixed

rart (approved):

PR #1374, opened 2019-10-26
[3576] Two forms are opened when clicking on a pencil on an embedded component

aachavarria:


static-assets/components/cstudio-common/common-api.js
```diff
@@ -1776,7 +1776,7 @@ var nodeOpen = false,
                     openDialogCb);
             },
 
-            performSimpleIceEdit: function (item, field, isEdit, callback, aux, isFlattenedInclude) {
+            performSimpleIceEdit: function (item, field, isEdit, callback, aux, isFlattenedInclude, display) {
```
rart: Rename to _openHidden_ encapsulate the logic of the visibility property. Give the param a default fallback value to false for backward compatibility.

rart (changes requested):

rart (approved):

PR #1375, opened 2019-10-28
[2471]Add Publishing Queue tab to Site Config > Publishing

aachavarria:
https://github.com/craftercms/craftercms/issues/2471


ui/app/src/components/PublishingQueue.tsx
```diff
@@ -133,15 +132,21 @@ const useStyles = makeStyles((theme: Theme) => ({
   }
 }));
 
-const currentFiltersInitialState:CurrentFilters = {
+const READY_FOR_LIVE = 'READY_FOR_LIVE';
```
rart: Move constants to `ui/app/src/models/publishing.ts`. Read them from there on PublishingPackage component too instead of sending as props.

rart (changes requested):

rart (approved):

PR #1376, opened 2019-10-29
[3595] Form control date-time to support inputs in 24 hour clock

suelje:
https://github.com/craftercms/craftercms/issues/3595


rart:
We found a couple of exceptions:
- When time was set to 00:00
- When time was greater than 12 in "PM" mode

We pulled your PR and worked on top of it here: https://github.com/craftercms/studio-ui/pull/1385

Since this one incorporates your PR, I'm closing this one 🙂 

Thanks for the contribution!

rart:
Actually, can you pull and verify that the fixes work as expected for you on your use case & env?

PR #1379, opened 2019-10-29
[3590, 3596] Create group view - no error message showing, display limit group name size

jvega190:
https://github.com/craftercms/craftercms/issues/3590 - [studio-ui] Create Group does not display error message when creating a duplicate group 
https://github.com/craftercms/craftercms/issues/3596 - [studio-ui] Display character limit for group display name when creating group


rart (changes requested):
Rename and describe PR accordingly

rart (approved):

PR #1394, opened 2019-11-04
[3593] Encapsulate Require.js in Studio's ICE support

jvega190:
https://github.com/craftercms/craftercms/issues/3593


rart (changes requested):
Take care of discussed cache.js changes

rart (approved):

PR #1402, opened 2019-11-05
[3610] Form shrinks after uploading a file when ICE is used to edit an RTE

aachavarria:
https://github.com/craftercms/craftercms/issues/3610


rart (commented):
finished?

aachavarria:
I need to check with vero first

rart (approved):

PR #1411, opened 2019-11-06
[3617] CMIS browse should show a progress bar when clonning an asset

jvega190:
https://github.com/craftercms/craftercms/issues/3617


static-assets/components/cstudio-browse/browseCMIS.js
```diff
@@ -282,6 +282,24 @@
                 }
             }
             CStudioAuthoring.Service.contentCloneCMIS(paramsJson, callbackContent);
+
+            const dialogContent = `<div class="loading-bar">
```
rart: Rubbish thing about using template quotes is that the string ends up with all that tab space on subsequent lines... Alternatively you could
- do array and join at the end 
- simple string concat
- consts/function declared at the top of the file with minimal indentation - In this case there's no interpolation so even simple consts would do.

To keep spaces on arrays, simply put spaces at the top of the string, or use block comments to create indentation.

static-assets/components/cstudio-browse/browseCMIS.js
```diff
@@ -282,6 +282,24 @@
...
+                                     <div class="loading-bar-inner" style="height: 3px; width: 100%; background-color: #2275d7;"/>
+                                   </div>
+                                   Cloning CMIS Asset
```
rart: Translate

jvega190: done!

static-assets/components/cstudio-browse/browseCMIS.js
jvega190: done, updated to string concat

ui/scss/util/_util.scss
```diff
@@ -397,3 +397,70 @@ a.no-decoration,
 .cb {
   clear: both;
 }
+
+@mixin cstudioLoadingBar($height, $color) {
+  .cstudio__loading-bar {
```
rart: The class inside of the mixin reduces it's reusability. Let the consumer set the class and the mixin simply adds the parameterized content

ui/scss/util/_util.scss
```diff
@@ -397,3 +397,70 @@ a.no-decoration,
...
+    height: $height;
+    position: relative;
+    background: #555;
+    box-shadow: inset 0 -1px 1px rgba(255, 255, 255, 0.3);
+
+    > span {
+      display: block;
+      height: 3px;
+      //background-color: #2275d7;
```
rart: X

rart (changes requested):

ui/app/src/utils/i18n-legacy.ts
```diff
@@ -483,6 +483,17 @@ const publishingMessages = defineMessages({
   }
 });
 
+const browseMessages = defineMessages({
+  cmis: {
+    id: 'browse.cmis',
```
rart: Name these ids more specifically. e.g. `browseCMIS`

rart (approved):

PR #1414, opened 2019-11-07
[3643] add support for multiple excludes to wcm-root-folder

jvega190:
https://github.com/craftercms/craftercms/issues/3643

### DIFF
--- a/static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
+++ b/static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
@@ -187,15 +187,17 @@
 
 
           if(config.params.excludes) {
-            if ( (typeof(config.params.excludes) == "object")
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
```diff
@@ -187,15 +187,17 @@
 
 
           if(config.params.excludes) {
-            if ( (typeof(config.params.excludes) == "object")
-              && (typeof(config.params.excludes.exclude) != "array")) {
-              if (config.params.excludes.exclude != undefined) {
-                var path = config.params.excludes.exclude;
+            if ( (typeof(config.params.excludes) == "object") && (config.params.excludes.exclude)) {
```
rart: Looks like the else is for a case you're now handling on the if too.

rart (changes requested):

rart (approved):

PR #1426, opened 2019-11-12
[#3530] Password requirements should be displayed when password requirements is enabled

vestrada:
https://github.com/craftercms/craftercms/issues/3530 - [studio-ui] Password requirements should be displayed when password requirements is enabled

### DIFF
--- a/static-assets/scripts/main.js
+++ b/static-assets/scripts/main.js
@@ -36,7 +36,8 @@
         passwordRequirementMessages = i18n.messages.passwordRequirementMessages,
         globalConfigMessages = i18n.messages.globalConfigMessages,
         words = i18n.messages.words,
-        usersAdminMessages = i18n.messages.usersAdminMessages,;
+        usersAdminMessages = i18n.messages.usersAdminMessages,
+        settingsMessages = i18n.messages.settingsMessages;
 
   app.run([
     '$rootScope', '$state', '$stateParams', 'authService', 'sitesService', 'Constants', '$http', '$cookies', '$location',
@@ -866,6 +867,10 @@
       $scope.crafterLogo = Constants.CRAFTER_LOGO;
       $scope.messages = {
         fulfillAllReqErrorMessage: formatMessage(passwordRequirementMessages.fulfillAllReqErrorMessage),
+        password: formatMessage(settingsMessages.password),
+        currentPassword: formatMessage(settingsMessages.currentPassword),
+        isRequired: formatMessage(settingsMessages.isRequired),
+        mustMatchPreviousEntry: formatMessage(settingsMessages.mustMatchPreviousEntry)
       };
 
       if($location.$$search.iframe){
--- a/static-assets/scripts/main.js
+++ b/static-assets/scripts/main.js
@@ -36,8 +36,7 @@
         passwordRequirementMessages = i18n.messages.passwordRequirementMessages,
         globalConfigMessages = i18n.messages.globalConfigMessages,
         words = i18n.messages.words,
-        usersAdminMessages = i18n.messages.usersAdminMessages,
-        settingsMessages = i18n.messages.settingsMessages;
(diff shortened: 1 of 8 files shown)


static-assets/scripts/main.js
```diff
@@ -747,34 +750,66 @@
...
+            ['']
+          )[0].match(/{(.*?)}/) ||
+          ['', '']
+        )
+      )[1].split(',')[1];
+
+      const minLength = (
+        (
+          (
+            generalRegExp.match(/\(\?<minMaxLength>(.*){(.*?)}\)/) ||
```
rart, line 765: wrong key

static-assets/scripts/main.js
```diff
@@ -747,34 +750,66 @@
...
+            ['']
+          )[0].match(/{(.*?)}/) ||
+          ['', '']
+        )
+      )[1].split(',')[0];
+
+      const maxLength = (
+        (
+          (
+            generalRegExp.match(/\(\?<minMaxLength>(.*){(.*?)}\)/) ||
```
rart, line 775: wrong key

static-assets/scripts/main.js
```diff
@@ -747,34 +750,66 @@
...
+        )
+      )[1].split(',')[1];
+
       const messages = {
         hasNumbers: formatMessage(passwordRequirementMessages.hasNumbers),
         hasLowercase: formatMessage(passwordRequirementMessages.hasLowercase),
         hasUppercase: formatMessage(passwordRequirementMessages.hasUppercase),
         hasSpecialChars: formatMessage(passwordRequirementMessages.hasSpecialChars, { chars: (allowedChars ? `(${allowedChars})` : '') }),
         noSpaces: formatMessage(passwordRequirementMessages.noSpaces),
         minLength: formatMessage(passwordRequirementMessages.minLength, { min }),
```
rart, line 788: you renamed var min above so it no longer exists. 
bear in mind the min message prop needs to be called min and the minmax is called minLength

rart: create separate vars and regex processing for each. you can’t mix to get min and max separately as they are not guaranteed to be together

rart (changes requested):

static-assets/scripts/main.js
vestrada: Hi Roy, 
I'm not sure if I'm understanding this changes good. I didn't remove min and max, I add other variables to support maxMinLength.

rart: Got it 👍🏻

ui/scss/user-dashboard/_admin.scss
```diff
@@ -242,25 +242,39 @@ table {
   }
 }
 
-.user-management-dialog .popover {
-  max-width: 500px;
```
rart, line 246: Without this rule, the popover size forced wrapping too early

rart: Doesn’t look like it is anywhere else in the new set of styles

rart (changes requested):

ui/app/src/utils/i18n-legacy.ts
```diff
@@ -154,6 +154,25 @@ const groupsAdminMessages = defineMessages({
   }
 });
 
+const settingsMessages = defineMessages({
```
rart: Rename to profileSettingsMessages. Top var and ids.

rart (changes requested):

ui/app/src/utils/i18n-legacy.ts
vestrada: Done

ui/scss/user-dashboard/_admin.scss
vestrada: Done

rart (approved):

PR #1458, opened 2019-11-21
[3688] RTE 5 doesn't strip script tags on save, only on load

jvega190:
https://github.com/craftercms/craftercms/issues/3688


jvega190:
closing in order to create it as draft (change for 3.1.5)

PR #1459, opened 2019-11-21
[3688] RTE 5 doesn't strip script tags on save, only on load #1458

jvega190:
https://github.com/craftercms/craftercms/issues/3688


sumerjabri:
@jvega190 what's this waiting for? (why draft)

jvega190:
@sumerjabri sorry it is ready for review, updated to regular PR

PR #1471, opened 2019-11-26
[3699] Page preview does not work after previewing an image

jvega190:
https://github.com/craftercms/craftercms/issues/3699

### DIFF
--- a/static-assets/scripts/host.js
+++ b/static-assets/scripts/host.js
@@ -620,7 +620,7 @@
     var path = hash.page,
       hashPage = hash.page;
 
-    if (path.match('^((\/static-assets)|(\/remote-assets)|(\/api))')) {
+    if (path.match(/^((\/static-assets)|(\/remote-assets)|(\/api))/)) {
       hasCheckIn = false;
     }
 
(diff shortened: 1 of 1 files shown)


static-assets/scripts/host.js
```diff
@@ -609,17 +609,21 @@
...
         base: previewAppBaseUri,
         url: hash.page
       });
-      }
+    }
 
     var path = hash.page,
       hashPage = hash.page;
 
+    if (path.startsWith('/static-assets/')) {
```
rart: instead of startsWith, match against `^((\/static-assets)|(\/remote-assets)|(\/api))`

static-assets/scripts/host.js
```diff
@@ -609,17 +609,21 @@
...
         base: previewAppBaseUri,
         url: hash.page
       });
-      }
+    }
 
     var path = hash.page,
       hashPage = hash.page;
 
+    if (path.match('^((\/static-assets)|(\/remote-assets)|(\/api))')) {
```
rart: Send directly as a reg ex literal (not a string).

rart (changes requested):

static-assets/scripts/host.js
rart: `/^((\/static-assets)|(\/remote-assets)|(\/api))/`

rart (approved):

PR #1474, opened 2019-11-28
[3704] image view for file names with umlauts doesn't work

jvega190:
https://github.com/craftercms/craftercms/issues/3704


static-assets/components/cstudio-common/common-api.js
```diff
@@ -5079,8 +5079,8 @@ var nodeOpen = false,
              */
             lookupContentItem: function(site, path, callback, isDraft, populateDependencies) {
 
-                path = unescape(path);
-                var serviceUri = this.lookupContentItemServiceUri + "?site=" + site + "&path=" + escape(path);
+                path = decodeURI(path);
```
rart, line 5085: Remove line. And use without encoding on next line.

rart (changes requested):

static-assets/components/cstudio-common/common-api.js
jvega190: explanation comment added

rart (approved):

PR #1475, opened 2019-11-28
[3706] Numeric Input control doesnt save default value unless its focused

jvega190:
https://github.com/craftercms/craftercms/issues/3706

### DIFF
--- a/static-assets/components/cstudio-forms/controls/numeric-input.js
+++ b/static-assets/components/cstudio-forms/controls/numeric-input.js
@@ -110,9 +110,9 @@ YAHOO.extend(CStudioForms.Controls.numericInput, CStudioForms.CStudioFormField, 
         max = el.maxValue,
 			  min = el.minValue;
 
-		if(max && max !== '' && parseInt(this.value) > max){
+		if(max != null && max !== '' && parseInt(this.value) > max){
 			this.value = max;
-		} else if(min && min !== '' && parseInt(this.value) < min){
+		} else if(min != null && min !== '' && parseInt(this.value) < min){
 			this.value = min;
 		}
 
--- a/static-assets/components/cstudio-forms/controls/numeric-input.js
+++ b/static-assets/components/cstudio-forms/controls/numeric-input.js
@@ -106,11 +106,11 @@ YAHOO.extend(CStudioForms.Controls.numericInput, CStudioForms.CStudioFormField, 
 	 * @param el element
 	 */
 	count: function(evt, countEl, el) {
-		var max = this.getAttribute("max") ,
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-forms/controls/numeric-input.js
```diff
@@ -106,11 +106,11 @@ YAHOO.extend(CStudioForms.Controls.numericInput, CStudioForms.CStudioFormField,
...
 	count: function(evt, countEl, el) {
-		var max = this.getAttribute("max") ,
-			min = this.getAttribute("min");
-		if(max && parseInt(this.value) > max){
+		var max = el.maxValue,
+			  min = el.minValue;
+		if(max && max !== '' && parseInt(this.value) > max){
 			this.value = max;
-		} else if(min && parseInt(this.value) < min){
+		} else if(min && min !== '' && parseInt(this.value) < min){
```
rart: This will be false on 0. Is that a possibility? is that the intended behaviour?

static-assets/components/cstudio-forms/controls/numeric-input.js
```diff
@@ -106,11 +106,11 @@ YAHOO.extend(CStudioForms.Controls.numericInput, CStudioForms.CStudioFormField,
 	 * @param el element
 	 */
 	count: function(evt, countEl, el) {
-		var max = this.getAttribute("max") ,
-			min = this.getAttribute("min");
-		if(max && parseInt(this.value) > max){
+		var max = el.maxValue,
```
rart: Where do these values come from and why is it a good idea to replace the former getattr call?

rart (changes requested):

static-assets/components/cstudio-forms/controls/numeric-input.js
jvega190: Those values (maxValue/minValue) belong to the element, from defined supportedProperties. The `getAttribute` calls were failing since they're not defined (maybe a copy/paste from somewhere else?).

static-assets/components/cstudio-forms/controls/numeric-input.js
jvega190: It's a bug, 0 must be considered. Fixing

jvega190: @rart I realized that the 0 comparison won't be false, because at that moment it's a string

rart: (0 would be false but if it's a string then it's not `0` 😉)

rart (approved):

PR #1488, opened 2019-12-12
add tmpl and htm to ediable assets

phuongnq:
Currently user cannot edit static-assets with extension of .tmpl or .htm
This PR fixes so that user can edit assets with such extension


rart (approved):
Hi, @phuongnq 🙂 Thanks for the contribution. This looks good from the menu stand point however this will cause a couple of 404s as the code editor will not find a mode for these types of files. I'll create a separate ticket for that as I realise we have some other cases where this sort of happens too.

PR #1492, opened 2019-12-12
[3727] Tree highlight bugs

jvega190:
https://github.com/craftercms/craftercms/issues/3727
 
- For the issue when scrolling content (that was causing wrong positioning of highlight), now it is getting the position of the item relative to the window, so it won't change if content is scrolled.

### DIFF
--- a/static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
+++ b/static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
@@ -989,7 +989,7 @@
 
         $treeParent.on('mouseenter', '.ygtvcell', function() {
           const target = $(this).find('.treenode-label')[0], // it's always one item (label)
-            top =  $(this).closest('.ygtvitem').offset().top - 48;
+                top = $(this).closest('.ygtvitem')[0].getBoundingClientRect().top - 48;
 
           $contextMenuEllipsis.show();
           $contextMenuEllipsis.attr('data-tree', tree.id);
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-contextual-nav/wcm-site-dropdown-mods/wcm-root-folder.js
```diff
@@ -1015,6 +1015,11 @@
           $sidebarHighlight.hide();
         });
 
+        $('#acn-dropdown-menu').on('scroll', function(){
```
rart, line 1018: Currently if you scroll the doc, the highlight and ellipsis gets mispositioned. Will this work if the non-sidebar area gets scrolled?

jvega190: This event fixes the issue when scrolling in the tree area, for the other issue (scrolling on the doc where the highlight gets mispositioned there's another fix, where the retrieval of the position of the element being highlighted is relative to the window instead of the doc. Please check this video: https://www.loom.com/share/a938978368d543edacefed80427dea1b

rart: I can see you're directly scrolling the sidebar. We need to test scrolling the main document without leaving the sidebar.

When you continue to scroll past the scroll on the sidebar (at least o mac/chrome) it will start scrolling the main document. Please check if things behave as they should in that case.

Let me know if that works.

jvega190: Tested that scenario, when scrolling the main document (past the scroll on the sidebar) highlight doesn't break

rart (approved):

PR #1504, opened 2019-12-19
[3632] Create the new assets panel

aachavarria:
https://github.com/craftercms/craftercms/issues/3632


ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -27,12 +44,104 @@ const translations = defineMessages({
...
+  }, [activeSite]);
+
   return (
     <ToolPanel title={translations.assetsPanel}>
-      <Typography component="h2" variant="subtitle1" style={{ padding: '10px' }}>
-        Assets Panel
-      </Typography>
+      <div className={classes.assetsPanelWrapper}>
+        <SearchBar onChange={() => {
+        }} keyword={''} classes={{ root: classes.search }}/>
```
rart: Format

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -27,12 +44,104 @@ const translations = defineMessages({
...
+    marginTop: 0,
+    alignSelf: 'inherit'
+  },
+  root: {
+    padding: '0 16px',
+    height: '70px'
+  }
+}));
+
+function AssetComponent(props: any) {
```
rart: Use same card styles/structure from main search results page minus the checkbox and the options menu will have different options. Refactor search component as appropriately for card re-usage.

ui/app/src/components/SearchBar.tsx
```diff
@@ -84,7 +87,7 @@ export default function SearchBar(props: SearchBarProps) {
   const [focus, setFocus] = useState(false);
   const {formatMessage} = useIntl();
   return (
-    <div className={clsx(classes.search, focus && 'focus')}>
+    <div className={clsx(classes.search, focus && 'focus', props.classes?.root && props.classes.root)}>
```
rart: `props.classes?.root` would be enough (no need for the && exp)

ui/app/src/components/SearchBar.tsx
```diff
@@ -76,6 +76,9 @@ interface SearchBarProps {
   autofocus?: boolean;
   backgroundColor?: string;
   placeholder?: string;
+  classes?: {
+    root: any;
```
rart: All inner props of `classes` should also be optional (`?`)

ui/app/.env
```diff
@@ -1,3 +1,3 @@
-# HOST=127.0.0.1
+HOST=authoring.sample.com
```
rart: Do not commit

ui/app/src/components/SearchBar.tsx
```diff
@@ -84,15 +87,15 @@ export default function SearchBar(props: SearchBarProps) {
...
+    <div className={clsx(classes.search, focus && 'focus', props.classes?.root)}>
       <SearchIcon className={classes.searchIcon}/>
       <InputBase
         onChange={e => onChange(e.target.value)}
         onFocus={() => setFocus(true)}
         onBlur={() => setFocus(false)}
         placeholder={placeholder || formatMessage(messages.placeholder)}
         autoFocus={autofocus}
-        value={keyword}
+        value={keyword !== 'undefined' && keyword}
```
rart: did you mean `typeof keyword`?

You need to always supply a value or react will whine about switching input between controlled/uncontrolled

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,207 @@
...
+  const onDragStart = (mediaItem: MediaItem) => hostToGuest$.next({
+    type: ASSET_DRAG_STARTED,
+    payload: mediaItem
+  });
+
+  const onDragEnd = () => hostToGuest$.next({
+    type: ASSET_DRAG_ENDED
+  });
+
+  setRequestForgeryToken();
```
rart: Should be able to delete this now, no?

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,207 @@
...
+    cursor: 'move',
+    marginBottom: '16px',
+  },
+  pagination: {
+    marginLeft: 'auto',
+    position: 'fixed',
+    bottom: 0,
+    background: 'white',
+    color: 'black',
+    width: '239px',
```
rart: There’s a DRAWER_WIDTH somewhere around there. In any case, preferably use a calc or percentage.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
aachavarria: I think no, if I delete it does not work

ui/app/src/components/SearchBar.tsx
aachavarria: react only complains when the value is null, right now is it's not throwing errors

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
rart: Please check why. This is set on the `CrafterCMSNextBridge.tsx`, which is always the root of the app. 

Are you on `/studio` when using the Webpack server?

ui/app/.env
rart: DO NOT COMMIT

rart: Check back here on Github's PR "Files changed" tab to make sure that you've successfully reverted and removed from commit.

ui/app/src/components/MediaCard.tsx
```diff
@@ -119,68 +113,53 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    headerButtonIcon: HeaderButtonIcon = MoreVertRounded,
+    handleHeaderButtonClick,
+    avatar: Avatar,
+    onDragStart,
+    onDragEnd
+  } = props;
+  const { name, path, type } = item;
+  const isList = currentView === 'list';
 
   const renderIcon = (type: string, path: string, name: string) => {
```
rart, line 164: Extract out as a component. You may send `classes` as a prop.

ui/app/src/components/MediaCard.tsx
```diff
@@ -119,68 +113,53 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  const {
+    handleEdit,
+    handlePreview,
+    handlePreviewAsset,
+    handleSelect,
+    selected,
+    item,
+    previewAppBaseUri,
+    currentView = 'grid',
+    hasCheckbox = true,
```
rart: Can we remove this prop? If the handleSelect prop is not supplied, then there's not checkbox. Also, rename all even handling-props to `on{Action}` (e.g. onEdit, onPreview, onSelect, etc)

rart: shift + f6 (shift+fn+f6) on IntelliJ for assisted rename. 

You'll get something like `handleEdit: onEdit`. Then, manually edit the props to the new name by removing the old name so it ends up as just `onEdit` (~handleEdit:~ onEdit => onEdit)

ui/app/src/components/SearchBar.tsx
```diff
@@ -84,15 +87,15 @@ export default function SearchBar(props: SearchBarProps) {
   const [focus, setFocus] = useState(false);
   const {formatMessage} = useIntl();
   return (
-    <div className={clsx(classes.search, focus && 'focus')}>
+    <div className={clsx(classes.search, focus && 'focus', props.classes?.root)}>
```
rart, line 91: Doesn't seem like the focus class has any styles associated to it. Remove?

ui/app/src/components/SearchBar.tsx
```diff
@@ -84,15 +87,15 @@ export default function SearchBar(props: SearchBarProps) {
...
       <SearchIcon className={classes.searchIcon}/>
       <InputBase
         onChange={e => onChange(e.target.value)}
         onFocus={() => setFocus(true)}
         onBlur={() => setFocus(false)}
         placeholder={placeholder || formatMessage(messages.placeholder)}
         autoFocus={autofocus}
-        value={keyword}
+        value={typeof keyword !== 'undefined' ? keyword : undefined}
```
rart: When will keyword be undefined?

React will complain as it switches from controlled<=>uncontrolled inputs. 

useState calls that control this component should be initialised with `''` rather than undefined.

ui/app/src/models/Card.ts
```diff
@@ -0,0 +1,24 @@
+/*
```
rart: This name is way too generic for it's purpose. Do even need a separate file for this?

ui/app/src/models/Search.ts
```diff
@@ -16,12 +16,12 @@
  */
 
 export interface ElasticParams {
-  query: string;
+  query?: string;
   keywords: string;
   offset: number;
   limit: number;
```
rart: Get these by extending from PaginationOptions

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -175,15 +184,19 @@ export function PreviewConcierge(props: any) {
...
 
+    // Retrieve all content types in the system
+    (!contentTypes && site) && dispatch(fetchContentTypes());
+    contentTypes && contentTypes$.next(contentTypes);
+
     let fetchSubscription;
     switch (selectedTool) {
       case 'craftercms.ice.assets':
-        // TODO: aaron to fetch assets here...
+        (assets.isFetching === null && site) && dispatch(fetchPanelAssetsItems(initialSearchParameters));
```
rart: When switching sites, are you resetting the state so the isFetching is null? Test site switching works as expected. The preview toolbar switcher is great for testing this.

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -225,6 +238,15 @@ export function PreviewConcierge(props: any) {
 
 }
 
+const initialSearchParameters: ElasticParams = {
```
rart: Get this from state. Remove from here.

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -175,15 +184,19 @@ export function PreviewConcierge(props: any) {
...
-        // TODO: aaron to fetch assets here...
+        (assets.isFetching === null && site) && dispatch(fetchPanelAssetsItems(initialSearchParameters));
         break;
       case 'craftercms.ice.components':
         // Retrieve all content types in the system
-        (!contentTypes && site) && dispatch(fetchContentTypes());
-        contentTypes && contentTypes$.next(contentTypes);
+        // (!contentTypes && site) && dispatch(fetchContentTypes());
+        // contentTypes && contentTypes$.next(contentTypes);
         break;
```
rart: x

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,232 @@
...
+import DragIndicatorRounded from '@material-ui/icons/DragIndicatorRounded';
+import EmptyState from "../../../components/SystemStatus/EmptyState";
 
 const translations = defineMessages({
   assetsPanel: {
     id: 'craftercms.ice.assets.title',
     defaultMessage: 'Assets'
+  },
+  itemsPerPage: {
+    id: 'search.itemsPerPage',
```
rart: Is this the exact same text and id from the search page?

Make sure they are the same or use a unique id for the Panel

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,232 @@
...
 });
 
+const initialSearchParameters: ElasticParams = {
+  keywords: '',
+  offset: 0,
+  limit: 10,
+  filters: {
+    'mime-type': ['image/png', 'image/jpeg', 'image/gif', 'video/mp4', 'image/svg+xml']
+  }
+};
```
rart: Don't need this. It's on the state.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,232 @@
...
     </ToolPanel>
   );
 }
+
+
+export function AssetsPanelUI(props) {
+  const {
+    classes,
+    assetsResource,
+    handleChangePage,
```
rart: use `onPageChanged`. (Follow the `on{ACTION}` convention.)

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,232 @@
...
+        labelRowsPerPage=""
+        count={total}
+        rowsPerPage={query.limit}
+        page={pageNumber}
+        backIconButtonProps={{
+          'aria-label': 'previous page',
+        }}
+        nextIconButtonProps={{
+          'aria-label': 'next page',
+        }}
```
rart: Translate

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,232 @@
...
+                onDragEnd={onDragEnd}
+              />
+            )
+          }
+        )
+      }
+      {
+        total === 0 &&
+        <EmptyState title={formatMessage(translations.noResults)}
+                    classes={{ image: classes.noResultsImage, title: classes.noResultsTitle }}/>
```
rart: Format

ui/app/src/modules/Search/Search.tsx
```diff
@@ -266,14 +286,12 @@ function Search(props: SearchProps) {
...
                 handleEdit={handleEdit}
-                handleDelete={handleDelete}
                 handlePreview={handlePreview}
                 handlePreviewAsset={handlePreviewAsset}
                 handleSelect={handleSelect}
                 selected={selected}
-                onGetUserPermissions={onGetUserPermissions}
-                mode={mode}
                 previewAppBaseUri={previewAppBaseUri}
+                handleHeaderButtonClick={handleHeaderButtonClick}
```
rart: Rename all `handle{...}` to `on{...}`

ui/app/src/modules/Search/Search.tsx
```diff
@@ -607,6 +657,37 @@ function Search(props: SearchProps) {
           {renderPreview(preview)}
         </div>
       </Dialog>
+      <Menu
+        id="options-menu"
```
rart: remove id

ui/app/src/services/content.ts
```diff
@@ -395,8 +395,25 @@ function writeContentUrl(qs: object) {
...
+  return getDOM(site, modelId).pipe(
+    switchMap((doc) => {
+      const qs = {
+        site,
+        path: modelId,
+        unlock: 'true',
+        fileName: getInnerHtml(doc.querySelector('file-name'))
+      };
+
+      let fieldNode = doc.querySelector(`:scope > ${fieldId}`);
```
rart: fieldNode may not yet exist. Check for null before setting the innerHTML. If null, you need to first append the new child and then you can set the innerHTML. Basically add an if in between the query and the assignment which body creates the element and appends to the doc.

ui/app/src/services/content.ts
```diff
@@ -410,7 +427,7 @@ export function insertComponent(
 ): Observable<any> {
   return getDOM(site, modelId).pipe(
     switchMap((doc) => {
-
+      debugger;
```
rart: x

ui/app/src/services/search.ts
```diff
@@ -16,11 +16,14 @@
...
+import { map } from "rxjs/operators";
+import { SearchResult } from "../models/Search";
 
-export function search(siteId: string, parameters: any = {}) {
+export function search(siteId: string, parameters: any = {}): Observable<SearchResult> {
   return post(`/studio/api/2/search/search.json?siteId=${siteId}`, parameters , {
     'Content-Type': 'application/json'
-  })
+  }).pipe(map(({ response }) => response.result))
 }
```
rart, line 27: 👏🏻

ui/app/src/state/actions/preview.ts
```diff
@@ -175,4 +179,20 @@ export function setItemBeingDragged(active: boolean): StandardAction {
...
+    payload: searchResult
+  }
+}
+
+export function fetchPanelAssetsItemsFailed(error): StandardAction {
+  return {
+    type: FETCH_PANEL_ASSETS_ITEMS_FAILED,
+    payload: error
+  };
+}
```
rart: You may use createAction for these too. 

(The mix of usage is mostly due to stuff that was created before we started using redux-toolkit)

ui/app/src/state/reducers/preview.ts
```diff
@@ -227,7 +242,32 @@ const reducer = createReducer<GlobalState['preview']>({
...
     return nextState;
-  }
+  },
+  [FETCH_PANEL_ASSETS_ITEMS]: (state, { payload: query }: { payload: ElasticParams }) => ({
+    ...state,
+    assets: { ...state.assets, isFetching: true, query: { ...state.assets.query, ...query } }
+  }),
+  [FETCH_PANEL_ASSETS_ITEMS_COMPLETE]: (state, { payload: searchResult }: { payload: SearchResult }) => {
+    let itemsLookupTable = createLookupTable<MediaItem>(searchResult.items, 'path');
+    let pageNumber = state.assets.query.offset / state.assets.query.limit;
```
rart: You need to Math.ceil this. Should always be an int.

rart: Check other places you may be doing this op for same issue.

ui/app/src/state/reducers/preview.ts
```diff
@@ -227,7 +242,32 @@ const reducer = createReducer<GlobalState['preview']>({
...
+        page,
+        count: searchResult.total,
+        isFetching: false,
+        error: null
+      }
+    }
+  },
+  [FETCH_PANEL_ASSETS_ITEMS_FAILED]: (state, { payload }) => ({
+    ...state,
+    assets: { ...state.assets, error: payload.response, isFetching: true }
```
rart: isFetching should be set to false here

ui/app/src/utils/hooks.ts
```diff
@@ -74,7 +74,7 @@ export function createResource<T = any>(factoryFn: () => Promise<T>) {
   return resource;
```
rart: Remove this file from your commit. There are a few updates that will take place here.

ui/guest/src/components/Guest.js
```diff
@@ -697,35 +691,21 @@ export function Guest(props) {
...
-          );
-        }
+        dropZones = [],
+        type;
 
-      });
+      if (asset.mimeType.includes('image/')) {
+        type = 'image';
+      } else if (asset.mimeType.includes('video/')) {
+        type = 'video-picker';
```
rart, line 884: what's a video-picker type?

rart (changes requested):

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
aachavarria: that should work when im on studio, but not on react app right?

ui/app/src/components/SearchBar.tsx
aachavarria: '&.focus': {
      backgroundColor: palette.white,
      boxShadow: '0px 0px 3px rgba(65, 69, 73, 0.15), 0px 4px 4px rgba(65, 69, 73, 0.15)'
    }

ui/guest/src/components/Guest.js
aachavarria: is the video asset type, I thought it was 'video' but it is named 'video-picker'

ui/app/src/models/Card.ts
aachavarria: deleted

ui/guest/src/components/GuestProxy.js
```diff
@@ -283,6 +288,26 @@ export function GuestProxy(props) {
 
           break;
         }
+        case UPDATE_FIELD_VALUE_OPERATION:
+          const { modelId, fieldId, value } = op.args;
+          const updatedField = $(`[data-craftercms-model-id="${modelId}"][data-craftercms-field-id="${fieldId}"]`);
```
rart, line 298: Repeat group items aren’t being considered

ui/guest/src/components/GuestProxy.js
```diff
@@ -283,6 +288,26 @@ export function GuestProxy(props) {
...
+          const model = contentController.getCachedModel(modelId);
+          const contentType = contentController.getCachedContentType(model.craftercms.contentType);
+          const assetType = contentType.fields[fieldId].type;
```
rart: This should be called fieldType

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
rart: I think it should work either way (that’s my point here)

ui/app/src/components/MediaCard.tsx
```diff
@@ -265,49 +232,37 @@ function MediaCard(props: MediaCardProps) {
...
-          }
-          {
-            permissions.edit === null &&
-            <MenuItem>{formatMessage(messages.loadingPermissions)}</MenuItem>
-          }
-        </Menu>
+        {
+          onHeaderButtonClick &&
+          <IconButton
+            aria-label="options"
```
rart: Translation

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,236 @@
...
+        <div className={classes.search}>
+          <SearchBar
+            onChange={handleSearchKeyword}
+            keyword={keyword}
+          />
+        </div>
+        <React.Suspense
+          fallback={
+            <LoadingState
+              title="Loading"
```
rart: Translation

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,236 @@
...
+    assetsResource,
+    onPageChanged,
+    onDragStart,
+    onDragEnd
+  } = props;
+  const { GUEST_BASE } = useSelector<GlobalState, GlobalState['env']>(state => state.env);
+  const assets: PagedEntityState<MediaItem> = assetsResource.read();
+  const { byId, count: total, query, page } = assets;
+  const pageNumber = Math.ceil(query.offset / query.limit);
+  const items = page[pageNumber];
```
rart: make this a pre-parsed prop of the resource. The resource should be nothing more than what the component needs

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,236 @@
...
+
+export function AssetsPanelUI(props) {
+  const {
+    classes,
+    assetsResource,
+    onPageChanged,
+    onDragStart,
+    onDragEnd
+  } = props;
+  const { GUEST_BASE } = useSelector<GlobalState, GlobalState['env']>(state => state.env);
```
rart: prop

ui/app/src/services/content.ts
```diff
@@ -395,8 +395,30 @@ function writeContentUrl(qs: object) {
...
+      const qs = {
+        site,
+        path: modelId,
+        unlock: 'true',
+        fileName: getInnerHtml(doc.querySelector('file-name'))
+      };
+
+      let fieldNode = doc.querySelector(`:scope > ${fieldId}`);
+      if (!fieldNode) {
+        fieldNode = document.createElement(fieldId);
```
rart: Though this works, create via `doc` rather than the browser’s document.

ui/app/src/services/content.ts
```diff
@@ -395,8 +395,30 @@ function writeContentUrl(qs: object) {
...
+        doc.append(fieldNode);
```
rart: appendChild

ui/app/src/state/reducers/preview.ts
```diff
@@ -227,7 +242,32 @@ const reducer = createReducer<GlobalState['preview']>({
...
     return nextState;
-  }
+  },
+  [FETCH_ASSETS_PANEL_ITEMS]: (state, { payload: query }: { payload: ElasticParams }) => ({
+    ...state,
+    assets: { ...state.assets, isFetching: true, query: { ...state.assets.query, ...query } }
+  }),
+  [FETCH_ASSETS_PANEL_ITEMS_COMPLETE]: (state, { payload: searchResult }: { payload: SearchResult }) => {
+    let itemsLookupTable = createLookupTable<MediaItem>(searchResult.items, 'path');
+    let pageNumber = Math.ceil(state.assets.query.offset / state.assets.query.limit);
```
rart: We’re having this calculation in many places. I’m thinking we can add the computed value to the state and just use. (i.e. `currentPage`)

rart (changes requested):

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
aachavarria: It is not sending the X-XSRF-TOKEN

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -85,7 +88,14 @@ export function PreviewConcierge(props: any) {
       switch (type) {
         case GUEST_CHECK_IN: {
 
-          hostToGuest$.next({ type: HOST_CHECK_IN, payload });
+          hostToGuest$.next({
```
rart: This approach is valid but so far we're only doing this so Guest can do the upload. Since we're working with data URLs, let's change the approach and just have Guest post the data to Host and Host would do the actual upload.

This would render this unnecessary, so for now just leave as `hostToGuest$.next({ type: HOST_CHECK_IN });`

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -159,6 +169,12 @@ export function PreviewConcierge(props: any) {
           break;
         }
         case UPDATE_FIELD_VALUE_OPERATION: {
+          const { modelId, fieldId, index, value } = payload;
```
rart: The children map has been migrated to contentController. You may now send the parent modelId for embedded components.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,346 @@
...
+    shouldReject: source => nnou(source.error),
+    errorSelector: source => source.error,
+    resultSelector: source => {
+      const items = source.page[source.pageNumber].map((id) => source.byId[id]);
+      return {
+        total: source.count,
+        limit: source.query.limit,
+        pageNumber: source.pageNumber,
+        items
+      }
```
rart: Let's keep the props consistent with the state prop.

```js
    resultSelector: source => {
       const items = source.page[source.pageNumber].map((id) => source.byId[id]);
        return {
          ...pluckProps(source, 'count', 'query.limit', 'pageNumber'),
          items
        }
```

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,346 @@
...
+
+  const onDragDrop = (e) => {
+    e.preventDefault();
+    e.stopPropagation();
+    const file = e.dataTransfer.files[0];
+    if (!file) {
+      return;
+    }
+
+    const reader = new FileReader();
```
rart, line 198: Abstract the upload logic so you can run it both from here and preview concierge for the two different use cases.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,346 @@
...
+    //it is a file from desktop??
+    if (event.dataTransfer.types.includes('Files')) {
+      setDragInProgress(true);
+    }
+  }, false);
+
+  document.addEventListener("dragleave", function (event) {
+    //needs to know if it left the page
+    //setDragInProgress(false);
+  }, false);
```
rart: 1. Check your code style settings to set single quote as the quote setting.
2. This things are side effects; they should be inside a useEfffect 😱 
3. It could be something like this:
```ts
// import fromEvent from rxjs
// ...
const elementRef = useRef();
// ...
useEffect(() => {
  const subscription = fromEvent(elementRef.current, 'dragenter').pipe(
    filter(e => e.dataTransfer?.types.includes('Files'))
  ).subscribe(() => {
    setDragInProgress(true);
  });
  return () => subscription.unsubscribe();
}, []);
useEffect(() => {
  if (dragInProgress) {
    const subscription = fromEvent(elementRef.current, 'dragend').subscribe(() => {
      setDragInProgress(false);
      // whatever else you need to do...
    });
    return () => subscription.unsubscribe();
  }
}, [dragInProgress]);
// ...
return (
  <RootElement ref={elementRef}...
```

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,346 @@
...
+                key={item.path}
+                item={item}
+                previewAppBaseUri={GUEST_BASE}
+                hasSubheader={false}
+                avatar={DragIndicatorRounded}
+                classes={{ root: classes.card }}
+                onDragStart={onDragStart}
+                onDragEnd={onDragEnd}
+              />
+            )
```
rart, line 367: Use simplified return syntax here.

ui/app/src/services/content.ts
```diff
@@ -398,8 +398,43 @@ function writeContentUrl(qs: object) {
   return `/studio/api/1/services/api/1/content/write-content.json?${qs.toString()}`;
 }
 
-export function updateField(modelId: string, fieldId: string, value: any): Observable<any> {
-  throw new Error('Not implemented.');
+export function updateField(site: string, modelId: string, fieldId: string, index: number, value: any): Observable<any> {
```
rart: Once you merge the latest updates, add the call to `updateModifiedDateElement(doc);`

ui/app/src/utils/path.ts
```diff
@@ -50,6 +50,31 @@ export function getPreviewURLFromPath(baseUrl: string, path: string) {
   return `${baseUrl}${url}`;
 }
 
+export function dataURItoBlob(dataURI) {
```
rart: - Move to string utils.
- Type the parameter.
- Rename to dataUriToBlob

ui/guest/package.json
```diff
@@ -20,6 +20,7 @@
     "react-dom": "^16.12.0",
     "react-scripts": "3.3.0-next.62",
     "rxjs": "^6.5.3",
+    "uppy": "^1.7.0",
```
rart: As mentioned, let's do the update Studio-side. So you may `yarn remove` this change (rollback).

ui/guest/src/components/Guest.js
```diff
@@ -51,6 +54,9 @@ import { DropMarker } from './DropMarker';
 import { appendStyleSheet } from '../styles';
 import { fromTopic, message$, post } from '../communicator';
 import Cookies from 'js-cookie';
+//import { XHRUpload, Core } from 'uppy';
+import Core from '@uppy/core';
+import XHRUpload from '@uppy/xhr-upload';
```
rart: Remove

ui/guest/src/components/Guest.js
```diff
@@ -659,6 +657,48 @@ export function Guest(props) {
           // TODO: Insert detached component
           break;
         }
+        case EditingStatus.UPLOAD_ASSET_FROM_DESKTOP: {
```
rart, line 642: Update to perform the actual upload studio-side

ui/guest/src/components/Guest.js
```diff
@@ -978,8 +1004,76 @@ export function Guest(props) {
           highlighted: {}
         }
       });
-    }
+    },
+
+    dragenter(e, record) {
+      if (stateRef.current.common.status === EditingStatus.LISTENING && e.originalEvent.dataTransfer.types.includes('Files')) {
```
rart: I think dataTransfer or types may be null/undefined and throw.

`e.originalEvent.dataTransfer?.types.?includes('...`

Break the if conditions each on it's own line.

ui/guest/src/components/Guest.js
```diff
@@ -1007,14 +1101,34 @@ export function Guest(props) {
...
+    }, false);
+
+    if (!elementInView) {
+      // TODO: Do this relative to the scroll position. Don't move if things are already in viewport. Be smarter.
+      let element = ElementRegistry.fromICEId(receptacles[0].id).element;
+      $(scrollElement).animate({
+        scrollTop: $(element).offset().top - 100
+      }, 300);
+    }
+  }
```
rart, line 1098: Is the bug we saw fixed?

ui/guest/src/components/Guest.js
```diff
@@ -1086,8 +1200,9 @@ export function Guest(props) {
       console.log('No Host was detected. In-Context Editing is off.');
     }, 700);
 
-    const hostDetectionSubscription = fromTopic(HOST_CHECK_IN).pipe(take(1)).subscribe(() => {
+    const hostDetectionSubscription = fromTopic(HOST_CHECK_IN).pipe(take(1)).subscribe((payload) => {
```
rart: Not needed anymore (upload will be done studio-side)

ui/guest/src/components/GuestProxy.js
```diff
@@ -285,6 +293,26 @@ export function GuestProxy(props) {
...
+        case UPDATE_FIELD_VALUE_OPERATION:
+          const { modelId, fieldId, index = 0, value } = op.args;
+          const updatedField = $(`[data-craftercms-model-id="${modelId}"][data-craftercms-field-id="${fieldId}"]`);
+          const model = contentController.getCachedModel(modelId);
+          const contentType = contentController.getCachedContentType(model.craftercms.contentType);
+          const fieldType = ContentTypeHelper.getField(contentType, fieldId).type;
+
+          if (fieldType === 'image') {
+            const tagName = updatedField.eq(index).prop('tagName');
+            if (tagName === 'IMG') {
```
rart: Let's do `.prop(tagName).toLowerCase()` and `if tagName === img`

ui/guest/src/util.js
```diff
@@ -498,3 +500,40 @@ export function addClickListener(element, type, handler) {
...
+  var elementBottom = elementTop + $(element).height();
+
+  if (fullyInView === true) {
+    return ((pageTop < elementTop) && (pageBottom > elementBottom));
+  } else {
+    return ((elementTop <= pageBottom) && (elementBottom >= pageTop));
+  }
+}
+
+export function dataURItoBlob(dataURI) {
```
rart: Name dataUriToBlob

rart (changes requested):

ui/guest/src/components/Guest.js
aachavarria: no, it is just the function scrollToReceptacle with changes to use the receptacles object

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
```diff
@@ -76,7 +76,7 @@ export function ToolPanel(props: ToolPanelProps): ReactElement | null {
     classes
   } = props;
   return (
-    <div className={classes}>
+    <div className={classes? classes: ''}>
```
rart: You can do `classes ?? ''`. 

However, I would think `className={classes}` would be ok as react would handle the nullish value, no? Why did you had to do this change?

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -234,7 +238,7 @@ export default function AssetsPanel() {
       const unmount$ = new Subject();
       fromEvent(elementRef.current, 'dragleave').pipe(
         takeUntil(unmount$)
-      ).subscribe((e: any) => {
+      ).subscribe(() => {
         clearTimeout(timeoutRef.current);
         timeoutRef.current = setTimeout(() => {
           clearTimeout(timeoutRef.current);
```
rart: Let's give this a try....
```js
useEffect(() => {
  if (dragInProgress) {
    const dragover$ = fromEvent(elementRef.current, 'dragover').pipe(
      tap((e) => {
        e.preventDefault();
        e.stopPropagation();
      })
    );
    const subscription = fromEvent(elementRef.current, 'dragleave').pipe(
      switchMap(() => interval(100).pipe(takeUntil(dragover$))),
      mapTo(false)
    ).subscribe(setDragInProgress);
    return () => {
      subscription.unsubscribe();
    };
  }
}, [dragInProgress]);
```

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
```diff
@@ -71,10 +72,11 @@ export function ToolPanel(props: ToolPanelProps): ReactElement | null {
   const {
     title,
     BackIcon,
-    onBack = () => dispatch(selectTool())
+    onBack = () => dispatch(selectTool()),
+    classes
   } = props;
   return (
-    <>
+    <div className={classes}>
```
rart: Do you really need to wrap the whole ToolPanel? Wouldn't it suffice with applying the class to the body (i.e. the section element)?

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
```diff
@@ -38,13 +38,14 @@ const useStyles = makeStyles((theme: Theme) => createStyles({
 type ToolPanelProps = PropsWithChildren<{
   title: string | MessageDescriptor;
   BackIcon?: ElementType,
-  onBack?: () => void
+  onBack?: () => void,
+  classes?: any
```
rart: Classes, as in other components should be an object with the classes for each piece of the ui. e.g. `{ header, body }`; type it.

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
```diff
@@ -38,13 +38,14 @@ const useStyles = makeStyles((theme: Theme) => createStyles({
...
 }>;
 
 interface PanelHeaderProps {
   title: string;
   BackIcon?: ElementType,
-  onBack: () => void
+  onBack: () => void,
```
rart: Check your code style settings for trailing comma to be disabled!

ui/guest/src/components/Guest.js
```diff
@@ -1212,6 +1199,27 @@ export function Guest(props) {
...
+    $(document).on('dragover', function () {
+      clearTimeout(persistence.dragLeaveTimeout);
+    });
+    $(document).on('dragleave', function () {
+      clearTimeout(persistence.dragLeaveTimeout);
+      persistence.dragLeaveTimeout = setTimeout(() => {
+        clearTimeout(persistence.dragLeaveTimeout);
+        fn.onDragEnd();
+      }, 100);
+    });
```
rart: 1. You're missing the destruction logic
2. Try the rxjs-way that I added on a prior comment for the AssetsPanel dragenter/dragleave/dragover

ui/guest/src/components/GuestProxy.js
```diff
@@ -84,7 +84,6 @@ export function GuestProxy(props) {
       .on('mouseover', '[data-craftercms-model-id]', handler)
       .on('mouseout', '[data-craftercms-model-id]', handler)
       .on('dragstart', '[data-craftercms-model-id]', handler)
-      .on('dragenter', '[data-craftercms-model-id]', handler)
       .on('dragover', '[data-craftercms-model-id]', handler)
       .on('dragleave', '[data-craftercms-model-id]', handler)
```
rart: Do we need dragleave?

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -227,39 +223,30 @@ export function PreviewConcierge(props: any) {
...
-          uppy.on('complete', (result: Core.UploadResult<any, any>) => {
-            if (result.successful.length) {
-              console.log('Upload Success');
+          uploadDataUrl(
+            site,
+            {
+              name: payload.name,
+              type: payload.type,
+              dataUrl: payload.dataUrl,
+            },
```
rart: pluckProps(payload, 'name', 'type', 'dataUrl')

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -198,27 +195,25 @@ export default function AssetsPanel() {
...
-        type: file.type,
-        data: blob,
-      });
+      uploadDataUrl(
+        site,
+        {
+          name: file.name,
+          type: file.type,
+          dataUrl: reader.result,
+        },
```
rart: pluckProps(...)

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
yureii00: because React was not handling it

aachavarria: fixed

ui/app/src/modules/Preview/PreviewConcierge.tsx
aachavarria: pending

ui/guest/src/components/Guest.js
aachavarria: pending

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -15,24 +15,382 @@
...
+  //     );
+  //     const subscription = fromEvent(elementRef.current, 'dragleave').pipe(
+  //       switchMap(() => interval(100).pipe(takeUntil(dragover$))),
+  //       mapTo(false)
+  //     ).subscribe(setDragInProgress);
+  //     return () => {
+  //       subscription.unsubscribe();
+  //     };
+  //   }
+  // }, [dragInProgress]);
```
rart: What's up with this?

ui/app/src/modules/Preview/Tools/ToolPanel.tsx
```diff
@@ -38,7 +38,10 @@ const useStyles = makeStyles((theme: Theme) => createStyles({
 type ToolPanelProps = PropsWithChildren<{
   title: string | MessageDescriptor;
   BackIcon?: ElementType,
-  onBack?: () => void
+  onBack?: () => void,
+  classes?: {
+    body: any;
```
rart: body should be optional too

ui/app/src/utils/string.ts
```diff
@@ -69,3 +69,27 @@ export default {
   dasherize,
   formatBytes
 };
+
+export function dataUriToBlob(dataURI) {
```
rart: type missing

ui/guest/src/util.js
rart: Can we remove this function from here? It's no longer used guest-side, is it?

rart (changes requested):

rart:
Please merge latest from 2019 branch from upstream so the changes you pulled from my fork don't show as changes of your PR.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
aachavarria: it is working but has a issue with the dragover

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
aachavarria: lets have a call

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -240,10 +277,9 @@ export function PreviewConcierge(props: any) {
     let fetchSubscription;
     switch (selectedTool) {
       case 'craftercms.ice.assets':
-        // TODO: aaron to fetch assets here...
+        (assets.isFetching === null && site) && dispatch(fetchAssetsPanelItems(assets.query));
```
rart: Needs to consider that there’s no error in this condition

ui/app/src/services/content.ts
```diff
@@ -398,8 +401,43 @@ function writeContentUrl(qs: object) {
...
+  parentModelId: string = null,
+  value: any
+): Observable<any> {
+  return performMutation(
+    site,
+    modelId,
+    parentModelId,
+    doc => {
+      let fieldNode;
+      if (typeof indexToUpdate === 'string') {
```
rart: This logic is not sufficient for all cases of update.

For example, see header content type. Scenario: updating social_media.icon[0]. You should probably use extractNode in either case. And it may need a little tweaking to support.

ui/guest/src/classes/ContentController.js
```diff
@@ -173,11 +172,19 @@ export class ContentController {
...
+      args: { modelId, fieldId, index, value }
     });
 
-    post(UPDATE_FIELD_VALUE_OPERATION, { modelId, fieldId });
+    post(UPDATE_FIELD_VALUE_OPERATION, {
+      modelId,
+      fieldId,
+      index,
+      value,
+      parentModelId: isNullOrUndefined(ModelHelper.prop(model, 'path'))
```
rart: This is repetitive. We should encapsulate and reuse.

ui/guest/src/components/Guest.js
```diff
@@ -971,8 +980,68 @@ export function Guest(props) {
...
+      scrollToReceptacle(validatedReceptacles);
 
+      validatedReceptacles
+        .forEach(({ id }) => {
+
+          const dropZone = ElementRegistry.compileDropZone(id);
+          dropZone.origin = false;
+          dropZones.push(dropZone);
+
+          siblings = [...siblings, ...dropZone.children];
```
rart: I don’t think siblings or children are applicable to asset drop zones.

rart (changes requested):
Check your code style settings. Set to always add semicolons and set continuation to 2 spaces as well.

See comments below

Check unresolved above (open hidden conversations).

Once you’re done with these, remove draft status.

ui/app/src/modules/Preview/PreviewConcierge.tsx
rart: What I meant was the fetch/load condition needs to consider error. Otherwise, it goes in a loop of fetching. The feedback display comes in the form of ErrorBoundary at the panel level so no need to display the snack.

ui/guest/src/classes/ContentController.js
```diff
@@ -544,6 +547,12 @@ export class ContentController {
 
 }
 
+function getParentModelId(modelId, model, models, children) {
```
rart: `model` can be retrieved by `models[modelId]` without additional costs

ui/guest/src/components/Guest.js
```diff
@@ -648,6 +639,41 @@ export function Guest(props) {
...
+              ).subscribe(function ({ payload }) {
+                if (payload.id === file.name) {
+                  contentController.updateField(
+                    record.modelId,
+                    record.fieldId[0],
+                    record.index,
+                    payload.path
+                  )
+                }
+              });
```
rart:
```js
              message$.pipe(
                filter((e) => 
                  (e.data?.type === DESKTOP_ASSET_UPLOAD_COMPLETE) &&
                  (e.data.id === file.name)
                ),
                map(e => e.data),
                take(1)
              ).subscribe(function () {
                  contentController.updateField(
                    record.modelId,
                    record.fieldId[0],
                    record.index,
                    payload.path
                  )
              });
```

rart (approved):

PR #1506, opened 2019-12-23
use fetch api instead of yui2 to POST write-content

phuongnq:
The reason is that yui2 seem has a bug on how to set headers.

I’ve tried to modify the `charset=euc-jp` but the real request didn’t update that. 

If I use `fetch`, _POST_ request actually set charset as I defined.

### DIFF
--- a/static-assets/components/cstudio-forms/template-editor.js
+++ b/static-assets/components/cstudio-forms/template-editor.js
@@ -561,15 +561,6 @@ CStudioAuthoring.Module.requireModule(
                     cancelEdit();
                   });
 
-									var saveSvcCb = {
-										success: function() {
-											modalEl.parentNode.removeChild(modalEl);
-											onSaveCb.success();
-										},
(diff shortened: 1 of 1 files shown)


rart (approved):

rart:
@phuongnq thanks for the PR!

We’re already replacing all of YUI stuff. In the case of AJAX, we use `rxjs/ajax` so we might eventually wrap this on our new centralized ajax layer (i.e. `studio-ui/ui/app/src/utils/ajax.ts`) but for now this is great!

FYI: As part of the encoding issue work we've been doing, we're probably modifying that hard coded header to a configuration-driven expression...

```'Content-Type': 'text/plain; charset=utf-8',``` 
will become =>
```'Content-Type': `text/plain; charset=${encoding}`,```

rart:
Will be merged as part of https://github.com/craftercms/studio-ui/pull/1508

PR #1517, opened 2020-01-08
Upgrading tinyMCE from 5.0.1 to 5.1.5. 

rart:
Should/can we do this upgrade now? Any concerns from anybody?

See https://www.tiny.cloud/docs/changelog for changes.

I found a very specific bug on our current v5 that’s fixed on the latest.


rart:
5.1.6 is out.

PR #1526, opened 2020-01-16
[3759] Request Publish dialog doesn't show item to publish.

jvega190:
https://github.com/craftercms/craftercms/issues/3759


jvega190:
`isValidateCommentOn` not being initialized was causing the issue

rart (approved):

PR #1550, opened 2020-02-03
[3756] Support component instance drag & drop (i.e. browse)

aachavarria:
https://github.com/craftercms/craftercms/issues/3756


rart (approved):

rart (changes requested):
Changing the component type should clear the search query.

PR #1552, opened 2020-02-03
[3627] Reactify Publish Dialog: Adjustments

jvega190:
Updates on publishDialog (submitting feedback and no scroll on dependencySelection component)
https://github.com/craftercms/craftercms/issues/3627


rart (changes requested):
Loading state/removing content is a bit much for this call. Let's just disable the UI - including buttons, add some opacity and change the button text into a spinner.

rart (approved):

PR #1556, opened 2020-02-06
SAML support

rart:


ui/app/src/models/User.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
```
sumerjabri: Fix copyright notice.

ui/app/src/services/auth.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
```
sumerjabri: Fix copyright notice.

sumerjabri (changes requested):

ui/app/src/models/User.ts
rart: Done

ui/app/src/services/auth.ts
rart: Done

sumerjabri (approved):

PR #1559, opened 2020-02-06
[3778] empty Global menu (no selection) - fix of internal routes navigation

jvega190:
https://github.com/craftercms/craftercms/issues/3778


rart:
Merge this branch with https://github.com/craftercms/studio-ui/pull/1560 and create a single PR with both updates so you can test all of this together **extensively**.

PR #1560, opened 2020-02-06
[3799] Invalid Unsaved Changes popup when language is changed in Account Settings

jvega190:
Redirection was happening before setting modified variable to false, causing the dialog to show
https://github.com/craftercms/craftercms/issues/3799

### DIFF
--- a/static-assets/scripts/main.js
+++ b/static-assets/scripts/main.js
@@ -1073,6 +1073,8 @@
           // set both cookies, on login (on user) it will get last selected
           localStorage.setItem('crafterStudioLanguage', $scope.langSelected);
           localStorage.setItem($scope.user.username + '_crafterStudioLanguage', $scope.langSelected);
+          $scope.isModified = false;
+
           let loginSuccess = new CustomEvent('setlocale', { 'detail': $scope.langSelected });
           document.dispatchEvent(loginSuccess);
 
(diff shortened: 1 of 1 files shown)


rart (approved):

rart (changes requested):
If you merge your two PRs (this one and [1559](https://github.com/craftercms/studio-ui/pull/1559)), changing the language incurs in unintended navigation. This is due to the setlocale event listener bluntly reinitializing the global nav which also navigates to the "defaultView" instead of simply replacing the labels.

Change the `setlocale` listener to call a function that sets that labels only.

```js
      function setLabels() {
        i18n = CrafterCMSNext.i18n;
        formatMessage = i18n.intl.formatMessage;
        globalMenuMessages = i18n.messages.globalMenuMessages;
        $scope.entities.forEach(function (entry, i) {
          entry.label = (
            globalMenuMessages[entry.id]
              ? formatMessage(globalMenuMessages[entry.id])
              : entry.label
          );
        });
      }

      document.addEventListener('setlocale', () => {
        setLabels();
      }, false);
```

**Test all these related concerns appropriately** - navigation, language setting, direct navigation to internal links from Preview and from sam page via nav items and manual url changes, etc.

rart:
https://github.com/craftercms/studio-ui/pull/1569

PR #1572, opened 2020-02-11
[3806] Complete community PR translations

jvega190:
https://github.com/craftercms/craftercms/issues/3806


rart (changes requested):
@jvega190 the idea of doing it on 2019 was to do it on top of the new dialogs that were reactified.

The translations updated on develop will get merged automatically on the regular develop=>2019 pulls that I do.

jvega190:
@rart So this PR should be closed, because the new dialogs don't use those texts. For scheduling we only display 'now' or 'later'

rart:
Verify copy is inline with "publish" instead of "go live" and if so, yes. Close.

jvega190:
Everything is with 'publish'. Closing PR

PR #1587, opened 2020-02-18
[3824] Dependencies dialog not visible after checking dependencies by using edit page url

aachavarria:
https://github.com/craftercms/craftercms/issues/3824


static-assets/components/cstudio-common/common-api.js
```diff
@@ -580,10 +580,11 @@ var nodeOpen = false,
                     dialogue.render(document.body);
                     dialogue.centreY();
 
+                    console.log('este??');
```
rart: x

static-assets/yui/container/container.js
```diff
@@ -5743,10 +5743,10 @@ version: 2.6.0
          * @param {Event} e The DOM event object
          */
         _onElementFocus : function(e){
-
+            var authDialog = document.getElementById('auth-monitor-dialog');
```
rart: Camelcase ID.
Missing the AuthMonitor.tsx update that adds this ID to the dialog.

Please revert all other changes in this file, leaving only the code changes. Make sure not to format. If necessary, edit the file with something other than IntelliJ so it doesn't remove trailing whitespaces and things of that nature.

static-assets/yui/container/container.js
```diff
@@ -5743,10 +5743,10 @@ version: 2.6.0
...
             var target = Event.getTarget(e);
 
-            if (target !== this.element && !Dom.isAncestor(this.element, target) && _currentModal == this) {
+            if (target !== this.element && (!authDialog && !Dom.isAncestor(authDialog, target)) && !Dom.isAncestor(this.element, target) && _currentModal == this) {
```
rart: Let's double check this condition...

This doesn't seem to cover when **there is an** auth dialog.

The original condition reads:
- Focused element is not the modal's element AND
- Focused element is not a child of the modal's element AND
- The modal is the topmost modal (current modal) THEN
- Set the focus to an element in this modal

We need to add:
```
AND... (
  There is no auth dialog OR (
    the focused element IS the auth dialog OR 
    the focused element IS a child of the auth dialog
  )
)
```

rart (changes requested):
1. See code comments
2. How did you edit the minified file?

static-assets/yui/container/container.js
rart: Actually, on second thought, the simple existence of an auth dialog should suffice to skip the focus trap.

rart (approved):

PR #1588, opened 2020-02-19
[3825] Improve default fields for page content-types

jvega190:
https://github.com/craftercms/craftercms/issues/3825


static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -1359,7 +1362,8 @@
             if (!deleteEl) {
               deleteEl = document.createElement('i');
               YDom.addClass(deleteEl, 'deleteControl fa fa-times-circle');
-              listeningEl.appendChild(deleteEl);
+
+              !defaultField && listeningEl.appendChild(deleteEl);
```
rart, line 1372: Would this not disable the elimination of the `file-name` control? (e.g. when one wants to use a auto-filename instead of a file name)

rart (changes requested):
What is that "disabled" prop that was added? How is it being used now? What happened before it existed?

static-assets/components/cstudio-dialogs/new-content-type.js
```diff
@@ -384,7 +384,35 @@ CStudioAuthoring.Dialogs.NewContentType = CStudioAuthoring.Dialogs.NewContentTyp
...
+                '</property>\r\n' +
+                '</properties>\r\n' +
+                '<constraints>\r\n' +
+                '<constraint>\r\n' +
+                '<name>required</name>\r\n' +
+                '<value>true</value>\r\n' +
+                '<type>boolean</type>\r\n' +
+                '</constraint>\r\n' +
+                '</constraints>\r\n' +
+                '</field>\r\n';
```
rart, line 414: @sumerjabri could we have a place in the system to place a default content type XML definition for pages/components and use that as a dynamic default when creating a type? Eventually, we could have a default editor too.

sumerjabri: Yes, we can pursue that down the road. This can be a requirement for FormEngine2 (I added it to the req doc).

static-assets/components/cstudio-admin/mods/content-types.js
rart: @jvega190 👆 ?

jvega190: @rart Yeah, it was part of the ticket https://github.com/craftercms/craftercms/issues/3825, it may need to be discussed

jvega190: Adding filename switcher to autofilename on components (not pages)

jvega190: @rart default auto-filename control was already there for components

rart (approved):

PR #1595, opened 2020-02-26
[3845] [studio] Add to audit log when a publishing package is cancelled

dejan-brkic:
added new audit action

https://github.com/craftercms/craftercms/issues/3845


static-assets/components/cstudio-admin/mods/encrypt-tool.js
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2007-2019 Crafter Software Corporation. All Rights Reserved.
```
sumerjabri: Fix copyright notice.

rart (changes requested):
Translations are incorrect for "CANCEL_PUBLISHING_PACKAGE" key on all but English.

sumerjabri (approved):

rart (changes requested):
`.../cstudio-admin/mods/encrypt-tool.js` - this is generated code, should not be checked in

rart (approved):

PR #1598, opened 2020-02-28
[3843] Make the `State` in the `Publishing Queue` filter multi-select

aachavarria:
https://github.com/craftercms/craftercms/issues/3843


rart:
Merged in https://github.com/craftercms/studio-ui/pull/1599

PR #1601, opened 2020-03-02
[3763] Bulk upload dialog enables Done button before all files are uploaded (before all status bars are 100%)

aachavarria:
https://github.com/craftercms/craftercms/issues/3763

### DIFF
--- a/static-assets/components/cstudio-components/dropbox.js
+++ b/static-assets/components/cstudio-components/dropbox.js
@@ -597,6 +597,7 @@ if (typeof HTMLElement != "undefined" && !HTMLElement.prototype.insertAdjacentEl
                 (Object.keys(me.manager.files).length === Object.keys(me.manager.complete).length)
               ) {
                 document.querySelector(".bulk-upload .buttons-container .close").classList.remove("disabled");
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-components/dropbox.js
```diff
@@ -593,7 +593,11 @@ if (typeof HTMLElement != "undefined" && !HTMLElement.prototype.insertAdjacentEl
                 me.manager.setCompleted(file.id);
                 me.showUploadProgress( elemProgress, 100, fileUI );
 
+              if (document.querySelector('.bulk-upload .buttons-container .close') &&
+                (Object.keys(me.manager.files).length === Object.keys(me.manager.complete).length)
+              ) {
                 document.querySelector(".bulk-upload .buttons-container .close").classList.remove("disabled");
+              }
```
rart: Adding/removing a disabled class is not enough to enable/disable a button. Set enabled/disabled accordingly.

Why do you need to check for the existence of the button?

aachavarria: because, the modal can be closed with esc btn

rart (approved):

PR #1605, opened 2020-03-02
[3852] Item selector bound to File Browse and File Uploaded From Desktop displays a wrong option in the edit Form

jvega190:
https://github.com/craftercms/craftercms/issues/3852

### DIFF
--- a/static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
+++ b/static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
@@ -16,19 +16,19 @@
 
 CStudioForms.Datasources.FileBrowseRepo= CStudioForms.Datasources.FileBrowseRepo ||
 function(id, form, properties, constraints)  {
-   	this.id = id;
-   	this.form = form;
-   	this.properties = properties;
-   	this.constraints = constraints;
-   	
-   	for(var i=0; i<properties.length; i++) {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
```diff
@@ -64,52 +64,27 @@ YAHOO.extend(CStudioForms.Datasources.FileBrowseRepo, CStudioForms.CStudioFormDa
...
-					},
-
-					failure: function() {
-					},
-
-					context: this });
-			}, createEl);
-
-			var browseEl = document.createElement("div");
-			browseEl.innerHTML = CMgs.format(langBundle, "browseExisting") + " - " + newElTitle;
```
rart, line 98: Was this not needed at all?
What was it's purpose?

rart (changes requested):

static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
jvega190: @rart The createEl div shouldn't go in this datasource, since the DS is for browsing content and not uploading (that's what createEl does). When having only the browse datasourse selected (in a control) it works as expected, but when selecting multiple it renders both browse and upload entries, which shouldn't happen.

rart (approved):

PR #1627, opened 2020-03-13
[3888] Add a new video picker control just for MediaConvert datasource

jvega190:
https://github.com/craftercms/craftercms/issues/3888


rart (approved):
Approved - needs refactor to address the code duplication. New ticket to be created.

PR #1631, opened 2020-03-16
[60-e] [studio-ui] block unauthorized roles from accessing the Site Config

dejan-brkic:
fixed issue with accessing site config consule using direct url

### DIFF
--- a/templates/web/site-config.ftl
+++ b/templates/web/site-config.ftl
@@ -17,7 +17,7 @@
 <!doctype html>
 <html xmlns="http://www.w3.org/1999/xhtml">
 
-<#if envConfig.role! != 'admin'>
+<#if !(envConfig.role! == 'admin' || envConfig.role! == 'developer')>
   <script>window.location='/studio/site-dashboard'</script>
 </#if>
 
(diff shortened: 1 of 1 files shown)


sumerjabri (approved):

russdanner (commented):
This code leaves the user looking at a blank screen. Why not redirect them as shown in the example? @dejan-brkic

templates/web/site-config.ftl
```diff
@@ -16,6 +16,11 @@
 
 <!doctype html>
 <html xmlns="http://www.w3.org/1999/xhtml">
+
+<#if !(envConfig.role! == 'admin' || envConfig.role! == 'developer')>
+  <script>window.location='/studio/site-dashboard'</script>
+</#if>
```
rart, line 22: I don't think client-side redirection is sufficient. Still allows viewing the source of the page (e.g. in chrome, you can do `view-source:localhost:8080/studio/site-config`). People could also disable JS, visit the page and reenable and start console manipulations. We should redirect server-side (e.g. freemarker util, groovy controller, headers). @sumerjabri @dejan-brkic

rart: In any case, this should be first line of the head.

rart: Here's an enhancement https://github.com/craftercms/studio-ui/pull/1636; Nevertheless, if we had a server-side mechanism we can use now, we should opt for that.

PR #1635, opened 2020-03-17
[3844] add configurable script support to tinymce5 form control - Elements whitelist

jvega190:
https://github.com/craftercms/craftercms/issues/3844


rart (approved):

rart:
@alhambrav this is an item for documentation.
@jvega190 create ticket and send sample config & default config updates. Default config could have the tag empty.

PR #1636, opened 2020-03-18
[60E] Block unauthorized roles from accessing site config view

rart:
Improve access restriction and resource obscurity


russdanner (approved):
Looks good

PR #1637, opened 2020-03-18
[1938] Viewing dependencies of a new article page with an uploaded image displays dependencies of image instead of page

jvega190:
https://github.com/craftercms/craftercms/issues/1938

### DIFF
--- a/static-assets/components/cstudio-view-controllers/dependencies.js
+++ b/static-assets/components/cstudio-view-controllers/dependencies.js
@@ -46,7 +46,6 @@
 
   function closeButtonClicked() {
     this.end();
-    $(document).off("keyup");
   }
 
   function eventsDelegation() {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-view-controllers/dependencies.js
```diff
@@ -99,13 +99,15 @@
...
     CStudioAuthoring.Service.loadItems(callback, data);
 
-    $(document).on("keyup", function (e) {
+    $(document).on('keyup', function (e) {
       if (e.keyCode === 27) {	// esc
         me.end();
-        $(document).off("keyup");
+        $(document).off('keyup');
       }
     });
```
rart: Off the handler, not the event.

rart (changes requested):
Please test generously.

rart (changes requested):
If the dialog is not closed via escape key, the listener still remains active. 
- Listen for the end event of the dialog to also destroy the event listener
- Make sure you remove the listener **before** you call on me.end on the escape listener
- Test

jvega190:
@rart I made a little refactor, I added a wrong commit message, but the commit is for the last comment you added

rart (approved):

PR #1644, opened 2020-03-23
[3907] Sequenced embedded component malformed XML upon write

aachavarria:
https://github.com/craftercms/craftercms/issues/3907


sumerjabri:
If you don't intend to have this be part of 3.1.6, then turn it into a `draft` PR or don't approve it.

rart (changes requested):
Looks like the nested node selectors are not carrying the `item-list="true"` atttribute

rart (approved):

PR #1657, opened 2020-03-27
[3873] Update legacy FormEngine to support new component receptacle mechanisms.

aachavarria:
https://github.com/craftercms/craftercms/issues/3873


ui/app/src/models/ContentType.ts
```diff
@@ -30,7 +30,8 @@ export interface ContentTypeField {
   type: string;
   sortable: boolean;
   validations: {
-    contentTypes: string[];
+    tags?: string[];
+    contentTypes?: string[];
```
rart: Don’t think these should be optional.

ui/legacy/src/components/cstudio-admin/mods/content-type-propsheet/contentTypes.js
```diff
@@ -23,7 +23,7 @@
   const STAR = '*';
 
   function createLookupTableFromCVS(cvs) {
-    return (cvs || '')
+    return !cvs ? [] : cvs
```
rart, line 26: Looks like it was correct before this change: you are expecting a string; not an array.

ui/legacy/src/components/cstudio-forms/data-sources/receptacles.js
```diff
@@ -0,0 +1,355 @@
...
+    this.type = '';
+    this.allowShared = false;
+    this.allowEmbedded = false;
+    this.enableSearch = false;
+    this.enableBrowse = false;
+    this.baseRepoPath = '/site/components';
+    this.baseBrowsePath = '/site/components';
+    //this.countOptions = 0;
+    const i18n = CrafterCMSNext.i18n;
+    this.formatMessage = i18n.intl.formatMessage;
```
rart: Extract this out of the class. Use it as a function instead of a class prop. Same for messages.

rart (changes requested):

ui/legacy/src/components/cstudio-admin/mods/content-type-propsheet/contentTypes.js
aachavarria: no because it is returning an empty array of  "", "".slice is == [""]

rart (approved):

PR #1671, opened 2020-04-03
[3930] Refactor Preview 2.0 ICE Panels to use "Suspencified" component

hungtsou:
https://github.com/craftercms/craftercms/issues/3930


ui/app/src/components/SystemStatus/Suspencified.tsx
```diff
@@ -35,6 +35,7 @@ type SuspenseWithEmptyStateProps<ResourceType = unknown> = PropsWithChildren<
 >;
 
 type SuspencifiedProps = PropsWithChildren<{
+  resource: Resource;
```
rart: x

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -166,14 +169,18 @@ export default function AssetsPanel() {
   const initialKeyword = useSelection((state) => state.preview.assets.query.keywords);
   const [keyword, setKeyword] = useState(initialKeyword);
   const [dragInProgress, setDragInProgress] = useState(false);
+  const [error, setError] = useState(null);
```
rart: Error comes from the resource. Not needed in the state at this level. You should remove the Panel error prop

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -166,14 +169,18 @@ export default function AssetsPanel() {
...
   const hostToGuest$ = getHostToGuestBus();
   const dispatch = useDispatch();
   const resource = useStateResourceSelection<AssetResource, PagedEntityState<MediaItem>>(
     (state) => state.preview.assets,
     {
       shouldRenew: (source, resource) => resource.complete,
       shouldResolve: (source) => !source.isFetching && nnou(source.page[source.pageNumber]),
-      shouldReject: (source) => nnou(source.error),
+      shouldReject: (source) => {
```
rart: revert

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -299,40 +305,73 @@ export default function AssetsPanel() {
       title={translations.assetsPanel}
       classes={dragInProgress ? { body: classes.noScroll } : null}
     >
-      <ErrorBoundary>
```
rart, line 294: No need to modify the components structure. Simply replace error boundary with the suspencified with empty state

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -299,40 +305,73 @@ export default function AssetsPanel() {
...
+          GUEST_BASE={GUEST_BASE}
+          onDragDrop={onDragDrop}
+          error={error}
+        />
+      </div>
     </ToolPanel>
   );
 }
 
+function Assets(props) {
```
rart: Unneeded if you maintain the prior structure

ui/app/src/modules/Preview/Tools/AudiencesPanel.tsx
```diff
@@ -199,20 +198,22 @@ export default function AudiencesPanel() {
...
-    <ErrorBoundary>
-      <React.Suspense
-        fallback={<LoadingState title={formatMessage(translations.audiencesPanelLoading)} />}
-      >
+    <Suspencified
+      resource={resource}
+      loadingStateProps={{
+        title: formatMessage(translations.audiencesPanelLoading)
+      }}
+      children={
```
rart: Use conventional children style (xml style child nodes)

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -471,12 +492,20 @@ export default function ContentTree() {
...
         {data.selected && (
           <>
-            <TreeItemCustom
-              nodes={data.lookupTable[data.selected]}
-              handleScroll={handleScroll}
-              handlePrevious={data.previous.length ? handlePrevious : null}
-              handleClick={handleClick}
-              handleOptions={handleOptions}
+            <Suspencified
+              resource={resource}
```
rart: x

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -471,12 +492,20 @@ export default function ContentTree() {
...
+              loadingStateProps={{
+                title: 'Loading'
+              }}
+              children={
```
rart: Use regular child style

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -131,6 +132,7 @@ interface Data {
   selected: string;
   previous: Array<string>;
   lookupTable: LookupTable<RenderTree>;
+  error: string;
```
rart: string? seems to divert from pattern

rart (changes requested):
- See comments.
- There doesn’t seem to be anything async about the content tree so Suspense probably doesn’t have purpose there and we may revert

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -256,7 +256,7 @@ function getChildren(
 }
 
 interface TreeItemCustomInterface {
-  nodes: RenderTree;
+  nodes: any;
```
rart: Incorrect type. Use Resource<RenderTree>

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -471,13 +486,15 @@ export default function ContentTree() {
       >
         {data.selected && (
```
rart: This condition is no longer required with Suspencified, is it?
remove the conditional on the handlePrevious prop and handle the logic of whether there are previous on the handler and the child component - since it has access to the resource, which contains “previous”, it can do the necessary checks.

rart (changes requested):
Please fully revert the non-changes on Suspencified

ui/app/src/modules/Preview/Tools/ContentTree.tsx
hungtsou: I have to set fallback types for this prop because it changes its type within its render. I tried to separate them on two props but by some reason it did not work and crashed the site with no error display.
changed to: 
nodes: Resource<Data> | RenderTree | any;

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -271,25 +270,13 @@ interface TreeItemCustomInterface {
 }
 
 function TreeItemCustom(props: TreeItemCustomInterface) {
-  const { resource, nodeId, handleScroll, handlePrevious, handleClick, handleOptions } = props;
+  const {resource, nodeId, handleScroll, handlePrevious, handleClick, handleOptions} = props;
   const classes = treeItemStyles({});
   const [over, setOver] = useState(false);
   let timeout = React.useRef<any>();
   const nodes = resource.read().nodeLookup[nodeId];
```
rart: This will always be a single node. Rename const to node.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -308,19 +295,35 @@ function TreeItemCustom(props: TreeItemCustomInterface) {
     return () => isMounted.current = false;
   }, []);
 
+  if (!nodes) {
```
rart: When why would a node be null?
Does moving the code chunk far from its prior line numbers have a purpose? If the nullish node is in fact a valid circumstance, returning earlier seems more efficient.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -349,18 +352,17 @@ function TreeItemCustom(props: TreeItemCustomInterface) {
       }}
     >
       {
-        nodes.children.map((node) => (
-          <TreeItemCustom key={node.id} nodeId={node.id} {...props} />
-        ))
+        nodes.children?.map(_nodeId => <TreeItemCustom {...props} key={String(_nodeId)} nodeId={String(_nodeId)}/>)
```
rart: use childNodeId as argument name
The need to cast is likely to come from improper typing

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -383,7 +384,7 @@ export default function ContentTree() {
       let parent = guest.models[guest.modelId];
       let contentType = contentTypesBranch.byId[parent.craftercms.contentType];
       let data: RenderTree = {
-        id: parent.craftercms.id,
+        id: 'root',
```
rart: Id should be the model id. This doesn’t work nor is it as discussed.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -392,38 +393,38 @@ export default function ContentTree() {
       setData({
         previous: [],
         selected: parent.craftercms.id,
-        lookupTable: createLookupTable<RenderTree>([data], 'modelId'),
-        nodeLookup: createNodesLookup([data], 'modelId')
+        nodeLookup: createNodesLookup<RenderTree>([data], 'modelId')
```
rart: Please pull from my fork and use the new hierarchicalToLookupTable object util.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -392,38 +393,38 @@ export default function ContentTree() {
...
-    //     nodeLookup: null
-    //   });
-    //   setExpanded(['root']);
-    // }
+    if (node.type === 'component' && node.id !== 'root') {
+      let model = guest.models[node.modelId];
+      let contentType = contentTypesBranch.byId[model.craftercms.contentType];
+      const {type, modelId, name, index, fieldId, parentId, embeddedParentPath} = node;
+      const nodeData = {
+        id: 'root',
```
rart: As discussed, this should be the modelId. Same for the above setData instance.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -392,38 +393,38 @@ export default function ContentTree() {
...
+        parentId,
+        embeddedParentPath
+      }
+
+      setData({
+        previous: [...data.previous, data.selected],
+        selected: model.craftercms.id,
+        nodeLookup: {
+          ...data.nodeLookup,
+          ...createNodesLookup([nodeData], 'modelId')
```
rart: Please pull from my fork and use the new hierarchicalToLookupTable object util.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -392,38 +393,38 @@ export default function ContentTree() {
...
+        }
+      });
+      setExpanded(['root']);
```
rart: Fix

ui/app/src/utils/object.ts
```diff
@@ -61,14 +61,24 @@ export function createLookupTable<T>(list: T[], idProp: string = 'id'): LookupTa
   return table;
 }
 
-export function createNodesLookup(list: any[], idProp: string = 'id'): LookupTable {
-  const table = {};
-  list.forEach((item) => {
-    table[retrieveProperty(item as any, idProp)] = item;
-    item.children?.forEach(children => {
-      table[retrieveProperty(children as any, 'id')] = children;
+export function createNodesLookup<T>(list: T[], idProp: string = 'id'): LookupTable<T> {
```
rart: Revert this whole file. Please pull from my fork and use the new hierarchicalToLookupTable object util.

rart (changes requested):
Format got borked. Your formatter is not configured with project settings.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
hungtsou: Yes, in this case I found it necessary because it seems that the component gets call with the nodeId of the selected item on child trees that are close and not render by default after the parent tree changes.

ui/app/src/modules/Preview/Tools/AssetsPanel.tsx
```diff
@@ -53,11 +52,11 @@ const translations = defineMessages({
...
   previousPage: {
-    id: 'pagination.previousPage',
+    id: 'craftercms.ice.assets.previousPage',
     defaultMessage: 'Previous page'
   },
   nextPage: {
-    id: 'pagination.nextPage',
+    id: 'craftercms.ice.assets.nextPage',
     defaultMessage: 'Next page'
   },
```
rart: Still wrong.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -27,21 +27,24 @@ import ChevronRightIcon from '@material-ui/icons/ChevronRightRounded';
...
-import { DRAWER_WIDTH, getHostToGuestBus } from '../previewContext';
+import {iconWithStrokeAndFill} from '../../../styles/icon';
+import {hierarchicalToLookupTable} from '../../../utils/object';
+import {CONTENT_TREE_FIELD_SELECTED} from '../../../state/actions/preview';
+import {DRAWER_WIDTH, getHostToGuestBus} from '../previewContext';
 import ComponentMenu from '../../../components/ComponentMenu';
+import Suspencified from '../../../components/SystemStatus/Suspencified';
+import {Resource} from '../../../models/Resource';
+
+const rootPrefix = 'root_';
```
rart: A field name on a content type may contain the substring `root_`.
Use `{root}_` as discussed. (instructions)

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -268,33 +272,44 @@ interface TreeItemCustomInterface {
 }
 
 function TreeItemCustom(props: TreeItemCustomInterface) {
-  const { nodes, handleScroll, handlePrevious, handleClick, handleOptions } = props;
+  const {node, nodeId, handleScroll, handlePrevious, handleClick, handleOptions} = props;
   const classes = treeItemStyles({});
   const [over, setOver] = useState(false);
   let timeout = React.useRef<any>();
-
+  const nodes = node.read().nodeLookup[nodeId];
```
rart: We agreed to the name `resource` for the resource param; not node.
`node` should be the name of the return value of `resource.read().lookup[nodeId]`.

As it is, is semantically incorrect (both node and nodes). The resource does not contain a node and the `read().lookup` is not a collection of nodes but one.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
rart: Don't fully understand what you mean but sounds like a buggy behaviour. Ping me to discuss.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -335,25 +350,24 @@ function TreeItemCustom(props: TreeItemCustomInterface) {
         ),
         expanded: classes.treeItemExpanded,
         group: classes.treeItemGroup,
-        iconContainer: nodes.id === 'root' ? classes.displayNone : classes.treeItemIconContainer
+        // iconContainer: nodes.id === 'root' ? classes.displayNone : classes.treeItemIconContainer
```
rart: x

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -335,25 +350,24 @@ function TreeItemCustom(props: TreeItemCustomInterface) {
...
 
 export default function ContentTree() {
   const classes = useStyles({});
   const guest = usePreviewGuest();
-  const { formatMessage } = useIntl();
+  const {formatMessage} = useIntl();
   const contentTypesBranch = useSelection((state) => state.contentTypes);
   const hostToGuest$ = getHostToGuestBus();
-  const [expanded, setExpanded] = React.useState<string[]>(['root']);
+  const [expanded, setExpanded] = React.useState<string[]>(['']);
```
rart: A blank string corresponds to no node. If the information is not available at this point, then leave empty and set in the effect that read the initial data. Make sure the rendering accounts for an empty value for expanded.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -424,17 +440,20 @@ export default function ContentTree() {
...
+    setExpanded([`${rootPrefix}${previous}`])
   };
 
   const handleChange = (event: any, nodes: string[]) => {
     if (
       event.target.classList.contains('toggle') ||
       event.target.parentElement.classList.contains('toggle')
     ) {
-      setExpanded([...nodes, 'root']);
+      setExpanded([...nodes]);
```
rart: If this had "root" before. Shouldn't it still have it?

ui/app/src/modules/Preview/ToolsPanel.tsx
```diff
@@ -275,7 +287,9 @@ export default function ToolsPanel() {
       classes={{ paper: classes.drawerPaper }}
     >
       {site ? (
-        <Tool id={toolMeta?.id} config={config}/>
+        <Suspencified loadingStateProps={{ title: `${formatMessage(translations.loading)}...` }}>
+          <Tool id={toolMeta?.id} config={config}resource={resource}/>
```
rart: missing space before `resource`

ui/app/src/modules/Preview/ToolsPanel.tsx
```diff
@@ -261,6 +265,14 @@ export default function ToolsPanel() {
   let toolMeta = tools?.find((desc) => desc.id === selectedTool);
   let config = toolMeta?.config;
 
+  const resource = useStateResource(tools, {
+    shouldRenew: () => false,
+    shouldResolve: (source) => nnou(source),
+    shouldReject: (source, resource) => resource.complete,
```
rart: Why reject if the resource is complete? Can't make sense of the logic.
Also, if there is a possibility of rejection, the errorSelector can't be null.

ui/app/src/modules/Preview/ToolsPanel.tsx
```diff
@@ -174,10 +179,10 @@ function UnknownPanel(props: any) {
   );
 }
 
-function ToolSelector() {
+function ToolSelector({ resource }) {
```
rart: Missing types/interface

rart (changes requested):
Format is still wrong on ContentTree. 
  - Did you configure your tool?
  - Did you run the formatter?

ui/app/src/modules/Preview/Tools/ContentTree.tsx
hungtsou: As we discuss the 'root' value is no longer needed.

rart (changes requested):
ContentTree format is **still** incorrect.

PR #1684, opened 2020-04-08
[3886] user should be able to pick a page via the RTE insert link

jvega190:
https://github.com/craftercms/craftercms/issues/3886


jvega190:
This is for 3.1.7/3.2.x

rart (changes requested):
Conflicts present. Easier way to solve them might be to first run prettier on the modified files since conflicts may be a product of the formatted 2019-originated file.

rart (approved):

PR #1689, opened 2020-04-13
[3901] Remove "Edit" option from context-nav when previewing videos and pdf files

jvega190:
https://github.com/craftercms/craftercms/issues/3901


rart (approved):
Is this still work in progress? It's still a draft. Make sure to reforma any files that come from the prior develop branch as those weren't formatted.

PR #1715, opened 2020-04-20
[3979][next] api/2/configuration/get_configuration_history doesn't work when default environment is set

aachavarria:
Fixing issue to use api2 with history config files;


static-assets/components/cstudio-view-controllers/history.js
```diff
@@ -462,33 +465,44 @@
...
+                      CStudioAuthoring.Service.revertContentItem(CStudioAuthoringContext.site, item, $dropdown.data('version'), {
                         success: function() {
                           if (CStudioAuthoringContext.isPreview) {
                             CStudioAuthoring.Operations.refreshPreview();
                           }
                           eventNS.data = item;
                           document.dispatchEvent(eventNS);
 
-                          _this.loadHistory(_this.selection);
+                          _this.loadConfigurationHistory(_this.selection);
```
rart, line 505: Looks like this will use config history for all files; not just config files. Does this work for non config files?

rart (changes requested):
See comments. Also, need to port this to 3.1.x and for this branch, this should be on the new dialog.

static-assets/components/cstudio-view-controllers/history.js
aachavarria: Yes if you see the code, there are two diff functions, loadConfigurationHistory, and loadHistory, config files are using loadConfigurationHistory, and the other are using loadHistory

rart (approved):

PR #1738, opened 2020-04-29
[3991][next] NumPad keys not working

jvega190:
https://github.com/craftercms/craftercms/issues/3991


static-assets/components/cstudio-admin/mods/content-type-propsheet/int.js
```diff
@@ -63,9 +61,17 @@ YAHOO.extend(
       return this.valueEl.value;
     },
 
-    isNumberKey: function(charCode) {
-      const isSubtractSign = charCode === 109 || charCode === 189 || charCode === 173;
-      return !(charCode != 43 && charCode > 31 && (charCode < 48 || charCode > 57) && !isSubtractSign);
+    isValidKey: function (evt) {
+      const key = evt.key,
```
rart: Please confirm:
- copy/paste works (keyboard & mouse)
- what if a NaN is pasted?
- ctrl+a does work? 
- Arrow keys work?
- What if you type something like 1.2.3?

I think we should validate final input instead of going crazy on the every keystroke

rart (changes requested):

static-assets/components/cstudio-admin/mods/content-type-propsheet/int.js
sumerjabri: What about the "enter" key from the Numpad (sorry if irrelevant to this particular ticket, but it was an issue in the past).

rart (approved):

PR #1755, opened 2020-05-06
[95E][next] Limit login redirects

rart:
- Validate against password requirements regexp [DONE]
- Replace login regexp on FTL template [PENDING(backend)]

### DIFF
--- a/ui/app/scripts/license.txt
+++ b/ui/app/scripts/license.txt
@@ -1,0 +1,15 @@
+/*
+ * Copyright (C) 2007-2020 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


ui/app/scripts/license.txt
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (C) 2007-2020 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
```
sumerjabri: Change the words
`GNU General Public License`
to
`GNU Lesser General Public License`

sumerjabri (changes requested):

ui/app/scripts/license.txt
rart: Done.

PR #1782, opened 2020-05-26
[3994] Cleanup legacy UI and Update to Next UI

jvega190:
https://github.com/craftercms/craftercms/issues/3994


rart (commented):
Looking good 🙂 

Is the "Request delete" case covered by the new dialog? (i.e. works for approver and non-approver types)?

jvega190:
@rart Request delete was not used anymore, there is no option to request delete for non-admin user (3.1.x)

rart (approved):

PR #1783, opened 2020-05-27
Page builder adjustments

aachavarria:
Using DefaultValues for text/html/images; 
Fixing issue with UploaderImageMask;

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -1112,6 +1112,8 @@ var nodeOpen = false,
           searchUrl += '&query=' + searchContext.query;
         }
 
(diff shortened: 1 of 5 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -1120,15 +1122,15 @@ var nodeOpen = false,
         var childSearch = null;
 
         if (
-          !searchId ||
+          searchId == '' ||
           searchId == null ||
           searchId == 'undefined' ||
```
rart, line 1127: These 3 lines are equivalent to the line you removed; hence, we can either leave `!searchId` or use `===` for these 3 checks.

PR #1789, opened 2020-05-29
[studio-ui] Include level-descriptor on ContentTypes Store #4091

aachavarria:
https://github.com/craftercms/craftercms/issues/4091


ui/guest/src/react/GuestProxy.tsx
```diff
@@ -40,7 +40,7 @@ import { GuestState } from '../store/models/GuestStore';
 import { notNullOrUndefined } from '../utils/object';
 import { forEach } from '../utils/array';
 import { popPiece, removeLastPiece } from '../utils/string';
-import { addAnimation } from '../utils/dom';
+import { addAnimation, getParentElementFromICEProps } from '../utils/dom';
```
rart: getParentElementFromICEProps => getElementByICEProps
move from dom to ElementRegistry

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -461,27 +463,22 @@ export function PreviewConcierge(props: any) {
           nou(audiencesPanel.model) &&
           nou(audiencesPanel.error)
         ) {
+          //TODO: Jose Vega re-render when this panel is opened and site is changed;
```
rart, line 466: What's the meaning of this?

ui/guest/src/utils/dom.ts
```diff
@@ -184,10 +188,10 @@ export function splitRect(rect: DOMRect, axis: string = X_AXIS): DOMRect[] {
...
 
 export function insertDropMarker({
-  $dropMarker,
-  insertPosition,
-  refElement
-}: {
+                                   $dropMarker,
+                                   insertPosition,
+                                   refElement
+                                 }: {
```
rart: Run prettier

rart (changes requested):

ui/app/src/modules/Preview/PreviewConcierge.tsx
aachavarria: this means that the audiencesPanel needs changes to support the re-render when the site changes on the address bar

ui/guest/src/react/GuestProxy.tsx
aachavarria: can't rename to getElementByICEProps, because it does not retrieve the Element Itself, it uses the index on null, so that means it uses the parent dropZone

aachavarria: also there is already a getElementByICEProps function

aachavarria: moved.

ui/app/src/modules/Preview/Tools/ContentTree.tsx
```diff
@@ -213,8 +246,10 @@ function getChildren(
 ) {
   let children = [];
   Object.keys(model).forEach((fieldName) => {
-    if (fieldName === 'craftercms') return;
-    const { type, name } = contentType.fields[fieldName];
+    if (fieldName === 'craftercms' || fieldName.endsWith('_raw')) return;
```
rart: If it endsWith _raw, it's al HTML field, we shouldn't skip.

ui/guest/src/constants.ts
```diff
@@ -70,5 +70,6 @@ export enum EditingStatus {
   EDITING_COMPONENT = 'EDITING_COMPONENT',
   EDITING_COMPONENT_INLINE = 'EDITING_COMPONENT_INLINE',
   UPLOAD_ASSET_FROM_DESKTOP = 'UPLOAD_ASSET_FROM_DESKTOP',
-  SHOW_RECEPTACLES = 'SHOW_RECEPTACLES'
+  SHOW_RECEPTACLES = 'SHOW_RECEPTACLES',
+  SELECTED_CONTENT_TREE_FIELD = 'SELECTED_CONTENT_TREE_FIELD'
```
rart: SELECTED_CONTENT_TREE_FIELD => SELECT_FIELD

ui/guest/src/utils/dom.ts
rart: > Run prettier

Format still pending

rart (changes requested):

ui/app/src/modules/Preview/Tools/ContentTree.tsx
aachavarria: im going to remove it, there is an another validation to skip the field if the contentType is not found

rart (approved):

PR #1794, opened 2020-06-02
[4099] Legacy tree - Content created under wrong path

jvega190:
https://github.com/craftercms/craftercms/issues/4099


jvega190:
@rart the issue is fixed but it needs more testing, will change it to reg PR after that

rart (approved):

PR #1813, opened 2020-06-17
[3928] Reactify History dialog

aachavarria:
https://github.com/craftercms/craftercms/issues/3928


ui/app/src/models/ContentInstance.ts
```diff
@@ -22,6 +22,7 @@ export interface ContentInstanceSystemProps {
   dateCreated: string;
   dateModified: string;
   contentTypeId: string;
+  versionNumber?: string; // TODO: remove this;
```
rart, line 25: ?

ui/app/src/modules/Content/History/CompareVersions.tsx
```diff
@@ -0,0 +1,406 @@
...
+  deleted: {
+    id: 'words.deleted',
+    defaultMessage: 'Deleted'
+  },
+  empty: {
+    id: 'words.empty',
+    defaultMessage: 'Empty'
+  },
+  noItems: {
+    id: 'compare.component.status.noItems',
```
rart: compareVersionsDialog.noItemsStatus

ui/app/src/models/ContentInstance.ts
aachavarria: that is because getVersions is returning a ContentInstance, but is needs versionNumber, the api is not ready, so im adding the versionNumber to the type in the meantime

rart (approved):

PR #1818, opened 2020-06-24
[3928] Reactify History dialog(adding Monaco)

aachavarria:
https://github.com/craftercms/craftercms/issues/3928

Adding monaco to public folder


rart (changes requested):
The idea is to add Monaco directly to static assets like other libs. Not as app assets that need to be built and bundled.

Let's see if we can pull from NPM and copy in a yarn task. That way we needn't check in the source 🙂

ui/app/public/index.html
```diff
@@ -30,8 +30,16 @@
...
   >
+  <link
+    href="/js/monaco/vs/editor/editor.main.css"
+    rel="stylesheet"
+  >
 </head>
 <body>
+<script>var require = { paths: { 'vs': '/js/monaco/vs' } };</script>
+<script src="/js/monaco/vs/loader.js"></script>
+<script src="/js/monaco/vs/editor/editor.main.nls.js"></script>
```
rart, line 41: What's this one for?

aachavarria: to load monaco editor, it needs that files

rart: yes, but what is this one specifically for? (main.nsl.js)

aachavarria: I think it contains the languages

PR #1821, opened 2020-06-24
[4144][3.1.x] Validate not available controls in content types and forms engine

jvega190:
https://github.com/craftercms/craftercms/issues/4144


rart (changes requested):
The system already tries to fetch the control. Isn't there any sort of failure callback available? Otherwise, the idea would be to do a lazy validation. Ping me to discuss.

rart (approved):

PR #1834, opened 2020-07-01
[studio-ui][3.1.x] Fix the "Postfixes" pop up is not closable - Add to all popovers

jvega190:
https://github.com/craftercms/craftercms/issues/4154

### DIFF
--- a/static-assets/scripts/main.js
+++ b/static-assets/scripts/main.js
@@ -826,7 +826,8 @@
           .focus(function () {
             let creatingPassValHTML = me.creatingPassValHTML($(this).get(0).value, staticTemplate);
             $(this).popover({
-              title: messages.passwordValidation,
+              title: `<span>${messages.passwordValidation}</span>` +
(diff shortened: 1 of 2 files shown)


static-assets/scripts/main.js
```diff
@@ -826,7 +826,8 @@
           .focus(function () {
             let creatingPassValHTML = me.creatingPassValHTML($(this).get(0).value, staticTemplate);
             $(this).popover({
-              title: messages.passwordValidation,
+              title: `<span>${messages.passwordValidation}</span>` +
+                `<button type="button" id="close" class="close fa fa-times" onclick="$(this).popover('hide');"/>`,
```
rart: Is the close id used elsewhere?

rart (changes requested):

static-assets/scripts/main.js
jvega190: No it isn't. Removing

rart (approved):

PR #1840, opened 2020-07-06
[4171][3.1.x] Quick-create dropdown with very large number of items disallows reaching all items based on screen-size

jvega190:
https://github.com/craftercms/craftercms/issues/4171


rart (approved):
@jvega190 develop needs to be tested too. The menu is now rectified so the fix — if required at all —would be different.

jvega190:
@rart already tested, working correctly in develop.

PR #1849, opened 2020-07-09
[4179][3.1.x] Studio styles for animation causing styling issues in sites using animate library

jvega190:
https://github.com/craftercms/craftercms/issues/4179


rart (changes requested):
use craftercms as the class prefix instead of studio. change both prs.

rart (approved):

PR #1855, opened 2020-07-13
[4187][3.1.x] Reject dialog overflows bottom border of the screen when too many files in rejection dialog

aachavarria:
https://github.com/craftercms/craftercms/issues/4187


rart (approved):

rart:
Please test same issue on develop with the new dialog

PR #1857, opened 2020-07-14
[4193] Deleting an invalid value from input with pattern-validation, does not remove invalid state

jvega190:
https://github.com/craftercms/craftercms/issues/4193


philipp-brock-thalia:
That was a quick win :)

rart (approved):

PR #1860, opened 2020-07-17
[4198][3.1.x] Change the character limit for username when creating a user

jvega190:
https://github.com/craftercms/craftercms/issues/4198


ui/scss/user-dashboard/_admin.scss
```diff
@@ -1114,4 +1114,8 @@ table {
 
 }
 
-
+.username.trim {
```
rart: Have a look if there’s a common sass file for all the spaces that need this to locate this. create it otherwise

rart (changes requested):

rart (changes requested):
The idea was rather to include the scss on the existing sheets so we don't have to pull in an additional sheet on the client.

jvega190:
@rart imported the styles in the `user-dashboard.scss` file (for angular) and for the other studio part I found that `global.scss` was not being updated anymore (the built file was under /themes/cstudioTheme`, so I updated the source and generated under `/styles`.

rart (approved):

PR #1861, opened 2020-07-17
[4198][next] Change the character limit for username when creating a user

jvega190:
https://github.com/craftercms/craftercms/issues/4198


rart (changes requested):
apply changes from 3.1.x

rart (changes requested):
- Remove common.css from entry.ftl
- Revert head.ftl & entry.ftl

rart (approved):

PR #1889, opened 2020-09-01
[4244] Allow using fetchRolesInSite without admin permissions

aachavarria:
https://github.com/craftercms/craftercms/issues/4244


rart (changes requested):
Please update the usage(s) of fetchRolesInSite to use the non admin where required (e.g. fetchInitialState)

rart (approved):

PR #1895, opened 2020-09-06
[3700][3.1.x] Display the name of the user whose password is being reset

aachavarria:
https://github.com/craftercms/craftercms/issues/3700

### DIFF
--- a/static-assets/ng-views/admin-users.html
+++ b/static-assets/ng-views/admin-users.html
@@ -304,14 +304,28 @@
 
 <script type="text/ng-template" id="resetPassword.html">
   <div class="modal-header">
-    <h3 class="modal-title">Reset Password</h3>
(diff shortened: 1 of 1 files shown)


static-assets/ng-views/admin-users.html
```diff
@@ -304,14 +304,28 @@ <h3 ng-if="!dialogMode" class="modal-title text-capitalize">{{user.name}}</h3>
 
 <script type="text/ng-template" id="resetPassword.html">
   <div class="modal-header">
-    <h3 class="modal-title">Reset Password</h3>
+    <h3 class="modal-title">{{ 'common.RESET_PASSWORD' | translate }}</h3>
```
rart, line 307: Does this translation already exist? Same for the others below..

rart: Verified.

rart (approved):

PR #1906, opened 2020-09-09
[4286][next] Bump node-sass from 4.13.0 to 4.13.1 in /ui/app

jvega190:
https://github.com/craftercms/craftercms/issues/4286

### DIFF
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -94,7 +94,7 @@
     "@babel/runtime": "^7.8.3",
     "babel-plugin-react-intl": "^5.1.16",
     "ncp": "^2.0.0",
-    "node-sass": "^4.12.0",
+    "node-sass": "4.13.1",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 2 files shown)


ui/app/package.json
```diff
@@ -94,7 +94,7 @@
     "@babel/runtime": "^7.8.3",
     "babel-plugin-react-intl": "^5.1.16",
     "ncp": "^2.0.0",
-    "node-sass": "^4.12.0",
+    "node-sass": "4.13.1",
```
rart: Let's use ^4.13.1 so we keep getting updates from dependabot. Not sure it would still send updates if an exact version is specified.

rart (changes requested):

ui/app/package.json
rart: Same change for all PRs to all branches

rart (approved):

PR #1909, opened 2020-09-10
[1632] In the Form Engine: Add a property sheet plugin type for repository path that allow author to browse a path 

aachavarria:
https://github.com/craftercms/craftercms/issues/1632


static-assets/components/cstudio-forms/data-sources/shared-content.js
```diff
@@ -367,8 +367,8 @@ YAHOO.extend(CStudioForms.Datasources.SharedContent, CStudioForms.CStudioFormDat
       {
         label: CMgs.format(langBundle, 'repositoryPath'),
         name: 'repoPath',
-        type: 'string',
-        defaultValue: '/site/',
+        type: 'repository',
```
rart: let's call this `content-path-input`

static-assets/themes/cstudioTheme/css/console.css
```diff
@@ -24,6 +24,25 @@ body, tbody {
   height: 100%;
 }
 
+.repository-selector {
```
rart: content-path-input

ui/app/src/components/Dialogs/PathBrowserDialog.tsx
```diff
@@ -0,0 +1,254 @@
...
+  ok: {
+    id: 'words.accept',
+    defaultMessage: 'Accept'
+  },
+  cancel: {
+    id: 'words.cancel',
+    defaultMessage: 'Cancel'
+  },
+  create: {
+    id: 'path.browser.create',
```
rart: pathBrowser.createFolderButtonLabel

ui/app/src/components/Dialogs/PathBrowserDialog.tsx
```diff
@@ -0,0 +1,254 @@
+/*
```
rart: Let's call this file/component PathSelectionDialog

ui/app/src/components/Dialogs/PathBrowserDialog.tsx
```diff
@@ -0,0 +1,254 @@
...
+  };
+
+  const onPathChanged = (path: string) => {
+    setCurrentPath(path);
+    setExpanded(getAllPaths(path, rootPath));
+  };
+
+  return (
+    <>
+      <DialogHeader title="Select Path" onDismiss={onClose} />
```
rart: Select Path should be i18n.
Add prop for `title` and in here, make "Select Path" the default message if the prop is not supplied.

ui/app/src/components/Navigation/FolderBrowserTreeView.tsx
```diff
@@ -0,0 +1,281 @@
...
+}
+
+function RenderTreeNode({ node }: RenderTreeNode) {
+  const classes = useStyles({});
+  return node.id === 'loading' ? (
+    <div className={classes.loading}>
+      <Typography>
+        <FormattedMessage id="folderBrowserTreeView.loading" defaultMessage={'Loading...'} />
+      </Typography>
+      <CircularProgress size={16} />
```
rart: Looks like this is still inverted. The spinner should be to the left of the message.

Use \<FormattedMessage id=words.loading/>&hellip;

ui/app/src/components/Dialogs/PathBrowserDialog.tsx
```diff
@@ -0,0 +1,254 @@
...
+  };
+
+  const onNodeSelected = (event: React.ChangeEvent<{}>, nodeId: string) => {
+    setCurrentPath(nodeId);
+  };
+
+  const onPathChanged = (path: string) => {
+    setCurrentPath(path);
+    setExpanded(getAllPaths(path, rootPath));
+  };
```
rart: While verifying a path, the accept button should be disabled.

When there's a manual path entry on the input, if the connection is slow enough, there's a moment when you could accept a path that does not exist.

ui/app/src/utils/path.ts
```diff
@@ -105,6 +105,14 @@ export function getParentsFromPath(path: string, rootPath: string): string[] {
   return [rootPath, ...splitPath.map((value, i) => `${rootPath}/${splitPath.slice(1, i + 1).join('/')}`).splice(2)];
 }
 
+export function getAllPaths(path: string, rootPath: string) {
```
rart: getIndividualPaths

static-assets/components/cstudio-forms/data-sources/shared-content.js
```diff
@@ -367,8 +367,8 @@ YAHOO.extend(CStudioForms.Datasources.SharedContent, CStudioForms.CStudioFormDat
...
+        defaultValue: '/site',
```
rart: make the default & root path `/site/` (with the slash at the end too)

rart (changes requested):

ui/legacy/src/components/cstudio-admin/mods/content-type-propsheet/repository.js
```diff
@@ -0,0 +1,111 @@
...
+          className="repository-selector--input"
+          placeholder='/site/...'
+          onBlur={validations.startsWith ? onBlur : null}
+          onChange={(e) => onChange(e.target.value)}
+        />
+        <i
+          className="repository-selector--icon fa fa-search"
+          aria-hidden="true"
+          onClick={openPathBrowser}
+        />
```
rart: Wrap with a button. Make sure you can tab between input and button in that order.

ui/app/src/components/Navigation/FolderBrowserTreeView.tsx
```diff
@@ -0,0 +1,281 @@
...
+  id: string;
+  name: string;
+  children?: TreeNode[];
+  fetched?: boolean;
+}
+
+const useStyles = makeStyles(() =>
+  createStyles({
+    wrapper: {
+      minHeight: '60vh',
```
rart: Move the min height to the dialog (body). And, let's use 40vh.

static-assets/components/cstudio-forms/data-sources/shared-content.js
rart: Have a look around for other inputs that would too benefit from using this selector. An example is the "Browse path" prop of data sources.

ui/app/src/components/Dialogs/PathSelectionDialog.tsx
```diff
@@ -0,0 +1,270 @@
...
+    setCurrentPath(path);
+    setExpanded(getIndividualPaths(path, rootPath));
+  };
+
+  return (
+    <>
+      <DialogHeader
+        title={title
+          ? title
+          : <FormattedMessage id="path.dialog.title" defaultMessage="Select Path" />}
```
rart: change message id to pathSelectionDialog.title

ui/app/src/utils/path.ts
```diff
@@ -105,6 +105,14 @@ export function getParentsFromPath(path: string, rootPath: string): string[] {
   return [rootPath, ...splitPath.map((value, i) => `${rootPath}/${splitPath.slice(1, i + 1).join('/')}`).splice(2)];
 }
 
+export function getIndividualPaths(path: string, rootPath: string) {
```
rart: why do you need the rootpath as a separate argument?

aachavarria: I was using two different rootpath. Now only one, we can remove it, if you want to

rart (approved):

PR #1924, opened 2020-09-17
[4302] Delete dialog submit fails with large submission comments

jvega190:
https://github.com/craftercms/craftercms/issues/4302

- Delete dialog submit fails with large submission comments
- Updated delete api to v2

### DIFF
--- a/ui/app/src/modules/Content/Delete/DeleteDialog.tsx
+++ b/ui/app/src/modules/Content/Delete/DeleteDialog.tsx
@@ -46,7 +46,7 @@ interface DeleteDialogContentUIProps {
   resource: Resource<DeleteDependencies>;
   items: SandboxItem[];
   submissionComment: string;
-  submissionCommentCount: any;
+  submissionCommentMaxLength: number;
   setSubmissionComment: Function;
   onSelectionChange?: Function;
 }
@@ -56,7 +56,7 @@ interface DeleteDialogUIProps {
   items: SandboxItem[];
   selectedItems: SandboxItem[];
   submissionComment: string;
-  submissionCommentCount: any;
+  submissionCommentMaxLength: number;
   setSubmissionComment: Function;
   apiState: any;
   handleSubmit: any;
@@ -121,6 +121,9 @@ const deleteDialogStyles = makeStyles((theme) =>
     depsContainer: {
       minHeight: '350px'
     },
+    countContainer: {
+      padding: '5px'
+    },
     submissionCommentCount: {
       fontSize: '14px',
       color: palette.gray.medium4
@@ -133,7 +136,7 @@ function DeleteDialogContentUI(props: DeleteDialogContentUIProps) {
     resource,
     items,
     submissionComment,
-    submissionCommentCount,
+    submissionCommentMaxLength,
     setSubmissionComment,
     onSelectionChange
   } = props;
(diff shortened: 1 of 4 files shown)


ui/app/src/modules/Content/Delete/DeleteDialog.tsx
```diff
@@ -273,14 +314,20 @@ function DeleteDialogWrapper(props: DeleteDialogProps) {
     );
   }, [selectedItems, setApiState, siteId]);
 
+  useEffect(() => {
+    setSubmissionCommentCount({
+      currentLength: submissionComment.length
```
rart: is there a reason to not use directly the comment to get it's length in place? that way we save some mem by dropping an additional state and effect.

ui/app/src/modules/Content/Delete/DeleteDialog.tsx
```diff
@@ -41,11 +40,13 @@ import CircularProgress from '@material-ui/core/CircularProgress';
 import Dialog from '@material-ui/core/Dialog';
 import palette from '../../../styles/palette';
 import Grid from '@material-ui/core/Grid';
+import { Typography } from '@material-ui/core';
 
 interface DeleteDialogContentUIProps {
   resource: Resource<DeleteDependencies>;
   items: SandboxItem[];
   submissionComment: string;
+  submissionCommentCount: any;
```
rart: type

ui/app/src/modules/Content/Delete/DeleteDialog.tsx
```diff
@@ -55,6 +56,7 @@ interface DeleteDialogUIProps {
   items: SandboxItem[];
   selectedItems: SandboxItem[];
   submissionComment: string;
+  submissionCommentCount: any;
```
rart: type

ui/app/src/modules/Content/Delete/DeleteDialog.tsx
```diff
@@ -156,8 +169,31 @@ function DeleteDialogContentUI(props: DeleteDialogContentUIProps) {
               InputProps={{
                 className: classes.textField
               }}
+              inputProps={{ maxLength: submissionCommentCount.maxLength }}
```
rart: submissionCommentCount is not a good name for the var that contains maxLength. also, consider just having a var for that specifically; what's the value of having it in an object with other stuff?

rart (changes requested):

ui/app/src/modules/Content/Delete/DeleteDialog.tsx
```diff
@@ -235,11 +277,11 @@ export default function DeleteDialog(props: DeleteDialogProps) {
 function DeleteDialogWrapper(props: DeleteDialogProps) {
   const { items, onClose, onDismiss, onSuccess } = props;
   const [submissionComment, setSubmissionComment] = useState('');
+  const submissionCommentMaxLength = 250;
```
rart: Let's move to GlobaState.env and supply to all dialogs with submission comments

rart (changes requested):

ui/app/src/modules/Content/Delete/DeleteDialog.tsx
jvega190: done, other dialogs will be done in here https://github.com/craftercms/craftercms/issues/4305

rart (approved):

PR #1938, opened 2020-10-01
[4324] Add spell check attribute to TinyMCE2 and 5

jvega190:
https://github.com/craftercms/craftercms/issues/4324


static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -560,9 +573,9 @@ CStudioAuthoring.Module.requireModule(
             valid_elements: '+*[*]',
             extended_valid_elements: '+*[*]',
             valid_children: '+*[*]',
-            valid_elements: '+*[*]',
```
jvega190, line 563: property was repeated

rart (approved):

PR #1939, opened 2020-10-01
[4324] Add spell check attribute to TinyMCE2 and 5

jvega190:
https://github.com/craftercms/craftercms/issues/4324


rart (approved):
@russdanner FYI this enables spellcheck **by default** and **by omission** — (i.e. from this update forward, spellcheck will be on/enabled for people who haven't specified a setting and the content type modeller will set the option to checked (on) by default too.

PR #1960, opened 2020-10-08
[studio-ui] Allow users without write permissions read-only access to item history

aachavarria:

### DIFF
--- a/ui/app/src/components/SystemStatus/GlobalDialogManager.tsx
+++ b/ui/app/src/components/SystemStatus/GlobalDialogManager.tsx
@@ -29,7 +29,6 @@ import { isPlainObject } from '../../utils/object';
 import ViewVersionDialog from '../../modules/Content/History/ViewVersionDialog';
 import CompareVersionsDialog from '../../modules/Content/History/CompareVersionsDialog';
 import RejectDialog from '../Dialogs/RejectDialog';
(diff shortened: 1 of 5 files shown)


ui/app/src/components/SystemStatus/GlobalDialogManager.tsx
```diff
@@ -211,6 +211,7 @@ function GlobalDialogManager() {
       <HistoryDialog
         open={state.history.open}
         versionsBranch={versionsBranch}
+        permissions={permissions}
```
rart: Should/could we send the permissions for the item in question instead of the whole thing?

ui/app/src/modules/Content/History/HistoryDialog.tsx
```diff
@@ -392,6 +405,7 @@ function HistoryDialog(props: HistoryDialogProps) {
         <SuspenseWithEmptyState resource={versionsResource}>
           <VersionList
             resource={versionsResource}
+            permissionsResource={permissionsResource}
```
rart: change resource prop to be `{ versions: Resource<...>; permissions: Resource<...>; } `

rart (changes requested):

rart (approved):

PR #1968, opened 2020-10-12
[4342][3.1.x] Add property to text input and text area form controls to escape script tags

jvega190:
https://github.com/craftercms/craftercms/issues/4342

### DIFF
--- a/static-assets/components/cstudio-forms/controls/input.js
+++ b/static-assets/components/cstudio-forms/controls/input.js
@@ -31,7 +31,7 @@ CStudioForms.Controls.Input =
     this.id = id;
     this.readonly = readonly;
     this.supportedPostFixes = ['_s', '_t'];
-    this.escapeScripts = false;
+    this.escapeContent = false;
 
     this.formatMessage = CrafterCMSNext.i18n.intl.formatMessage;
     this.controlsCommonMessages = CrafterCMSNext.i18n.messages.controlsCommonMessages;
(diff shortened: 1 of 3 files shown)


rart (changes requested):
More than escaping scripts, we’re escaping content so we should probably name the option and all around escapeContent.

static-assets/components/cstudio-forms/controls/input.js
```diff
@@ -169,7 +175,9 @@ YAHOO.extend(CStudioForms.Controls.Input, CStudioForms.CStudioFormField, {
     this.inputEl = inputEl;
     YAHOO.util.Dom.addClass(inputEl, 'datum');
     YAHOO.util.Dom.addClass(inputEl, 'cstudio-form-control-input');
-    inputEl.value = (this.value = '_not-set') ? config.defaultValue : this.value;
+
+    const valueToSet = this.escapeContent ? CStudioForms.Util.unEscapeXml(this.value) : this.value;
+    inputEl.value = (this.value = '_not-set') ? config.defaultValue : valueToSet;
```
rart: This is an odd statement. Looks like it will always be '_not-set' and will always use config.defaultValue.

`(this.value = '_not-set')` always sets the value and returns that string, that value is always truthy so condition is always true.

Perhaps this was meant to check equality of this.value instead of assign it? Please review.

static-assets/components/cstudio-forms/controls/textarea.js
```diff
@@ -132,7 +137,9 @@ YAHOO.extend(CStudioForms.Controls.Textarea, CStudioForms.CStudioFormField, {
     this.inputEl = inputEl;
     YAHOO.util.Dom.addClass(inputEl, 'datum');
     YAHOO.util.Dom.addClass(inputEl, 'cstudio-form-control-input');
-    inputEl.value = (this.value = '_not-set') ? config.defaultValue : this.value;
+
+    const valueToSet = this.escapeContent ? CStudioForms.Util.unEscapeXml(this.value) : this.value;
+    inputEl.value = (this.value = '_not-set') ? config.defaultValue : valueToSet;
```
rart: Same deal as above

rart (changes requested):

static-assets/components/cstudio-forms/controls/input.js
jvega190: yes, I checked the code and seems like the intention was to check equality instead of assigning value

rart (approved):

PR #1970, opened 2020-10-13
 [4342][next] Add property to text input and text area form controls to escape script tags 

jvega190:
https://github.com/craftercms/craftercms/issues/4342


rart (changes requested):
Please address 3.1.x review and carry updates here

rart (approved):

PR #1977, opened 2020-10-16
[4347] Add support in /api/2/audit for friendly site name.

jvega190:
https://github.com/craftercms/craftercms/issues/4347

### DIFF
--- a/static-assets/ng-views/audit.html
+++ b/static-assets/ng-views/audit.html
@@ -22,7 +22,7 @@
     <div class="audit-nav">
         <div class="nav-containers mb20">
             <label class="pr5">{{ 'admin.audit.SITE' | translate }}: </label>
-            <span class="pr20" ng-show="!audit.sites">{{ audit.site }}</span>
+            <span class="pr20" ng-show="!audit.sites">{{sites.byId[audit.site].name}}</span>
             <div class="btn-group" uib-dropdown is-open="sitesDropdown.isopen" style="margin-right: 5px;" ng-show="audit.sites">
                 <button type="button" class="btn btn-default dropdown-toggle text-capitalize" uib-dropdown-toggle title="{{audit.site || 'admin.audit.ALL_SITES' | translate}}" >
                     {{audit.siteLabel || audit.site || 'admin.audit.ALL_SITES' | translate }}  <span class="caret"></span>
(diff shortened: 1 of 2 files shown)


static-assets/scripts/admin.js
```diff
@@ -214,8 +214,8 @@
         });
       };
 
-      this.getSpecificAudit = function(auditId) {
-        return $http.get(audit(auditId));
+      this.getSpecificAudit = function(siteName) {
+        return $http.get(audit(siteName));
```
rart, line 218: The API changed to retrieve by name instead of ID?

static-assets/scripts/admin.js
```diff
@@ -423,6 +423,7 @@
       audit.allTimeZones = moment.tz.names();
       audit.sort = 'date';
       $scope.originValues = [$translate.instant('admin.audit.ALL_ORIGINS'), 'API', 'GIT'];
+      $scope.sites = CrafterCMSNext.system.store.getState().sites;
```
rart, line 426: Best if we prep for this being async.

rart, line 426: I think the API now gives you the site name too, so we don't need to keep a separate site lookup, no?
If we do, best if we prep for this being async.

static-assets/scripts/admin.js
rart, line 218: The API changed to retrieve by name instead of ID?

static-assets/ng-views/audit.html
```diff
@@ -219,7 +221,7 @@ <h1>{{ 'admin.audit.AUDIT' | translate }}</h1>
             <tr dir-paginate="log in audit.logs | filter:q | itemsPerPage: audit.logsPerPage"
                 current-page="audit.pagination.current" total-items="audit.totalLogs">
                 <td class="min-cel-width">{{log.operationTimestamp | amTimezone:audit.timeZone | amDateFormat:'MM-DD-YYYY hh:mm:ssa'}}</td>
-                <td>{{log.siteName}}</td>
+                <td>{{ sites.byId[log.siteName] ? sites.byId[log.siteName].name : log.siteName }}</td>
```
rart, line 224: I think the API now gives you the site name too, so we don't need to keep a separate site lookup, no?

rart (changes requested):

static-assets/scripts/admin.js
jvega190: yes it did change, it does retrieve by name

rart: @sumerjabri was it intended to change the API to retrieve by name instead of id?
If this is the case, this is outdated: https://app.swaggerhub.com/apis/craftercms/studio_3_2#/audit/getAuditLogById

sumerjabri: You can show by name, but API keys off of ID.

jvega190:
Closing due to changes made with outdated API

PR #1981, opened 2020-10-20
[studio-ui] Remove ace editor from our code, pull from npm and copy during build

rart:
This PR removes all of ace editor source from our repo. The idea is to pull at build time and copy on to the right place in studio's static assets where it's served from. Upgrading, now becomes a matter of changing the package.json version instead of overriding/merging the new version of the library on top of our copy of itself.

**Needs to be merged together with https://github.com/craftercms/studio/pull/2223**


ui/ace/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "main": "index.js",
+  "author": "Roy Art",
+  "license": "LGPL-3.0-only",
+  "private": true,
+  "dependencies": {
+    "ace-builds": "^1.4.12"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "build": "chmod +x build.sh; ./build.sh"
```
rart, line 13: @sumerjabri @jdrossl please review:

The [gradle build](https://github.com/craftercms/studio/blob/1433f32c5fc326c6e63bd7463a8f3dd6ec5aeb40/pom.xml#L655) will execute this task. Essentially, adds exec permissions to the build.sh (i.e. `chmod +x build.sh`) and then runs bash script (i.e. `./build.sh`), which simply copies a bunch of files to the target location (see next PR comment below).

Will this work in all relevant build execution contexts/scenarios?

Is there a better way to do this? It's important that the [npm install](https://github.com/craftercms/studio/blob/1433f32c5fc326c6e63bd7463a8f3dd6ec5aeb40/pom.xml#L645) task runs first, but once that completes, the bash needn't run via a yarn task if there's a better way.

I could do this via node apis instead of bash but they'd end up doing same thing.

ui/ace/build.sh
```diff
@@ -0,0 +1,19 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+rm -rf ../../static-assets/libs/ace
+cp -r node_modules/ace-builds/src-min-noconflict ../../static-assets/libs/ace
+cat src/ace-append.js >> ../../static-assets/libs/ace/ace.js
```
rart, line 19: @sumerjabri @jdrossl please review:

This is the bash script content that the [gradle build](https://github.com/craftercms/studio/blob/1433f32c5fc326c6e63bd7463a8f3dd6ec5aeb40/pom.xml#L655) executes.

ui/ace/package.json
jdrossl: If you only need to copy files I would suggest doing it in maven instead of a new script, take a look at https://github.com/craftercms/studio/blob/develop/pom.xml#L747

sumerjabri (approved):
We'll replace the bash copy with Maven in the next PR

ui/ace/package.json
rart: Besides the copying of files, the bash script also performs a file concatenation. See line 19 of the build.sh below.

sumerjabri: @jdrossl any ideas here or will we go with bash?

jdrossl: maven can only copy files, not append to them

PR #1984, opened 2020-10-21
[3959] Create folder in Scripts blocks the use of {} in the name which are used to indicate parametrized portions of the url

aachavarria:
https://github.com/craftercms/craftercms/issues/3959#event-3904554465


rart (changes requested):
Use encodeURIComponent for all instances of path encoding

rart (approved):

PR #1998, opened 2020-11-03
[studio-ui] Show only supported actions in the History dialog #4377

aachavarria:
https://github.com/craftercms/craftercms/issues/4377


rart (changes requested):
Missing handling of more/other types of assets, etc. 
I think that we only current support xml docs. Please test.

rart (approved):

PR #2009, opened 2020-11-05
[4387][3.1.x] Don't allow underscore (_) in site names

jvega190:
https://github.com/craftercms/craftercms/issues/4387

### DIFF
--- a/ui/app/src/components/BlueprintForm.tsx
+++ b/ui/app/src/components/BlueprintForm.tsx
@@ -120,8 +120,9 @@ function BlueprintForm(props: BlueprintFormProps) {
       setInputs({ [e.target.name]: e.target.checked, submitted: false });
     } else if (e.target.name === 'siteId') {
       const invalidSiteId = (e.target.value.startsWith('0') || e.target.value.startsWith('-') || e.target.value.startsWith('_'));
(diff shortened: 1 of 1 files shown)


ui/app/src/components/BlueprintForm.tsx
```diff
@@ -120,8 +120,9 @@ function BlueprintForm(props: BlueprintFormProps) {
       setInputs({ [e.target.name]: e.target.checked, submitted: false });
     } else if (e.target.name === 'siteId') {
       const invalidSiteId = (e.target.value.startsWith('0') || e.target.value.startsWith('-') || e.target.value.startsWith('_'));
+      const siteId = e.target.value.replace(/[^a-zA-Z0-9-_]/g, '').replace(/_/g, '-').toLowerCase();
```
rart, line 123: the first regex is stripping off the underscores, so the second won't do anything

rart (changes requested):

ui/app/src/components/BlueprintForm.tsx
jvega190: that regex starts with `^`, so it will actually strip everything that is not `a-zA-Z0-9-_`

rart (approved):

sumerjabri (approved):

PR #2010, opened 2020-11-05
[4192] Add friendly name to sites list 

jvega190:
https://github.com/craftercms/craftercms/issues/4192
Update from https://github.com/craftercms/craftercms/issues/4192#issuecomment-722440046. Request only to send sitename if changed by user, otherwise it's null on the request


rart:
Will be fixed by backend.

PR #2011, opened 2020-11-06
[4377] Show only supported actions in the History dialog

aachavarria:
https://github.com/craftercms/craftercms/issues/4377


rart (changes requested):
Conflicts

rart (approved):

PR #2024, opened 2020-11-13
[4401] Add router to Site Config and embedded mode

jvega190:
https://github.com/craftercms/craftercms/issues/4401


templates/web/site-config.ftl
```diff
@@ -58,7 +59,13 @@
 
 </head>
 <body class="yui-skin-cstudioTheme">
-<div id="admin-console" class="categories-panel-active"></div>
+  <#if RequestParameters.mode?? && RequestParameters.mode == "embedded">
+    <#assign mode = RequestParameters.mode />
+  <#else>
+    <#assign mode = ""/>
+  </#if>
```
rart: You may want to put this above and, when mode is embedded, put some classes on the body and/or the root elem to target the sidebar & toolbar hide styles.

static-assets/themes/cstudioTheme/css/console.css
```diff
@@ -116,6 +116,15 @@ input[type="checkbox"].content-type-selector--checkbox {
...
 }
 
+#admin-console.embedded #categories-panel{
+  display: none;
+}
+
+#admin-console.embedded #cstudio-admin-console-workarea {
+  width: 100%;
+  margin-left: 0;
+}
```
rart: Doesn't look like you're using these yet.

templates/web/site-config.ftl
jvega190: @rart right now the class if being set to the `admin-console` root element, and styles applied to that class

templates/web/site-config.ftl
```diff
@@ -58,7 +59,13 @@
...
+
+  <div id="admin-console" class="categories-panel-active ${mode}"></div>
```
jvega190: @rart here ^

static-assets/themes/cstudioTheme/css/console.css
jvega190: The class is being set to the `#admin-console` element

templates/web/site-config.ftl
jvega190: I moved the class up to the body (I think it's clearer)

templates/web/site-config.ftl
```diff
@@ -57,8 +58,14 @@
...
 </head>
-<body class="yui-skin-cstudioTheme">
-<div id="admin-console" class="categories-panel-active"></div>
+<#if RequestParameters.mode?? && RequestParameters.mode == "embedded">
+  <#assign mode = RequestParameters.mode />
+<#else>
+  <#assign mode = ""/>
+</#if>
+
+<body class="yui-skin-cstudioTheme ${mode}">
```
rart: To make this safer, create a freemarker variable called "embedded" which you set to true if the query arg mode == embedded. Then, you set the embedded class name manually. This way there's no possibility of HTML injection.

templates/web/site-config.ftl
```diff
@@ -42,6 +42,7 @@
 
   <script src="/studio/static-assets/libs/momentjs/moment.min.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
   <script src="/studio/static-assets/libs/momentjs/moment-timezone-with-data-2012-2022.min.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
+  <script src="/studio/static-assets/libs/routie/routie.min.js?version=${UIBuildId!.now?string('Mddyyyy')}"></script>
```
rart: add routie version to its file name and remove the build num

ui/app/src/utils/i18n-legacy.ts
```diff
@@ -405,6 +405,14 @@ const contentTypesMessages = defineMessages({
...
     id: 'siteConfig.controlNotAvailable',
     defaultMessage: 'Control not available'
+  },
+  loadModuleError: {
+    id: 'siteConfig.loadModuleError',
+    defaultMessage: 'Unable to load "{tool}". Check Site tools configuration.'
+  },
+  toolNotFound: {
+    id: 'siteConfig.toolNotFound',
+    defaultMessage: 'Tool not found.'
```
rart: Include the tool id that wasn't found

static-assets/themes/cstudioTheme/css/console.css
```diff
@@ -116,6 +116,15 @@ input[type="checkbox"].content-type-selector--checkbox {
...
 
+.embedded #admin-console #categories-panel{
+  display: none;
+}
+
+.embedded #admin-console #cstudio-admin-console-workarea {
+  width: 100%;
+  margin-left: 0;
+}
+
```
rart, line 145: There's a big margin at the top when the toolbar is removed. I don't see any styles addressing that; are you addressing somewhere else?

static-assets/components/cstudio-common/common-api.js
```diff
@@ -6579,7 +6580,7 @@ var nodeOpen = false,
       /**
        * dynamically add a javascript file
        */
-      addJavascript: function(script, cache, callback) {
```
rart, line 6582: The change in the API is worrisome. Did you check the **entire** codebase for usages of addJavascript not using the original argument configuration? Did you searched for the "addJavascript" string alone or with its whole path?

Are cache and callback arts not used inside the function?

static-assets/components/cstudio-admin/base.js
```diff
@@ -21,14 +21,20 @@
 
   CStudioAdminConsole = {
     toolContainerEls: [],
+    isEmbedded: false,
```
rart: You could directly check here if the embedded class is present, right? No need to delay.

static-assets/components/cstudio-admin/base.js
```diff
@@ -58,6 +64,38 @@
...
     },
 
+    initRouter(tools) {
+      const _self = this;
+      const toolsNames = tools.map((tool) => tool.name);
+
+      routie('tool/:name?', function(name) {
+        if (toolsNames.includes(name)) {
+          _self.currentRoute = name;
+          if (_self.toolsModules[name]) {
```
rart, line 72: provided you already checked the name is correct above (line 72's if), is this if necessary?

static-assets/components/cstudio-admin/base.js
```diff
@@ -58,6 +64,38 @@
...
+          elem.className = 'work-area-error';
+          $('#cstudio-admin-console-workarea').html(elem);
+          CrafterCMSNext.render(elem, 'ErrorState', {
+            graphicUrl: '/studio/static-assets/images/warning_state.svg',
+            classes: {
+              root: 'craftercms-error-state'
+            },
+            error: {
+              code: '',
+              message: formatMessage(contentTypesMessages.toolNotFound)
```
rart: Include tool id in error message

static-assets/components/cstudio-admin/base.js
```diff
@@ -95,11 +143,24 @@
...
               toolContainerEl: toolContainerEl
             };
-
+            const tool = config.tools.tool[j];
             CStudioAuthoring.Module.requireModule(
-              'cstudio-console-tools-' + config.tools.tool[j].name,
-              '/static-assets/components/cstudio-admin/mods/' + config.tools.tool[j].name + '.js',
-              { config: config.tools.tool[j] },
+              'cstudio-console-tools-' + tool.name,
+              '/static-assets/components/cstudio-admin/mods/' + tool.name + '.js',
```
rart: Use modern syntax for string interpolation

rart (changes requested):

static-assets/components/cstudio-admin/base.js
jvega190: @rart yes it is necessary, the toolsNames list has the names (ids) of the tools that are to be loaded, the toolsModules are the ones already loaded

static-assets/components/cstudio-common/common-api.js
jvega190: I searched for the string `addJavascript` (alone) all over the code, and I found no usages with the optional arguments, also, in the code neither cache nor callback are being used

rart (approved):

PR #2033, opened 2020-11-19
[4415] Components Panel DnD for creating new component fails when content type has textarea or boolean fields

jvega190:
https://github.com/craftercms/craftercms/issues/4415

### DIFF
--- a/ui/app/src/services/contentTypes.ts
+++ b/ui/app/src/services/contentTypes.ts
@@ -26,7 +26,7 @@ import {
   LegacyFormDefinitionSection
 } from '../models/ContentType';
 import { LookupTable } from '../models/LookupTable';
-import { camelize, isBlank } from '../utils/string';
+import { camelize, capitalize, isBlank } from '../utils/string';
 import { forkJoin, Observable, of, zip } from 'rxjs';
 import { errorSelectorApi1, get } from '../utils/ajax';
 import { catchError, map, pluck, switchMap } from 'rxjs/operators';
@@ -40,7 +40,19 @@ const typeMap = {
   'image-picker': 'image'
 };
 
-const systemValidationsNames = ['itemManager', 'minSize', 'maxSize', 'maxlength', 'readonly', 'width', 'height'];
+const systemValidationsNames = [
+  'itemManager',
+  'minSize',
+  'maxSize',
+  'maxlength',
+  'readonly',
+  'width',
+  'height',
+  'minWidth',
+  'minHeight',
+  'maxWidth',
+  'maxHeight'
+];
 
 const systemValidationsKeysMap = {
   minSize: 'minCount',
@@ -50,7 +62,11 @@ const systemValidationsKeysMap = {
   tags: 'allowedContentTypeTags',
   readonly: 'readOnly',
   width: 'width',
-  height: 'height'
+  height: 'height',
+  minWidth: 'minWidth',
+  minHeight: 'minHeight',
+  maxWidth: 'maxWidth',
+  maxHeight: 'maxHeight'
 };
 
 function bestGuessParse(value: any) {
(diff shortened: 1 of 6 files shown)


ui/guest/src/utils/contentType.ts
```diff
@@ -75,9 +76,20 @@ export function getDefaultValue(field: ContentTypeField): string {
         return 'https://via.placeholder.com/150';
```
rart: Try using image dimensions from content type definition to set the right pic size

rart (commented):
Is this finished?

ui/guest/src/utils/contentType.ts
```diff
@@ -71,13 +72,30 @@ export function getDefaultValue(field: ContentTypeField): string {
...
-        // TODO use validation width, height
-        return 'https://via.placeholder.com/150';
+        // since it's a string, empty string needs to be validated
+        const width = JSON.parse(field.validations.width?.value).min
+          ? JSON.parse(field.validations.width?.value).min
+          : 150;
+        const height = JSON.parse(field.validations.height?.value).min
+          ? JSON.parse(field.validations.height?.value).min
+          : width;
+        return `https://via.placeholder.com/${width}x${height}`;
```
rart: What's a string?
By this time should be whatever it must be. Parsing should go to the service.

ui/app/src/services/contentTypes.ts
```diff
@@ -40,15 +40,17 @@ const typeMap = {
...
 const systemValidationsKeysMap = {
   minSize: 'minCount',
   maxSize: 'maxCount',
   maxlength: 'maxLength',
   contentTypes: 'allowedContentTypes',
   tags: 'allowedContentTypeTags',
-  readonly: 'readOnly'
+  readonly: 'readOnly',
+  width: 'width',
+  height: 'height'
```
rart: Do we currently have min/max width/height? Do we have both exact width/height & max/min or just one?

rart (changes requested):

ui/guest/src/utils/contentType.ts
jvega190: Done parsing in the service. Also, new fields for minWidth, maxWidth, minHeight and maxHeight

ui/app/src/services/contentTypes.ts
jvega190: Added new fields for minWidth, maxWidth, minHeight and maxHeight

ui/app/src/services/contentTypes.ts
```diff
@@ -71,7 +89,29 @@ function getFieldValidations(
...
+            name: prop.name,
+            type: prop.type,
+            value: parsedValidation.min
+          };
+          table[`max${capitalize(prop.name)}`] = {
+            name: prop.name,
+            type: prop.type,
+            value: parsedValidation.max
+          };
+        }
```
jvega190, line 111: @rart just for you to know, values are set unparsed in here, but later in code values are parsed before returning the validation fields.

rart: - Where about?
- Why wait/distribute?

ui/guest/src/classes/ICERegistry.ts
```diff
@@ -59,7 +59,13 @@ const validationChecks: { [key in ValidationKeys]: Function } = {
...
   maxLength() {},
   readOnly() {},
-  required() {}
+  required() {},
+  width() {},
+  height() {},
+  minWidth() {},
+  minHeight() {},
+  maxWidth() {},
+  maxHeight() {}
```
rart, line 68: What were the purpose of these? How come we have so many dummy functions? @aachavarria

aachavarria: That functions are used when the `drop_zone_enter` and `drop_zone_leave` events are triggered.
Right now we are only checking validation for "minCount" and "maxCount".
There are a pending ticket to add more validations: https://github.com/craftercms/craftercms/issues/4052

rart (approved):

PR #2034, opened 2020-11-20
Add router to Site Config and embedded mode - enhancement

jvega190:
https://github.com/craftercms/craftercms/issues/4401
Enhance no tool selected view (don't show while loading tool)

### DIFF
--- a/static-assets/themes/cstudioTheme/css/console.css
+++ b/static-assets/themes/cstudioTheme/css/console.css
@@ -76,6 +76,10 @@ input[type="checkbox"].content-type-selector--checkbox {
   flex-direction: column;
 }
 
+.embedded .work-area-empty, .embedded .work-area-error {
+  height: 100vh;
+}
+
 .work-area-empty img, .work-area-error img {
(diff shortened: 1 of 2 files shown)


static-assets/themes/cstudioTheme/css/console.css
```diff
@@ -69,9 +69,8 @@ input[type="checkbox"].content-type-selector--checkbox {
 }
 
 .work-area-empty, .work-area-error {
-  height: 100vh;
+  height: calc(100vh - 55px);
   display: flex;
-  margin-top: 50px;
```
rart, line 74: How does this work in the case of nothing selected with the top bar present?

rart (changes requested):

static-assets/themes/cstudioTheme/css/console.css
jvega190: It works good, actually in embedded mode it doesn't take the full height, will add the styles for embedded

rart (approved):

PR #2042, opened 2020-11-26
[4422] Persist content clipboard across refreshes

aachavarria:
https://github.com/craftercms/craftercms/issues/4422


ui/app/src/services/content.ts
```diff
@@ -942,40 +943,20 @@ export function getChildrenByPath(
...
-  return copy(site, path).pipe(
-    switchMap(() => paste(site, getParentPath(parentPath))),
-    // Duplicate only does shallow copy, so return the copied path
-    pluck('0')
-  );
+export function duplicate(siteId: string, path: string): Observable<string> {
+  return postJSON('/studio/api/2/content/duplicate', {
+    siteId,
+    path
+  }).pipe(pluck('response', 'item'));
```
rart: Return the whole response object so we don’t have to change the API if later on there’s more to return

rart (changes requested):

rart (approved):

rart:
Will be merged on https://github.com/craftercms/studio-ui/pull/2041

PR #2044, opened 2020-11-26
[4422] Persist content clipboard across refreshes

aachavarria:
https://github.com/craftercms/craftercms/issues/4422


ui/app/src/components/Navigation/PathNavigator/PathNavigator.tsx
```diff
@@ -216,16 +216,24 @@ export default function PathNavigator(props: WidgetProps) {
...
+                })
+              );
+            }
+          } else {
+            // payload.clipboard.type === 'CUT
+            const parentPath = getParentPath(payload.target);
+            if (parentPath === withoutIndex(state.currentPath)) {
+              dispatch(pathNavigatorRefresh({ id }));
+            }
           }
```
rart, line 237: What happens if you try to paste-from-cut an item on a child? We may need a validation there since I don't think that's allowed.

ui/app/src/components/Navigation/PathNavigator/PathNavigator.tsx
```diff
@@ -216,16 +216,24 @@ export default function PathNavigator(props: WidgetProps) {
         }
         case folderCreated.type:
         case itemsPasted.type: {
```
rart, line 218: I'm thinking we should always send/refresh the path to which you're pasting on. If you're there refresh, if you're not there, get the navigator to display that path on which you pasted the new item (same for cut or copy).

ui/app/src/models/Item.ts
```diff
@@ -99,7 +99,7 @@ export interface LegacyItem {
   [prop: string]: any;
 }
 
-export interface CopyItem {
-  uri?: string;
-  children?: CopyItem[];
+export interface PasteItem {
+  path?: string;
```
rart: 🤦🏻‍♂️

ui/app/src/state/epics/content.ts
```diff
@@ -161,34 +154,16 @@ const content = [
...
       ofType(pasteItem.type),
       withLatestFrom(state$),
       switchMap(([{ payload }, state]) => {
-        return paste(state.sites.active, payload.path).pipe(
-          map((resultingPaths) => {
+        return paste(state.sites.active, payload.path, state.content.clipboard).pipe(
+          map(({ items }) => {
             return batchActions([
-              emitSystemEvent(itemsPasted({ target: payload.path, resultingPaths })),
+              emitSystemEvent(itemsPasted({ target: payload.path, clipboard: state.content.clipboard })),
```
rart: Isn't the state.clipboard empty by this point?

ui/app/src/utils/path.ts
```diff
@@ -131,6 +131,31 @@ export function getIndividualPaths(path: string, rootPath?: string): string[] {
   }
 }
 
+export function getPasteItemFromPath(sourcePath: string, paths: string[]): PasteItem {
+  sourcePath = withoutIndex(sourcePath);
```
rart: Should this be done to all paths?

ui/app/src/state/epics/content.ts
aachavarria: no because we are calling unSetClipBoard after this, so in this point the state still contains the clipboard

ui/app/src/utils/path.ts
aachavarria: already fixed

static-assets/themes/cstudioTheme/yui/assets/skin.css
```diff
@@ -5644,7 +5644,8 @@ h2.yui-editor-skipheader, h3.yui-editor-skipheader {
   border-bottom-color: #243356;
 }
 
-.ygtvitem {
+.ygtv-highlight0:hover {
+  background-color: #e2e2e2;
```
rart: Use rgba(...). Try to mimic material. (We're going to require some theming work legacy side to support dark mode.)

ui/app/src/components/SystemStatus/GlobalDialogManager.tsx
```diff
@@ -399,6 +400,18 @@ function GlobalDialogManager() {
         onOk={createCallback(state.pathSelection.onOk, dispatch)}
       />
       {/* endregion */}
+
+      {/* region Item Menu */}
+      {state.itemMenu.open && (
```
rart: This will disrupt animation. Why the if?

ui/scss/src/temp.scss
```diff
@@ -416,7 +416,21 @@ div#acn-active-content {
...
+      padding: 3px 10px;
+    }
+
+    &:hover .item-menu {
+      display: block;
+      position: absolute;
+      top: 0;
+      right: 0;
+      cursor: pointer;
+    }
```
rart: Are these styles for the ellipsis? Perhaps better to name as such, then. E.g. item-menu-trigger

rart (changes requested):

rart (approved):

PR #2045, opened 2020-12-01
[4421] Create LegacyDashboardFrame component

jvega190:
Free empty space when top bar is invisible
Fix edit doesn't work
Update clear selected to immediately clear selection without too many actions

https://github.com/craftercms/craftercms/issues/4421

### DIFF
--- a/static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
+++ b/static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
@@ -1692,15 +1692,27 @@ WcmDashboardWidgetCommon.refreshDashboard = function(inst) {
   );
 };
 
+const dashboardsIds = ['MyRecentActivity', 'recentlyMadeLive', 'approvedScheduledItems', 'GoLiveQueue'];
+
 /**
  * refresh all dashboards
  */
 WcmDashboardWidgetCommon.refreshAllDashboards = function() {
   if (typeof WcmDashboardWidgetCommon != 'undefined') {
-    WcmDashboardWidgetCommon.refreshDashboard('MyRecentActivity');
-    WcmDashboardWidgetCommon.refreshDashboard('recentlyMadeLive');
-    WcmDashboardWidgetCommon.refreshDashboard('approvedScheduledItems');
-    WcmDashboardWidgetCommon.refreshDashboard('GoLiveQueue');
+    dashboardsIds.forEach((dashboardId) => {
+      WcmDashboardWidgetCommon.refreshDashboard(dashboardId);
+    });
     CStudioAuthoring.SelectedContent.clear();
+  }
+};
+
+WcmDashboardWidgetCommon.clearSelections = function() {
+  if (typeof WcmDashboardWidgetCommon != 'undefined') {
+    dashboardsIds.forEach((dashboardId) => {
+      $(`#${dashboardId}-table`)
+        .find('input[type="checkbox"]')
+        .prop('checked', false);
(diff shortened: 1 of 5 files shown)


static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
```diff
@@ -1692,15 +1692,27 @@ WcmDashboardWidgetCommon.refreshDashboard = function(inst) {
...
   }
 };
+
+WcmDashboardWidgetCommon.clearSelections = function() {
+  if (typeof WcmDashboardWidgetCommon != 'undefined') {
+    dashboardsIds.forEach((dashboardId) => {
+      $(`#${dashboardId}-table`)
+        .find('input[type="checkbox"]')
+        .prop('checked', false);
+      CStudioAuthoring.SelectedContent.clear();
```
rart: This line needn't be inside the loop, does it?

static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
```diff
@@ -1692,15 +1692,27 @@ WcmDashboardWidgetCommon.refreshDashboard = function(inst) {
...
     CStudioAuthoring.SelectedContent.clear();
   }
 };
+
+WcmDashboardWidgetCommon.clearSelections = function() {
+  if (typeof WcmDashboardWidgetCommon != 'undefined') {
+    dashboardsIds.forEach((dashboardId) => {
+      $(`#${dashboardId}-table`)
+        .find('input[type="checkbox"]')
+        .prop('checked', false);
```
rart, line 1714: Test if this is enough. Does it continue to work correctly if you check the unchecked again, new ones, etc

ui/scss/src/_studio-view.scss
```diff
@@ -206,6 +206,7 @@
   }
 
   .site-dashboard {
```
rart, line 209: Above on line 157 (`#activeContentActions`), I see the selected content thingie is positioned with visibility hidden. This would overlay and block pointer events from what's below. Check that that isn't an issue now that the top margin is removed.

rart (changes requested):

static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
jvega190: not needed, moving out from loop

static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
```diff
@@ -1692,15 +1692,27 @@ WcmDashboardWidgetCommon.refreshDashboard = function(inst) {
   );
 };
 
+const dashboardsIds = ['MyRecentActivity', 'recentlyMadeLive', 'approvedScheduledItems', 'GoLiveQueue'];
```
rart, line 1695: Don't these come from some XML file? We should use what comes from config somehow.

rart (changes requested):

static-assets/components/cstudio-dashboard-widgets/lib/wcm-dashboardwidget-common.js
jvega190: @rart no, right now the dashboard js files are being imported in the site-dashboard.ftl file, not being retrieved from a config.

rart (approved):

PR #2057, opened 2020-12-09
Page Builder Issues: Fixing Operations using Nested fieldID

aachavarria:


ui/guest/src/classes/ContentController.ts
```diff
@@ -241,12 +241,11 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
   // Insert in desired position
   result.splice(index, 0, item);
 
+  setCollection(model, fieldId, typeof index === 'string' ? removeLastPiece(index) : index, result);
```
rart: - Index may not have `.`, even when a string. `removeLastPiece` will return an empty string in that scenario.
- It's important we don't mutate existing objects. The setCollection chain looks suspicious, like it would mutate the containing collections of the target one. We need to spread/create-new objects for each of the pieces of the nested prop/collection.
- Perhaps the method(s) rather belong to utils/object

ui/guest/src/utils/array.ts
```diff
@@ -36,3 +36,10 @@ export function forEach<T = unknown, R = undefined>(
   }
   return emptyReturnValue;
 }
+
+export function mergeArraysAlternatively(a: string[], b: string[]): string[] {
```
rart, line 46: Add docs to function

ui/guest/src/utils/ice.ts
```diff
@@ -81,3 +81,13 @@ export function getCollection(model: ContentInstance, fieldId: string, index: st
...
 }
+
+export function setCollection(model: ContentInstance, fieldId: string, index: number | string, collection: string[]) {
+  if (fieldId.includes('.')) {
+    const concatFieldId = mergeArraysAlternatively(fieldId.split('.'), index.toString().split('.')).join('.');
+    setProperty(model, concatFieldId, collection);
+  } else {
+    Model.value(model, fieldId, collection);
+  }
+  return model;
```
rart: This will mutate nested property object references.

rart (changes requested):

ui/guest/src/classes/ContentController.ts
```diff
@@ -234,19 +234,22 @@ export function updateField(modelId: string, fieldId: string, index: string | nu
 
 export function insertItem(modelId: string, fieldId: string, index: number | string, item: ContentInstance): void {
   const models = getCachedModels();
-  const model = { ...models[modelId] };
-  const collection = Model.value(model, fieldId);
+  const collection = Model.value(models[modelId], fieldId);
```
rart, line 237: What happens if this is a nested collection? There's no use of index so how would it get to the right index.

ui/guest/src/store/reducers/root.ts
```diff
@@ -611,12 +612,12 @@ const drop_zone_enter: GuestReducer = (state, action) => {
 // region drop_zone_leave
 // TODO: Not pure
 const drop_zone_leave: GuestReducer = (state, action) => {
-  const { iceId } = action.payload;
+  const { physicalRecordId } = action.payload;
```
rart: Please update the uses of "physical" to "element"

ui/guest/src/classes/ContentController.ts
aachavarria: This function is not used at all, I think it is just a placeholder, also it is outdated

rart (approved):

PR #2073, opened 2020-12-18
Revert Content Issues;

aachavarria:
Fixing issues: Reverting content does not refresh preview;

### DIFF
--- a/ui/app/src/state/epics/itemVersions.ts
+++ b/ui/app/src/state/epics/itemVersions.ts
@@ -39,6 +39,9 @@ import { historyDialogClosed } from '../actions/dialogs';
 import { getHistory as getConfigurationHistory } from '../../services/configuration';
 import { reloadDetailedItem } from '../actions/content';
 import { showRevertItemSuccessNotification } from '../actions/system';
+import { batchActions } from '../actions/misc';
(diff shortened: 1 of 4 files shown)


ui/app/src/state/epics/itemVersions.ts
```diff
@@ -85,14 +88,22 @@ export default [
...
-      switchMap(({ payload }) => [
-        fetchItemVersions(),
-        showRevertItemSuccessNotification(),
-        reloadDetailedItem({ path: payload.path })
-      ])
+      withLatestFrom(state$),
+      switchMap(([{ payload }, state]) => {
+        if (payload.path === state.preview.guest.path) {
+          getHostToGuestBus().next({ type: RELOAD_REQUEST });
+        }
```
rart, line 98: This logic applies to next too? Isn't this all for legacy? In which case, the host to guest bus doesn't do anything...

aachavarria: This is for legacy but this is needed, now legacy LegacyConcierge is updating the preview.guest, also it is subscribe to RELOAD_REQUEST to call "CStudioAuthoring.Operations.refreshPreview()"

rart (approved):

PR #2074, opened 2020-12-19
[studio-ui] Legacy basic theme support #4433 - Add dark mode theming to legacy preview tools

jvega190:
https://github.com/craftercms/craftercms/issues/4433


jvega190:
@rart left as draft to clarify a couple of things first

rart (approved):

PR #2077, opened 2020-12-21
[3084] add drag and drop image insertion support to TinyMCE5

jvega190:
Add paste plugin to tinymce to enable drag and drop 
https://github.com/craftercms/craftercms/issues/3084


static-assets/components/cstudio-forms/controls/rte-tinymce5.js
```diff
@@ -284,7 +284,7 @@ CStudioAuthoring.Module.requireModule(
           templates = rteConfig.templates && rteConfig.templates.template ? rteConfig.templates.template : null;
 
           // https://www.tiny.cloud/docs/plugins/
-          pluginList = rteConfig.plugins;
+          pluginList = rteConfig.plugins + ' paste';
```
rart, line 289: What if it's already there?
Also, add comment stating why we are hard coding paste plugin.

rart (changes requested):

static-assets/components/cstudio-forms/controls/rte-tinymce5.js
jvega190: It doesn't affect if plugin also comes from configuration file.
Comment added

rart (approved):

PR #2058, opened 2020-12-09
[4383] Make active filters prominent in search results page

jvega190:
https://github.com/craftercms/craftercms/issues/4383

PR #2079, opened 2020-12-22
[4383] Make active filters prominent in search results page

rart:
Reopened from https://github.com/craftercms/studio-ui/pull/2058 to fix odd behaviours on the original PR.

craftercms/craftercms#4383


rart:
@jvega190 There are issues still.

- [x] Odd whitespace at the bottom of default mode results area.
Screen Shot 2020-12-22 at 9 26 02 AM

- [x] Embedded mode taking up all the space. Doesn't look like a modal.
image

- [x] Filters should close by clicking away from them when the drawer is not permanent.

- [x] The search bar overlaps with the view switching button when making the window quite small. (Removing the `width: 100%` from the search bar seems to fix.)
image

jvega190:
I can't check the previous checkboxes so I will do it in here
- [x] Odd whitespace at the bottom of default mode results area.
- [x] Embedded mode taking up all the space. Doesn't look like a modal. - Updated as in latest updates in figma
- [x] Filters should close by clicking away from them when the drawer is not permanent.
- [x] The search bar overlaps with the view switching button when making the window quite small. (Removing the width: 100% from the search bar seems to fix.)

jvega190:
@rart this is ready, since you opened the PR I can't move to 'ready to review'

ui/app/src/components/Controls/SearchBar.tsx
```diff
@@ -29,7 +29,6 @@ const useStyles = makeStyles((theme: Theme) =>
     search: {
       position: 'relative',
       background: (props: any) => props.background ?? theme.palette.background.default,
-      width: '100%',
```
rart, line 32: This component is used in many different places. Did you test ALL of its usages to check that it didn't break elsewhere?

rart (commented):
@jvega190

ui/app/src/components/Controls/SearchBar.tsx
jvega190: All good, tested in all used places

rart:
https://github.com/craftercms/studio-ui/pull/2079

PR #2098, opened 2021-01-08
[4459] Reject dialog doesn't load items.

jvega190:
Fix for error showing on console - an id needs to be decoded before retrieving an item from the resultMap array

https://github.com/craftercms/craftercms/issues/4459

### DIFF
--- a/static-assets/components/cstudio-dashboard-widgets/go-live-queue.js
+++ b/static-assets/components/cstudio-dashboard-widgets/go-live-queue.js
@@ -323,7 +323,7 @@ CStudioAuthoringWidgets.GoLiveQueueDashboard = function (widgetId, pageId) {
    */
   this.handleTopDownPathDependenciesItemsClick = function (event, matchedEl) {
     var isChecked = matchedEl.checked;
-    var selectedElementURI = matchedEl.id;
+    var selectedElementURI = decodeURIComponent(matchedEl.id);
     var item = CStudioAuthoringWidgets.GoLiveQueueDashboard.resultMap[selectedElementURI];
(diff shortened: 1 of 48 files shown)


static-assets/components/cstudio-dashboard-widgets/go-live-queue.js
```diff
@@ -323,7 +323,7 @@ CStudioAuthoringWidgets.GoLiveQueueDashboard = function (widgetId, pageId) {
    */
   this.handleTopDownPathDependenciesItemsClick = function (event, matchedEl) {
     var isChecked = matchedEl.checked;
-    var selectedElementURI = matchedEl.id;
+    var selectedElementURI = decodeURIComponent(matchedEl.id);
```
rart: This fixes the console error but reject dialog still doesn't render the selected items on the dialog.

rart (changes requested):

static-assets/components/cstudio-dialogs/reject.js
```diff
@@ -24,6 +24,10 @@ CStudioAuthoring.Dialogs.DialogReject = function () {
   this.moduleName = 'reject';
   this.reasonHash = [];
   this.uncheckedItemsArrayNew = [];
+
+  const i18n = CrafterCMSNext.i18n;
```
rart: this isn’t used any more is it?

rart (approved):

static-assets/components/cstudio-dialogs/reject.js
jvega190: is isn't, removed

PR #2101, opened 2021-01-12
[3.0.x] issues/ee-214

jvega190:


jvega190:
@rart pending some details with upload that I need to discuss with you

rart (approved):
FYI the actual code changes performed for the fix can be observed free of code format differences here: https://github.com/jvega190/studio-ui/pull/1/files?diff=unified&w=1

PR #2104, opened 2021-01-12
[3959] Create folder in Scripts blocks the use of {} in the name which are used to indicate parametrized portions of the url

aachavarria:
https://github.com/craftercms/craftercms/issues/3959

Fixing UI Issues;
Pending work on BE because "/studio/api/1/services/api/1/content/get-items-tree.json" is not returning children

### DIFF
--- a/ui/app/src/components/Dialogs/CreateFileDialog.tsx
+++ b/ui/app/src/components/Dialogs/CreateFileDialog.tsx
@@ -98,7 +98,7 @@ function CreateFileUI(props: CreateFileUIProps) {
     setState({ inProgress: true, submitted: true });
 
     if (name) {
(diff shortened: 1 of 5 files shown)


ui/app/src/components/Dialogs/CreateFileDialog.tsx
```diff
@@ -97,7 +98,7 @@ function CreateFileUI(props: CreateFileUIProps) {
     setState({ inProgress: true, submitted: true });
 
     if (name) {
-      const fileName = type === 'controller' ? `${name}.groovy` : `${name}.ftl`;
+      const fileName = type === 'controller' ? `${encodeURI(name)}.groovy` : `${encodeURI(name)}.ftl`;
```
rart: use `encodeURIComponent`. Change in create folder if wrong there too.

rart (changes requested):

rart (approved):

PR #2053, opened 2020-12-09
[4430] Implement JWT

rart:
https://github.com/craftercms/craftercms/issues/4430

ToDo:
- [x] How does JWT change "Reset password" processes
- [x] How does JWT change "Recover password" processes
- [x] Assess/adjust AuthMonitor (re-login dialog)

PR #2105, opened 2021-01-12
[4453] Migrate local storage to persisted user props

aachavarria:
https://github.com/craftercms/craftercms/issues/4453


rart:
Will be merged via JWT 

https://github.com/rart/studio-ui/pull/3
https://github.com/craftercms/studio-ui/pull/2053

PR #2106, opened 2021-01-12
[4447] Design & implement the Access Token Management UI

aachavarria:
https://github.com/craftercms/craftercms/issues/4447


rart:
Will be merged via JWT 

https://github.com/rart/studio-ui/pull/4
https://github.com/craftercms/studio-ui/pull/2053

PR #2153, opened 2021-02-08
[4498] Add bulk getSandboxItems API calls

aachavarria:
https://github.com/craftercms/craftercms/issues/4498

### DIFF
--- a/ui/app/src/services/content.ts
+++ b/ui/app/src/services/content.ts
@@ -46,6 +46,7 @@ import { getPasteItemFromPath } from '../utils/path';
 import { StandardAction } from '../models/StandardAction';
 import { GetChildrenResponse } from '../models/GetChildrenResponse';
 import { GetItemWithChildrenResponse } from '../models/GetItemWithChildrenResponse';
+import { FetchItemsByPathOptions } from '../models/FetchItemsByPath';
 
 export function getComponentInstanceHTML(path: string): Observable<string> {
   return getText(`/crafter-controller/component.html?path=${path}`).pipe(pluck('response'));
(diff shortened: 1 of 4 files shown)


ui/app/src/services/content.ts
```diff
@@ -807,23 +812,14 @@ export function getChildrenByPath(
...
-  }
-): Observable<DetailedItem[]> {
-  const requests: Observable<DetailedItem>[] = [];
-  paths.forEach((path) =>
-    requests.push(
-      getDetailedItem(
-        site,
-        path === '/site/website' ? (options?.skipHomePathOverride ? path : '/site/website/index.xml') : path
-      )
+export function fetchItemsByPath(siteId: string, paths: string[], preferContent = true): Observable<DetailedItem[]> {
```
rart: LLeave the third arg as an object of options. Default it to true. Create an interface just like with get children.

ui/app/src/services/content.ts
```diff
@@ -807,23 +812,14 @@ export function getChildrenByPath(
...
+  const qs = toQueryString({ siteId, paths, preferContent });
+  return get(`/studio/api/2/content/sandbox_items_by_path${qs}`).pipe(
+    pluck('response', 'items'),
+    map((items) =>
+      items.map((item) => ({ ...parseSandBoxItemToDetailedItem(item), stateMap: createItemStateMap(item.state) }))
```
rart: Add options.castAsDetailedItem and only apply this if such option is true. Try to create the function overloads to specify when it returns Detailed or Sandbox

rart (changes requested):

rart (approved):

ui/app/src/services/content.ts
```diff
@@ -807,23 +813,29 @@ export function getChildrenByPath(
...
-        site,
-        path === '/site/website' ? (options?.skipHomePathOverride ? path : '/site/website/index.xml') : path
-      )
+  options: FetchItemsByPathOptions
+): Observable<DetailedItem[]>;
+export function fetchItemsByPath(
+  siteId: string,
+  paths: string[],
+  options?: FetchItemsByPathOptions
+): Observable<SandboxItem[] | DetailedItem[]> {
```
rart: Typings not sufficient for correct type deduction.

rart (approved):

ui/app/src/services/content.ts
```diff
@@ -807,23 +813,39 @@ export function getChildrenByPath(
...
+  siteId: string,
+  paths: string[],
+  options: FetchItemsByPathOptions
+): Observable<SandboxItem[]>;
+export function fetchItemsByPath(
+  siteId: string,
+  paths: string[],
+  options?: FetchItemsByPathOptions
+): Observable<SandboxItem[] | DetailedItem[]> {
+  const { castAsDetailedItem = false, preferContent = true } = options;
```
rart: Need to protect against options being undefined (i.e. can't restructure properties from undefined)
```suggestion
  const { castAsDetailedItem = false, preferContent = true } = options ?? {};
```

rart (approved):

PR #2154, opened 2021-02-08
[4509][3.1.x] RTE wrap for long lines of code or text

jvega190:
https://github.com/craftercms/craftercms/issues/4509


static-assets/modules/editors/tinymce/v5/tinymce/plugins/acecode/plugin.js
```diff
@@ -65,7 +65,9 @@
 				var dialog = document.getElementsByClassName('tox-dialog--width-lg')[0];
 				dialog.classList.add('fullscreen');
 
-				mce_editor = ace.edit('mce-ace-editor-block');
+				mce_editor = ace.edit('mce-ace-editor-block', {
+				  wrap: true
+        });
```
rart: @russdanner @sumerjabri I think this is a preference and I'm not sure which should be the default.
Some people don't like wraps and looks like we're changing behaviour that's been there for a long time.

We have some ace configuration one of the config xmls (don't recall which at the moment). Shall we add this one there?

sumerjabri: I think we add the option/config anyway.

@russdanner I'm in favor of the default being `wrap` because that's my personal preference, however, it's your call what should the default be.

I would have loved to see a button on ACE letting me choose in real-time.

jvega190:
Now it is configurable from the rte (tinymce 5) configuration file. I'll update the bps to include it. Default was set to false (current behavior), waiting for Russ' response.
@sumerjabri should the configuration update be handled by the Upgrade Manager too?

rart (approved):
We could merge and change the default if Russ comes back requesting the change of default.

PR #2186, opened 2021-02-22
[4539] Editing repeating groups with next ICE updates model/xml incorrectly

jvega190:
https://github.com/craftercms/craftercms/issues/4539

### DIFF
--- a/ui/guest/src/controls/rte.ts
+++ b/ui/guest/src/controls/rte.ts
@@ -113,7 +113,7 @@ export function initTinyMCE(
         function save() {
           const content = getContent();
           if (changed) {
-            contentController.updateField(record.modelId, field.id, record.index, content);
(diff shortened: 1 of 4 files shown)


ui/guest/src/controls/rte.ts
```diff
@@ -113,7 +113,7 @@ export function initTinyMCE(
         function save() {
           const content = getContent();
           if (changed) {
-            contentController.updateField(record.modelId, field.id, record.index, content);
+            contentController.updateField(record.modelId, record.fieldId[0], record.index, content);
```
jvega190, line 116: @rart the problem in here was that for example, when having a repeating group like:
`sections_o: [
     section_html,
     section_html, ...
]`
`field.id` would be `section_html`, but we need the full id (`record.fieldId` has the full value `sections_o.section_html`).

PR #2211, opened 2021-03-05
4.0.0/Stability

rart:
AuthMonitor restore logout process for correct redirection in SSO, Guest bugfixes


ui/app/src/models/GlobalState.ts
```diff
@@ -84,6 +84,15 @@ export interface GuestData {
   itemBeingDragged: boolean;
 }
 
+// TODO:
+//   Assses extracting these props from `GuestData` to avoid reloading models
```
sumerjabri: Typo, change to assess

sumerjabri (changes requested):

ui/app/src/models/GlobalState.ts
rart: Fixed.

sumerjabri (approved):

PR #2214, opened 2021-03-08
[4486][next] RTE 5 toolbar disappears when anything is clicked

jvega190:
https://github.com/craftercms/craftercms/issues/4486


rart (changes requested):
See comments on 3.1.x PR

rart (approved):

PR #2222, opened 2021-03-11
[studio-ui] Dark Theme issues

aachavarria:

### DIFF
--- a/ui/scss/src/_dark-mode.scss
+++ b/ui/scss/src/_dark-mode.scss
@@ -139,9 +139,9 @@ $dark-mode-input-border-focus: #007aff;
   }
 
   a {
-    color: $dark-mode-button-primary !important;
+    color: $dark-mode-button-primary;
     &:hover {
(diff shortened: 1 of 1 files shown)


ui/scss/src/_dark-mode.scss
```diff
@@ -307,7 +307,7 @@ $dark-mode-input-border-focus: #007aff;
     .acn-parent-folder,
     .acn-admin-console,
     .ygtvcell {
-      color: $dark-mode-font-color;
+      color: $dark-mode-font-color !important;
```
rart: Can we increment the specificity of the relevant selector instead of !important?

rart (changes requested):

ui/scss/src/_dark-mode.scss
aachavarria: no, because we have a dark: a: color !important; and this is needed to override the default blue color of the links

rart (approved):

PR #2230, opened 2021-03-22
pageBuilder updates

jvega190:
- Default value for date-time type
- Initialize validations as an empty object
- Validate that content exists


ui/app/src/utils/xml.ts
```diff
@@ -126,7 +126,7 @@ export function createElements(doc: XMLDocument, element: Element, data: object)
         console.error('[createElements] Path not implemented.');
       } else if (content instanceof Element) {
         elem.appendChild(content);
-      } else {
+      } else if (content) {
```
rart: This might be an issue with `0` or `""`, so best to be specific
```suggestion
      } else if (content !== null && content !== void 0) {
```

rart (approved):

PR #2236, opened 2021-03-25
[studio-ui] PathNavigators shared logic related refactors

aachavarria:
https://github.com/craftercms/craftercms/issues/4575


ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -419,6 +425,46 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
       })
     );
   };
+
+  const onPreview = (item: DetailedItem) => {
+    if (item.systemType === 'component' || item.systemType === 'taxonomy') {
+      const src = `${legacyFormSrc}site=${site}&path=${item.path}&type=form&readonly=true`;
```
rart: We should clean up all over moving this url construction to a single place. Possibly an epic or action builder or an util that one of those would invoke.

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -419,6 +425,46 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
...
+    } else if (item.mimeType.startsWith('image/')) {
+      dispatch(
+        showPreviewDialog({
+          type: 'image',
+          title: item.label,
+          url: item.path
+        })
+      );
+    } else {
+      fetchContentXML(site, item.path).subscribe((content) => {
```
rart, line 491: This here isn’t great either. 
Also, there’s no visibility of system status while loading the content.

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -419,6 +425,46 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
...
+        let mode = 'txt';
+
+        if (item.systemType === 'renderingTemplate') {
```
rart: All this mode selection logic needs to be encapsulated and made reusable.

rart (changes requested):

static-assets/components/cstudio-contextual-nav/quick-create.js
```diff
@@ -173,11 +172,12 @@ CStudioAuthoring.ContextualNav.WcmQuickCreate = CStudioAuthoring.ContextualNav.W
...
-        showEditDialog({
-          src,
-          inProgress: false
-        });
+      const onQuickCreateItemSelected = (props) => {
+        showEditDialog(
+          Object.assign(props, {
+            inProgress: false
+          })
+        );
```
rart, line 180: The changes don't look compatible. Previously receiving a `src` (string?) and now an object `props`. I don't see the counter part of this change. Was this broken?

ui/app/src/components/PathNavigator/utils.ts
```diff
@@ -17,23 +17,41 @@
...
 export function isNavigable(item: DetailedItem | SandboxItem): boolean {
-  return item.previewUrl !== null;
+  // Assets have a valid previewUrl but we don't want assets to show in the
+  // Guest iFrame but rather as a preview dialog.
+  return item.previewUrl !== null && item.systemType !== 'asset';
+}
+
+export function isEditableViaFormEditor(item: DetailedItem | SandboxItem): boolean {
+  return item.systemType === 'component' || item.systemType === 'taxonomy';
+}
```
rart: Problem is this function is lying because pages are editable via form too.

ui/app/src/modules/Preview/Tools/EditFormPanel.tsx
```diff
@@ -189,12 +189,38 @@ function EditFormPanelBody(props: EditFormPanelBodyProps) {
...
+
+      return {
+        authoringBase,
+        site,
+        path: parentPath,
+        isHidden: true,
+        modelId: selectedId
+      };
+    }
+  }, [authoringBase, childrenMap, model, models, path, selectedId, site]);
```
rart: Doesn't look like you're using this logic elsewhere. Why not do this directly inside openDialog?

ui/app/src/utils/itemActions.ts
```diff
@@ -726,12 +728,14 @@ export const itemActionDispatcher = ({
...
-        let src = `${legacyFormSrc}site=${site}&path=${encodeURIComponent(item.path)}&type=asset`;
+        let src = `${authoringBase}/legacy/form?site=${site}&path=${encodeURIComponent(item.path)}&type=asset`;
         dispatch(showCodeEditorDialog({ src }));
         break;
       }
       case 'viewCodeEditor': {
-        let src = `${legacyFormSrc}site=${site}&path=${encodeURIComponent(item.path)}&type=asset&readonly=true`;
+        let src = `${authoringBase}/legacy/form?site=${site}&path=${encodeURIComponent(
+          item.path
+        )}&type=asset&readonly=true`;
```
rart: Shouldn't these use `getEditFormSrc`?

ui/app/src/utils/path.ts
```diff
@@ -144,3 +145,38 @@ export function getPasteItemFromPath(path: string, paths: string[]): PasteItem {
...
+    site,
+    path,
+    type: 'form',
+    readonly,
+    isHidden,
+    modelId,
+    changeTemplate,
+    contentTypeId,
+    isNewContent
+  });
```
rart, line 180: Make sure query-string lib excludes undefined props

rart (changes requested):

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -205,7 +216,7 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
 
   useEffect(() => {
     keywordByPathRef.current = keywordByPath;
-  }, [keywordByPath]);
+  }, [keywordByPath, itemsByPath]);
```
rart: `itemsByPath` not used on effect. Why is it listed as a dependency?

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -241,8 +252,33 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
       switch (type) {
         case itemsPasted.type:
         case folderCreated.type: {
-          if (payload.clipboard.type === 'CUT') {
-            debugger;
+          if (payload.clipboard?.type === 'CUT') {
+            const sourceNode =
+              nodesByPathRef.current[getParentPath(payload.clipboard.sourcePath)] ??
+              nodesByPathRef.current[withIndex(getParentPath(payload.clipboard.sourcePath))];
```
rart: extract `getParentPath` call to a const to avoid double computation

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -241,8 +252,33 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
       switch (type) {
         case itemsPasted.type:
         case folderCreated.type: {
-          if (payload.clipboard.type === 'CUT') {
-            debugger;
+          if (payload.clipboard?.type === 'CUT') {
```
rart, line 255: Confirm: it should go to else statement whenever is not a cut **including** paste items via copy?
Otherwise, this condition needs to be revisited and rather use `type` to make the determination.

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -241,8 +252,33 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
...
+            const sourceNode =
+              nodesByPathRef.current[getParentPath(payload.clipboard.sourcePath)] ??
+              nodesByPathRef.current[withIndex(getParentPath(payload.clipboard.sourcePath))];
+            const targetNode =
+              nodesByPathRef.current[payload.target] ?? nodesByPathRef.current[withIndex(payload.target)];
```
rart: Should we be doing rather this:
```suggestion
              nodesByPathRef.current[withoutIndex(payload.target)] ?? nodesByPathRef.current[withIndex(payload.target)];
```
We don't know if an items is registered with or without index on the state so if not using both `withoutIndex` and `withIndex` we might be checking for the same thing and not getting the result we want.

We should make an util that does this
```ts
export function lookupItemByPath(path: string, lookupTable: LookupTable<Sandbox>): Sandbox
export function lookupItemByPath(path: string, lookupTable: LookupTable<Detailed>): Detailed 
export function lookupItemByPath(path: string, lookupTable: LookupTable<Sandbox | Detailed>): Sandbox | Detailed {
  return lookupTable[withoutIndex(path)] ?? lookupTable[withInde(path)];
}
```

Same deal would apply above to `sourceNode`. See if there's anywhere else that could use the new util.

rart (changes requested):

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
aachavarria: that is ok, it should go to the else statement for any operation excluding CUT, in this case COPY or folderCreated

ui/app/src/components/PathNavigatorTree/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import LookupTable from '../../models/LookupTable';
+import { withIndex, withoutIndex } from '../../utils/path';
+import { TreeNode } from './PathNavigatorTreeUI';
+
+export function lookupNodeByPath(path: string, lookupTable: LookupTable<TreeNode>): TreeNode {
+  return lookupTable[withoutIndex(path)] ?? lookupTable[withIndex(path)];
+}
```
rart:
```suggestion
export function lookupItemByPath<T = DetailedItem>(path: string, lookupTable: LookupTable<T>): T {
  return lookupTable[withoutIndex(path)] ?? lookupTable[withIndex(path)];
}
```

rart: This should be on utils/content

rart (approved):

PR #2238, opened 2021-03-25
pageBuilder updates

jvega190:
- Export fromTopic
- Fetch guest model on operation that require the model to update
- Add i18n to operation messages


ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -406,10 +458,11 @@ export function PreviewConcierge(props: any) {
               issueDescriptorRequest(path, fetchPrimaryGuestModelComplete);
             }
           } /* else if (type === FETCH_GUEST_MODEL) */ else {
-            if (payload.path?.startsWith('/')) {
-              issueDescriptorRequest(payload.path, fetchGuestModelComplete);
+            const path = payload?.path ? payload.path : guest.path;
```
rart: When would there ever not be a path on the payload? This action should always state what model needs loading.

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -146,6 +146,58 @@ const guestMessages = defineMessages({
   dropTargetsNotFound: {
     id: 'register.dropTargetsNotFound',
     defaultMessage: 'There are no drop targets for {contentType} components'
+  },
+  sortOperationComplete: {
+    id: 'operations.sortOperationComplete',
+    defaultMessage: 'Sort operation completed.'
```
rart, line 152: All these messages are a bit technical. We should revisit the copy.

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -574,15 +631,16 @@ export function PreviewConcierge(props: any) {
             parentModelId ? models[parentModelId].craftercms.path : null
           ).subscribe(
             () => {
+              guestToHost$.next({ type: FETCH_GUEST_MODEL });
```
rart: Don’t like the idea of using this channel for this purpose.
Let’s find a better way of reusing the loading logic.

rart (changes requested):

rart (approved):

PR #2241, opened 2021-03-27
Stop over-sizing node selector box when max is 1.

tfield:
Highly visible usability improvement in node selectors when only one item is allowed to be selected.


rart:
Thanks for the PR! Your contribution — with some adjustments — would be merged through [2253](https://github.com/craftercms/studio-ui/pull/2253).

PR #2258, opened 2021-04-05
[4620] Line numbers goes over the action bar of configurations

jvega190:
https://github.com/craftercms/craftercms/issues/4620


rart (approved):
@jvega190 the control bar is used in several places. Please do some code browsing to find places of usage and additional testing to make sure all keeps working on those views on their different state.

PR #2260, opened 2021-04-06
[4622] Create Repository dialog closes when you hit the <enter> key after inputting URL

jvega190:
https://github.com/craftercms/craftercms/issues/4622

### DIFF
--- a/static-assets/ng-views/admin-repository.html
+++ b/static-assets/ng-views/admin-repository.html
@@ -397,9 +397,10 @@
       >{{ 'common.CANCEL_LABEL' | translate }}
       </button>
       <button
+        id="createRepoSubmit"
         class="btn btn-primary text-capitalize"
         type="submit"
-        ng-disabled="createNameForm.$invalid"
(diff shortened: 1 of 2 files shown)


rart (approved):

static-assets/ng-views/admin-repository.html
```diff
@@ -392,13 +392,14 @@ <h3 class="modal-title">{{dialogTitle}}</h3>
...
+        type="button"
         class="btn btn-default" ng-click="hideModal()"
       >{{ 'common.CANCEL_LABEL' | translate }}
       </button>
       <button
         class="btn btn-primary text-capitalize"
-        ng-click="okModalFunction(repo)"
-        ng-disabled="createNameForm.$invalid"
+        type="submit"
+        ng-class="{'disabled': createNameForm.$invalid}"
```
jvega190, line 402: styles were not needed since disabled class matches bootstrap disabled styles (only styles, not behavior)

rart (approved):

PR #2265, opened 2021-04-13
Fixing issues about restoring state on Path Navigator Tree

aachavarria:

### DIFF
--- a/ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
+++ b/ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
@@ -16,15 +16,7 @@
 
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import PathNavigatorTreeUI, { TreeNode } from './PathNavigatorTreeUI';
-import {
-  useActiveSiteId,
-  useActiveUser,
-  useEnv,
-  useItemsByPath,
-  useMount,
-  useSelection,
-  useSubject
-} from '../../utils/hooks';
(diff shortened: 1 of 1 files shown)


ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -127,6 +135,13 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
 
   const dispatch = useDispatch();
 
+  useMount(() => {
+    if (state) {
+      // Restoring state from state;
+      state.expanded.forEach((path) => fetchingPathsRef.current.push(path));
+    }
+  });
+
```
rart: We don't need an effect for this.
```suggestion
if (state && fetchingPathsRef.current === null) {
  // Restoring previously loaded state from redux
  state.expanded.forEach((path) => fetchingPathsRef.current.push(path));
} else {
  fetchingPathsRef.current = [];
}
```
Just make sure to initialize the ref in null for this to work.

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -386,7 +401,7 @@ export default function PathNavigatorTree(props: PathNavigatorTreeProps) {
         })
       );
     } else {
-      if (childrenByParentPath[path]) {
+      if (nodesByPathRef[path]) {
```
rart: Call me to review these changes.

rart (changes requested):

rart (approved):

PR #2274, opened 2021-04-15
[4649][next] Add note in Remote Repositories not to use Studio as a git merge and conflict resolution platform

jvega190:
https://github.com/craftercms/craftercms/issues/4649

### DIFF
--- a/ui/app/src/utils/i18n-legacy.ts
+++ b/ui/app/src/utils/i18n-legacy.ts
@@ -497,6 +497,10 @@ export const words = defineMessages({
   select: {
     id: 'words.select',
     defaultMessage: 'Select'
+  },
+  note: {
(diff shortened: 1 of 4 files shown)


ui/app/src/utils/i18n-legacy.ts
```diff
@@ -491,6 +497,10 @@ export const words = defineMessages({
   select: {
     id: 'words.select',
     defaultMessage: 'Select'
+  },
+  note: {
+    id: 'words.note',
+    defaultMessage: 'Note'
```
rart: not being used

rart (changes requested):
Please ping me to review styles

rart (approved):

PR #2277, opened 2021-04-16
[4649][3.1.x] Add note in Remote Repositories not to use Studio as a git merge and conflict resolution platform

jvega190:
https://github.com/craftercms/craftercms/issues/4649

### DIFF
--- a/static-assets/ng-views/admin-repository.html
+++ b/static-assets/ng-views/admin-repository.html
@@ -223,6 +223,14 @@
       </div>
     </div>
   </div>
+  <div class="row">
+    <div class="col-md-12 tac">
+      <div class="note tal">
+        <span class="note-title">{{repositories.repoMessages.repositoriesNoteLabel}}:</span>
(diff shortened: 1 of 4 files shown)


ui/app/src/utils/i18n-legacy.ts
```diff
@@ -439,7 +445,7 @@ const words = defineMessages({
   reason: {
     id: 'words.reason',
     defaultMessage: 'Reason'
-  }
+  },
```
rart: 👆 x trailing comma

Did you remove prettier auto format on save?

rart (changes requested):

static-assets/ng-views/admin-repository.html
```diff
@@ -151,6 +152,16 @@ <h5>{{repositories.repoMessages.unreachableRemote(repo.name)}}</h5>
           </a>
         </div>
 
+        <div class="mb30">
```
jvega190, line 155: This is under a div that already hides when no conflicts are present

rart (approved):

PR #2294, opened 2021-04-26
Fix pathNavigator title definition issue

jvega190:


ui/app/src/components/SystemStatus/EmptyState.tsx
```diff
@@ -65,7 +65,10 @@ export default function EmptyState(props: EmptyStateProps) {
   const classes = useStyles(props.styles);
   const { formatMessage } = useIntl();
   const { image = emptyImage, classes: propClasses, children } = props;
-  const title = React.isValidElement(props.title) ? props.title : formatMessage(props.title as MessageDescriptor);
+  const title =
+    React.isValidElement(props.title) || typeof props.title === 'string'
+      ? props.title
+      : formatMessage(props.title as MessageDescriptor);
   const subtitle =
     React.isValidElement(props.subtitle) || nou(props.subtitle)
```
rart: Same fix as above needs to be applied here.
What happened with react-is?
Make sure you tested this accepts string, number or a JSX element but not Component functions (should be instances)

rart (changes requested):

ui/app/src/components/SystemStatus/EmptyState.tsx
jvega190: react-is behaves the same as isValidElement (strings are not valid). For numbers, I'd have to add it to the condition (string || number). Should I add it?

rart (approved):

PR #2304, opened 2021-05-05
[4322][3.1.x] Bugfix/bulk upload nested failed

phuongnq:
A suggestion fix for issue https://github.com/craftercms/craftercms/issues/4322

While checking file existing, we are checking filename which is not correct for nested structure. Better to use relativePath to identify file with same name but different path.

### DIFF
--- a/ui/app/src/components/BulkUpload.tsx
+++ b/ui/app/src/components/BulkUpload.tsx
@@ -352,8 +352,8 @@ const DropZone = React.forwardRef((props: DropZoneProps, ref: any) => {
       });
   }
 
-  function checkFileExist(newFile: File) {
-    return !uppy.getFiles().some((file) => file.name === newFile.name && file.type === newFile.type);
+  function checkFileExist(newFile: File & { relativePath?: string }) {
+    return !uppy.getFiles().some((file) => file.meta.relativePath === newFile.relativePath && file.type === newFile.type);
   }
 
   function removeDragData(event: React.DragEvent<HTMLElement>) {


ui/app/src/components/BulkUpload.tsx
```diff
@@ -352,8 +352,8 @@ const DropZone = React.forwardRef((props: DropZoneProps, ref: any) => {
       });
   }
 
-  function checkFileExist(newFile: File) {
-    return !uppy.getFiles().some((file) => file.name === newFile.name && file.type === newFile.type);
+  function checkFileExist(newFile: File & { relativePath?: string }) {
+    return !uppy.getFiles().some((file) => file.meta.relativePath === newFile.relativePath && file.type === newFile.type);
```
rart, line 356: Why is the relative path in different locations of the file objects? (i.e. file.meta.relativePath vs file.relativePath)

phuongnq: The original logic is from here:
```
interval(50)
      .pipe(takeUntil(cancelRequestObservable$), take(files.length))
      .subscribe((index) => {
        const file: File & { relativePath?: string } = files[index];
        checkFileExist(file) &&
          uppy.addFile({
            name: file.name,
            type: file.type,
            data: file,
            meta: {
              relativePath: file.relativePath || null
            }
          });
      });
```

We only call `checkFileExist` function from here.
As you can see, the variable `file` is of `UppyFile` interface with `meta.relativePath`, the `newFile` is of type `File` so they have different location.

rart:
Holding this on my plate to check & address the inconsistent types that we've discovered.

rart:
Merged @ https://github.com/craftercms/studio-ui/pull/2312

PR #2307, opened 2021-05-06
[4059][3.1.x] Fix form engine close exception

jvega190:
https://github.com/craftercms/craftercms/issues/4059

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -1367,6 +1367,8 @@ var CStudioForms =
                 CStudioAuthoring.Service.createServiceUri(serviceUrl),
                 {
                   success: function () {
+                    YAHOO.util.Event.removeListener(window, 'beforeunload', unloadFn, me);
+
                     var getContentItemCb = {
                       success: function (contentTO) {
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -1534,7 +1536,7 @@ var CStudioForms =
               path = CStudioAuthoring.Utils.getQueryVariable(location.search, 'path');
               if (path && path.indexOf('.xml') != -1) {
                 var entityId = buildEntityIdFn(null);
-                CStudioAuthoring.Service.unlockContentItemSync(CStudioAuthoringContext.site, entityId);
+                CrafterCMSNext.services.content.unlock(CStudioAuthoringContext.site, entityId).subscribe();
```
rart, line 1539: When is this executed? Is there a possibility of the form window closing before this request completes?

jvega190: this is called in the beforeunload event

rart: This may be a bit of a raise condition but will roll with it as there's no other way of handling this case

rart (approved):

PR #2308, opened 2021-05-06
[4690][3.1.x] Form engine - "view" mode does not allow inspection of embedded items

jvega190:
https://github.com/craftercms/craftercms/issues/4690


jvega190:
There's one detail that I need to discuss before making it ready for review.

rart (approved):

PR #2330, opened 2021-05-20
[4724] Only system_admin users should create/delete sites

aachavarria:
https://github.com/craftercms/craftercms/issues/4724
https://github.com/craftercms/craftercms/issues/4712
https://github.com/craftercms/craftercms/issues/4657


ui/app/src/components/PreviewAddressBar/PreviewAddressBar.tsx
```diff
@@ -155,7 +155,7 @@ export function AddressBar(props: AddressBarProps) {
   return (
     <>
       <Tooltip title={<FormattedMessage id="previewAddressBar.reloadButtonLabel" defaultMessage="Reload this page" />}>
-        <IconButton title={formatMessage(translations.reload)} onClick={onRefresh}>
+        <IconButton onClick={onRefresh}>
```
rart, line 146: Removing these means formatMessage and the translations are now of no use. Must clean that up.

ui/app/src/components/GlobalAppToolbar/GlobalAppToolbar.tsx
```diff
@@ -126,9 +126,11 @@ export const GlobalAppToolbar = React.memo<GlobalAppToolbarProps>(function(props
         )}
         {startContent}
         <section className={clsx(classes.title, props.classes?.title)}>
-          <Typography variant="h5" component="h1" className={classes.ellipsis}>
-            {title}
-          </Typography>
+          {title && (
```
rart: If now they are both optional, we should wrap the section in a conditional too

ui/app/src/components/SitesManagement/SitesManagement.tsx
```diff
@@ -76,6 +77,7 @@ export default function SitesManagement() {
   const isFetching = sitesBranch.isFetching;
   const [publishingStatusLookup, setPublishingStatusLookup] = useSpreadState<LookupTable<PublishingStatus>>({});
   const [selectedSiteStatus, setSelectedSiteStatus] = useState<PublishingStatus>(null);
+  const [permissionsLookup, setPermissionsLookup] = useState(null);
```
rart: Would be easier to set to {} or an object with the 3 permissions set to false so later on you needn't check for nullishness

ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -180,8 +180,8 @@ const reducer = createReducer<LookupTable<PathNavigatorTreeStateProps>>(
         });
         total[path] = data[path].total;
 
-        if (!keywordByPath[path] && children[path].length) {
-          // If the expanded node has no children and is not filtered, it's a leaf node and there's no point keeping it in `expanded`
+        if (keywordByPath[path] || children[path].length) {
+          // If the expanded node has no children or is not filtered, it's a leaf node and there's no point keeping it in `expanded`
```
rart: Either the condition or the description is wrong. Please double check/fix.
What is driving this change?

rart (changes requested):
This PR contains work for other tickets as well. Please link the other relevant tickets on the description - reactify global app (sites) & reactify site tools (config).

ui/app/src/components/PreviewAddressBar/PreviewAddressBar.tsx
aachavarria: I think no, because the tooltip wrapping this icon is using the translation

ui/app/src/components/SitesManagement/SitesManagement.tsx
aachavarria: yes but, the resource is checking for nullishness, if I set it to {}, the resource needs to do a Object.values, Which is better?

ui/app/src/state/reducers/pathNavigatorTree.ts
aachavarria: doing changes on the createPresenceTable, and the description is wrong

ui/app/src/components/SitesManagement/SitesManagement.tsx
rart: Create an empty object on object utils called `foo` (out of the component) and compare equality instead of nullishness on the resource.

ui/app/src/state/reducers/pathNavigatorTree.ts
rart: don't understand your comment

rart (approved):

PR #2343, opened 2021-05-26
[4737] Editing allowed when item is locked via PageBuilder 

aachavarria:
https://github.com/craftercms/craftercms/issues/4737


ui/app/src/utils/content.ts
```diff
@@ -124,6 +124,24 @@ export function isImage(path: string): boolean {
...
+export function getComputedEditMode({
+  item,
+  username,
+  editMode
+}: {
+  item: DetailedItem;
+  username: string;
+  editMode: boolean;
+}): boolean {
+  return isItemLockedForMe(item, username) === true || hasEditAction(item.availableActions) === false
```
rart, line 140: Item could be null
```suggestion
  return isItemLockedForMe(item, username) === true || hasEditAction(item?.availableActions) === false
```

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -315,19 +316,22 @@ export function PreviewConcierge(props: any) {
...
-  }, [dispatch, currentItemPath, site]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [items[currentItemPath], editMode, user.username]);
 
   useEffect(() => {
-    if (write === false && editMode) {
-      getHostToGuestBus().next({ type: HOST_CHECK_IN, payload: { editMode: false } });
+    if (currentItemPath && site) {
+      dispatch(completeDetailedItem({ path: currentItemPath }));
     }
```
rart: Could we omit re-loading the detailed item?

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -438,7 +442,10 @@ export function PreviewConcierge(props: any) {
               dispatch(setPreviewChoice({ site, choice: '2' }));
             }
 
-            getHostToGuestBus().next({ type: HOST_CHECK_IN, payload: { editMode, highlightMode } });
+            getHostToGuestBus().next({
+              type: HOST_CHECK_IN,
+              payload: { editMode: false, highlightMode }
```
rart, line 439: How come editMode is hard set to false here?

rart (changes requested):

ui/app/src/utils/content.ts
aachavarria: this function should no receive a null item

aachavarria: it is not in the interface, the component using it should validate the item

ui/app/src/modules/Preview/PreviewConcierge.tsx
aachavarria: we talked about this yesterday, guest is going to start false and the effect on the top of previewConcierge is calculating the getComputedEditMode, and updating Guest

rart (approved):

PR #2346, opened 2021-05-27
Fix GlobalApp race condition with root url

rart:


rart:
@sumerjabri merging now for a community support request

PR #2353, opened 2021-05-31
[4675] Create publish history table and use that to report

dejan-brkic:
Updated url for deployment history

https://github.com/craftercms/craftercms/issues/4675

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -3018,7 +3018,7 @@ var nodeOpen = false,
 
       // DEPLOYMENT SERVICES
       // READ OPS
-      getDeploymentHistoryServiceUrl: '/api/1/services/api/1/deployment/get-deployment-history.json',
+      getDeploymentHistoryServiceUrl: '/api/2/publish/history.json',
       getScheduledItemsServiceUrl: '/api/1/services/api/1/deployment/get-scheduled-items.json',
       getDependenciesServiceUrl: '/api/1/services/api/1/dependency/get-dependencies.json',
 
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -3018,7 +3018,7 @@ var nodeOpen = false,
 
       // DEPLOYMENT SERVICES
       // READ OPS
-      getDeploymentHistoryServiceUrl: '/api/1/services/api/1/deployment/get-deployment-history.json',
+      getDeploymentHistoryServiceUrl: '/api/2/publish/history.json',
```
rart, line 3021: This const is not used anymore. We can get rid of it.

ui/app/src/services/dashboard.ts
```diff
@@ -109,5 +109,5 @@ export function fetchLegacyDeploymentHistory(
     ...(filterBy && { filterType: filterBy })
   });
 
-  return get(`/studio/api/1/services/api/1/deployment/get-deployment-history.json${qs}`).pipe(pluck('response'));
+  return get(`/studio/api/2/publish/history.json${qs}`).pipe(pluck('response'));
```
rart, line 112: Is API 2 returning same response format?

rart:
Was the old/api1 removed completely? Will calling it redirect or fail?

ui/app/src/services/dashboard.ts
dejan-brkic: it is responding in the same format as 3.1

sumerjabri (approved):

PR #2365, opened 2021-06-04
[4752] The date and time control in content type is not displaying in content form except when "Use custom timezone" property is checked

jvega190:
https://github.com/craftercms/craftercms/issues/4752


rart (approved):
@jvega190 does this not apply to 3.1? Why?

jvega190:
> @jvega190 does this not apply to 3.1? Why?

@rart the issue was with a wrapper element that was added in 4.0.x for the styles update. So the element doesn't exist in 3.1

rart (approved):

PR #2372, opened 2021-06-09
Update legacy item parse functions

jvega190:
- Update legacy item parse functions
 - Move staged style in ItemPublishingTargetIcon before the live style for live to have higher precedence (when both live and staged are true)

### DIFF
--- a/ui/app/src/components/ItemPublishingTargetIcon/ItemPublishingTargetIcon.tsx
+++ b/ui/app/src/components/ItemPublishingTargetIcon/ItemPublishingTargetIcon.tsx
@@ -44,13 +44,13 @@ const useStyles = makeStyles(() =>
       color: palette.gray.medium2,
       ...styles.root
     }),
-    publishingTargetLive: (styles) => ({
-      color: palette.green.main,
-      ...styles.publishingTargetLive
-    }),
     publishingTargetStaged: (styles) => ({
       color: palette.blue.main,
       ...styles.publishingTargetStaged
+    }),
+    publishingTargetLive: (styles) => ({
+      color: palette.green.main,
+      ...styles.publishingTargetLive
     }),
     publishingIcon: (styles) => ({
       ...styles.publishingIcon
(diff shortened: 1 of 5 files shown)


ui/app/src/components/ItemPublishingTargetIcon/ItemPublishingTargetIcon.tsx
```diff
@@ -44,14 +44,14 @@ const useStyles = makeStyles(() =>
       color: palette.gray.medium2,
       ...styles.root
     }),
-    publishingTargetLive: (styles) => ({
-      color: palette.green.main,
-      ...styles.publishingTargetLive
-    }),
```
rart: What's change this for?

ui/app/src/utils/state.ts
```diff
@@ -29,11 +29,11 @@ export function getStoredGlobalMenuSiteViewPreference(user: string): 'grid' | 'l
 
 export function getStateMapFromLegacyItem(item: LegacyItem): ItemStateMap {
   return {
-    live: false,
+    live: item.endpoint === 'live',
     locked: false,
     modified: false,
```
rart: There must be a way of knowing these as the old ui shows the lock and pencil icons

rart (changes requested):
Please also re-create the bitmap. Theres already a function that calculates a value based on the stateMap object

ui/app/src/components/ItemPublishingTargetIcon/ItemPublishingTargetIcon.tsx
jvega190: When both live and staged are true (as we discussed yesterday), both styles apply to the PublishingTargetIcon component, and since the live styles were before the staging, it was showing the color for staging.

rart: We don't want both classes applied in those cases so instead of this, let's update the logic below to only include one. Let's roll this back.

rart:
```js
item.stateMap.live ? classes.publishingTargetLive : item.stateMap.staged ? classes.publishingTargetStaged : false
```

rart (approved):

PR #2374, opened 2021-06-10
[4774] Reactify Recent Activity Dashboard Widget

jvega190:
https://github.com/craftercms/craftercms/issues/4774


rart (changes requested):
Is this ready?

ToDo:
- Integrate with dashboard via ui.xml
- use Intl.DateTimeFormat abstraction function

rart (approved):

PR #2382, opened 2021-06-17
[4783] PageBuilder RTE enhancements for Studio v4.0.0

aachavarria:
https://github.com/craftercms/craftercms/issues/4783


ui/guest/src/controls/rte.ts
```diff
@@ -33,7 +33,6 @@ export function initTinyMCE(
   const { field } = iceRegistry.getReferentialEntries(record.iceIds[0]);
   const type = field?.type;
   const plugins = ['paste'];
-  type === 'html' && plugins.push('quickbars');
```
rart, line 36: Plain text should not show a toolbar or allow for any options.

ui/guest/src/react/ZoneMarker.tsx
```diff
@@ -33,6 +33,7 @@ export function ZoneMarker(props: ZoneMarkerProps) {
   useEffect(() => {
     setZoneStyle(getZoneMarkerStyle(rect));
   }, [rect]);
+
```
rart: Rollback so we don't have a dummy change on the change log

rart (changes requested):

rart (approved):

PR #2395, opened 2021-06-23
[studio-ui] Parsing xml issues related with unicodes values 

aachavarria:

### DIFF
--- a/ui/app/src/services/configuration.ts
+++ b/ui/app/src/services/configuration.ts
@@ -65,10 +65,10 @@ export function fetchConfigurationJSON(
         parseNodeValue: false,
         tagValueProcessor: (value) =>
           value
-            .replaceAll(/&lt;/g, '<')
-            .replaceAll(/&gt;/g, '>')
-            .replaceAll(/&quot;/g, '"')
-            .replaceAll(/&amp;/g, '&')
+            .replace(/&lt;/g, '<')
+            .replace(/&gt;/g, '>')
+            .replace(/&quot;/g, '"')
+            .replace(/&amp;/g, '&')
       });
     })
   );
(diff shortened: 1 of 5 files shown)


ui/app/src/utils/string.ts
```diff
@@ -157,3 +157,11 @@ export function getSimplifiedVersion(version: string, options: { minor?: boolean
...
   return pieces.join('.');
 }
+
+export function unEscapeXml(value: string): string {
+  return value
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&amp;/g, '&');
+}
```
rart: Going forward, we're not going to be doing these transforms so it'd rather not encourage them by having this util (unless in the future we find a good reason for this). Let's do it inline in the tag value processor.

ui/app/src/services/configuration.ts
```diff
@@ -60,11 +61,14 @@ export function fetchConfigurationJSON(
...
-      deserialize(conf, {
-        parseNodeValue: false
-      })
-    )
+    map((conf) => {
+      return deserialize(conf, {
+        parseNodeValue: false,
+        tagValueProcessor: (value) => {
+          return unEscapeXml(value);
+        }
```
rart:
```suggestion
        tagValueProcessor: (value) => value
          .replace(/&lt;/g, '<')
          .replace(/&gt;/g, '>')
          .replace(/&quot;/g, '"')
          .replace(/&amp;/g, '&')
```

rart (approved):

rart (changes requested):
See comments above.
(Meant to press request changes instead of approve 😅)

rart (approved):

PR #2412, opened 2021-07-07
[4651] [studio] Add keyword parameter to the Search User API

aachavarria:
https://github.com/craftercms/craftercms/issues/4651

### DIFF
--- a/ui/app/src/services/users.ts
+++ b/ui/app/src/services/users.ts
@@ -49,9 +49,8 @@ export function trash(username: string): Observable<true> {
   return del(`/studio/api/2/users?username=${encodeURIComponent(username)}`).pipe(mapTo(true));
 }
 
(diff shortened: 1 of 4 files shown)


ui/app/src/services/users.ts
```diff
@@ -49,8 +49,9 @@ export function trash(username: string): Observable<true> {
   return del(`/studio/api/2/users?username=${encodeURIComponent(username)}`).pipe(mapTo(true));
 }
 
-export function fetchAll(options?: PaginationOptions): Observable<PagedArray<User>> {
+export function fetchAll(keyword?: string, options?: PaginationOptions): Observable<PagedArray<User>> {
```
rart: Add it to options argument.
Don't modify the pagination options interface just add `& { keyword?: string }`

rart (changes requested):

rart (approved):

PR #2429, opened 2021-07-13
[4836] Global config sample file Append after current content does not work

jvega190:
https://github.com/craftercms/craftercms/issues/4836

### DIFF
--- a/ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
+++ b/ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
@@ -89,7 +89,8 @@ export default function GlobalConfigManagement() {
     if (type === 'replace') {
       setContent(sample);
     } else {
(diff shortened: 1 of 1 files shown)


ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
```diff
@@ -89,7 +89,8 @@ export default function GlobalConfigManagement() {
     if (type === 'replace') {
       setContent(sample);
     } else {
-      setContent(content + sample);
+      const currentContent = aceEditorRef.current.getValue();
```
jvega190, line 92: @rart the issue was that if you edit the editor the content is not set on each update (I guess that it's to avoid calling n times setContent while typing). So, I'm getting the content that will be append to the sample directly from the editor value

rart (approved):

PR #2437, opened 2021-07-14
[4829][next] Browse interface does not display folders in tree nav with no child items

jvega190:
https://github.com/craftercms/craftercms/issues/4829

### DIFF
--- a/static-assets/components/cstudio-browse/browse.js
+++ b/static-assets/components/cstudio-browse/browse.js
@@ -197,7 +197,7 @@
     var parsed = JSON.parse(JSON.stringify(obj), function(key, value) {
       if (key === 'children') {
         $.each(value, function(index, elem) {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-browse/browse.js
```diff
@@ -197,7 +197,7 @@
     var parsed = JSON.parse(JSON.stringify(obj), function(key, value) {
       if (key === 'children') {
         $.each(value, function(index, elem) {
-          if (elem.numOfChildren === 0) {
+          if (!elem.folder) {
```
rart: Would pages with inner children be ignored by this?

jvega190: That's correct. I updated it to show empty folders and also pages with inner children

rart (approved):

PR #2463, opened 2021-07-29
[4857] Remove RTE 2 from Studio 4

jvega190:
https://github.com/craftercms/craftercms/issues/4857


static-assets/components/cstudio-admin/mods/content-type-propsheet/variable.js
```diff
@@ -66,7 +66,6 @@ YAHOO.extend(
             case 'transcoded-video-picker':
               addPostfixes = '_o';
               break;
-            case 'rte':
```
rart: Might be best to leave `rte` as the name as the plan is to keep one and keep it up to date

rart (changes requested):

rart (approved):

PR #2464, opened 2021-07-29
[4825] Editor for content type controller newline behavior

jvega190:
https://github.com/craftercms/craftercms/issues/4825

### DIFF
--- a/static-assets/components/cstudio-forms/template-editor.js
+++ b/static-assets/components/cstudio-forms/template-editor.js
@@ -395,7 +395,7 @@ CStudioAuthoring.Module.requireModule(
 
                         aceEditor.getSession().on('change', function() {
                           aceEditor.isModified = true;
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/template-editor.js
```diff
@@ -395,7 +395,7 @@ CStudioAuthoring.Module.requireModule(
 
                         aceEditor.getSession().on('change', function() {
                           aceEditor.isModified = true;
-                          onSaveCb.pendingChanges(true);
+                          onSaveCb.pendingChanges?.(true);
```
jvega190, line 398: When opening the editor from the content-type, it doesn't use the embedded_legacy_form, so it doesn't need/use pendingChanges. That was causing the strange behavior

rart (approved):

PR #2472, opened 2021-08-03
[2864] Only inactives cluster nodes should have the option to be removed from cluster

jvega190:
https://github.com/craftercms/craftercms/issues/2864

### DIFF
--- a/ui/app/src/components/ClusterGrid/ClusterGridUI.tsx
+++ b/ui/app/src/components/ClusterGrid/ClusterGridUI.tsx
@@ -94,7 +94,7 @@ export default function ClusterGridUI(props: ClusterGridProps) {
                 {cluster.gitAuthType}
               </GlobalAppGridCell>
               <GlobalAppGridCell align="left" className="width10">
-                {cluster.state === 'ACTIVE' && (
+                {cluster.state === 'INACTIVE' && (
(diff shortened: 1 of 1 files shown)


ui/app/src/components/ClusterGrid/ClusterGridUI.tsx
```diff
@@ -94,19 +94,21 @@ export default function ClusterGridUI(props: ClusterGridProps) {
...
-                    <FormattedMessage id="clusterGrid.confirmDeleteCluster" defaultMessage="Delete this cluster?" />
-                  }
-                  iconTooltip={<FormattedMessage id="clusterGrid.deleteCluster" defaultMessage="Delete cluster" />}
-                  icon={DeleteRoundedIcon}
-                  iconColor="action"
-                  onConfirm={() => {
-                    onDeleteCluster(cluster);
-                  }}
-                />
+                {cluster.state === 'ACTIVE' && (
```
rart: Condition is backwards. Only **INACTIVE** clusters should be removable.
```suggestion
                {cluster.state === 'INACTIVE' && (
```
Check screenshot on ticket to verify the right string.

rart (changes requested):

rart (approved):

PR #2478, opened 2021-08-05
[263E] Hide the Edit column from all dashboards if no Write permission is present

jvega190:
https://github.com/craftersoftware/craftercms/issues/263


jvega190:
Closing this PR because another approach will be taken

PR #2480, opened 2021-08-06
[263E] Hide the Edit column from all dashboards if no Write permission is present

jvega190:
https://github.com/craftersoftware/craftercms/issues/263


rart (commented):
- Still in draft stage. Is this not finished?
- Once finished, lets port to 4 too.

rart (approved):

PR #2499, opened 2021-08-18
[studio-ui] Path navigator update

aachavarria:
[studio-ui] Level descriptor as a child of an article doesn't list in the sidebar as expected

https://github.com/craftercms/craftercms/issues/4945

### DIFF
--- a/ui/app/src/components/PathNavigator/PathNavigator.tsx
+++ b/ui/app/src/components/PathNavigator/PathNavigator.tsx
@@ -73,6 +73,7 @@ import { useItemsByPath } from '../../utils/hooks/useItemsByPath';
 import { useSubject } from '../../utils/hooks/useSubject';
 import { useSiteLocales } from '../../utils/hooks/useSiteLocales';
 import { useMount } from '../../utils/hooks/useMount';
+import { nnou } from '../../utils/object';
 
 interface Menu {
   path?: string;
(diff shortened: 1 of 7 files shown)


ui/app/src/state/epics/pathNavigator.ts
```diff
@@ -57,6 +57,8 @@ export default [
           pathNavigatorFetchParentItems({
             id,
             path: storedState ? storedState.currentPath : payload.path,
+            ...(storedState?.offset && { offset: storedState.offset }),
```
rart: storedState.offset being 0 would result in a false which shouldn't

ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -167,6 +168,12 @@ export default function PathNavigator(props: PathNavigatorProps) {
   const siteLocales = useSiteLocales();
   const hasActiveSession = useSelection((state) => state.auth.active);
 
+  useEffect(() => {
+    if (nnou(state?.keyword)) {
+      setKeyword(state.keyword);
```
rart, line 173: Could we use state.keyword and not have a duplicate local state keyword var?

ui/app/src/state/reducers/pathNavigator.ts
```diff
@@ -164,6 +170,10 @@ const reducer = createReducer<LookupTable<PathNavigatorStateProps>>(
           itemsInPath: children.map((item) => item.path),
           levelDescriptor: children.levelDescriptor?.path ?? null,
           breadcrumb: getIndividualPaths(withoutIndex(currentPath), withoutIndex(rootPath)),
+          leaves: [
+            ...state[id].leaves,
+            ...children.filter((item) => item.childrenCount === 0).map((item) => item.path)
```
rart: Create a local util fn that receives children and returns an array of leaf paths and use it across all these instances

rart (changes requested):

ui/app/src/components/PathNavigator/PathNavigator.tsx
aachavarria: no at all, we need to keep one keyword value updated(the one the user should see in the pathNav) and another one debounced( the one is sent to the epics to fetch the data)

rart (approved):

PR #2520, opened 2021-08-30
[4981] Disabled pages have no indications it is disabled when previewing or in the dashboard

jvega190:
https://github.com/craftercms/craftercms/issues/4981


rart (approved):

rart:
You still need to add disabled to the icon guide, @jvega190

PR #2523, opened 2021-08-31
[4983] [3.1.x] Admin is considered as reviewer

phuongnq:
https://github.com/craftercms/craftercms/issues/4983

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -8386,33 +8386,20 @@ var nodeOpen = false,
       isReviewer: function (cb) {
         var callback = {
           success: function (data) {
-            var roles = data,
-              isRev = false,
-              topRoles = false;
-            for (var i = 0; i < roles.length; i++) {
-              if (
-                roles[i].toLocaleLowerCase() == 'admin' ||
-                roles[i].toLocaleLowerCase() == 'developer' ||
-                roles[i].toLocaleLowerCase() == 'publisher' ||
-                roles[i].toLocaleLowerCase() == 'author'
-              ) {
-                topRoles = true;
-              }
-              if (roles[i].toLocaleLowerCase() == 'reviewer' && !topRoles) {
-                isRev = true;
-                break;
-              } else {
-                isRev = false;
-              }
-            }
-            cb(isRev);
+            cb(!CStudioAuthoring.Utils.hasPerm(CStudioAuthoring.Constants.PERMISSION_WRITE, data.permissions));
           },
-          failure: function (response) {
-            console.log(response);
+          failure: function (err) {
+            console.log(err);
           }
         };
 
-        CStudioAuthoring.Service.getUserRoles(callback);
+        var path = '/site'; // default to check permissions on site
+        var selectedContents = CStudioAuthoring.SelectedContent.getSelectedContent();
+        if (selectedContents && selectedContents.length > 0 && selectedContents[0].uri) {
+          path = selectedContents[0].uri;
+        }
+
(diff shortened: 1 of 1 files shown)


rart (commented):
@russdanner,

when is a user considered reviewer? This method's logic is a bit odd. I'd recommend something like
```js
isReviewer: function(cb) {
  CStudioAuthoring.Service.getUserRoles({
    success: function(roles) {
      cb(!roles.some((role) => role !== 'reviewer') && roles.includes('reviewer'));
    },
    ...
```
or
```js
isReviewer: function(cb) {
  CStudioAuthoring.Service.getUserRoles({
    success: function(roles) {
      cb(roles[0] === 'reviewer');
    },
    ...
```
Is the idea that the user **only** has the reviewer role? Looks like that's what this revision would do.

This changes the semantics of how this works from answering _does the user have the reviewer role before all other roles?_ to _does the user have only the reviewer role?_, which could impact something else so please review and consider. Please review and consider changes.

As far as I can see, these are the places where this is used:
image

Also, as a side note, I don't see the point of this using `toLocaleLowerCase` as opposed to `toLowerCase`.

phuongnq:
For a Studio instance, users can always add more roles (https://docs.craftercms.org/en/3.1/system-administrators/activities/authoring/users-groups/putting-it-all-together.html#create-a-new-role).

So I think changing to check if user has only `reviewer` role is not enough.  I think the original logic is that, if user does not have top level roles and has 'reviewer' role than the user is a reviewer. This is what this PR does.

For the error, currently we have this issue: 

Since backend use HashMap to get all roles of admin, they are not in ordered. Sometimes we get

`['admin', 'developer', 'reviewer']`, sometimes `['reviewer', 'developer', 'admin']`.

The current bug is that if the set is:

* `['reviewer', 'developer', 'admin']` => user is reviewer

* `['admin', 'developer', 'reviewer']` => user is not reviewer

russdanner:
I don't think we should be checking for role name at all.  The UI should not care about role names, just permissions.  A role is just a way to group a set of permissions to a path. @rart @phuongnq

russdanner (changes requested):
Code should check for write permission, not role.

phuongnq:
@russdanner  @rart 
I updated to check if user has 'write' permission. If no write permission for previewing path, consider as 'reviewer'

static-assets/components/cstudio-common/common-api.js
```diff
@@ -8374,33 +8374,29 @@ var nodeOpen = false,
...
-                roles[i].toLocaleLowerCase() == 'author'
-              ) {
-                topRoles = true;
-              }
-              if (roles[i].toLocaleLowerCase() == 'reviewer' && !topRoles) {
-                isRev = true;
+            var permissions = data.permissions;
+            var hasWritePermission = false;
+            for (var i = 0; i < permissions.length; i++) {
+              if (permissions[i].toLowerCase() === 'write') {
```
sumerjabri: @rart we must put these permissions into constants as they change over time.

sumerjabri (changes requested):

static-assets/components/cstudio-common/common-api.js
sumerjabri: @phuongnq let's see what @rart says about the constant and then decide what to do.

phuongnq: @sumerjabri I updated to use constant defined in `common-api.js`. Kindly check this again.

sumerjabri (approved):

sumerjabri:
@rart please let me know if this is approved for merging or not yet.
CC @russdanner

rart (approved):

PR #2534, opened 2021-09-06
[4972] Improve Search behavior in Browse Components tool of Page Builder

aachavarria:
BrowseComponentsPanel: Moving styles and translation to his own file.
Updates on BrowseComponents Panel to use all contentTypes filter.
fetchItemsByContentType service: fixing service to return only components of the selected contentType;

https://github.com/craftercms/craftercms/issues/4972

### DIFF
--- a/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
+++ b/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
@@ -21,8 +21,6 @@ import { nnou, pluckProps } from '../../utils/object';
 import { ErrorBoundary } from '../SystemStatus/ErrorBoundary';
 import LoadingState from '../SystemStatus/LoadingState';
 import ContentInstance from '../../models/ContentInstance';
-import { DraggablePanelListItem } from '../../modules/Preview/Tools/DraggablePanelListItem';
-import List from '@material-ui/core/List';
 import {
   COMPONENT_INSTANCE_DRAG_ENDED,
   COMPONENT_INSTANCE_DRAG_STARTED,
@@ -32,19 +30,17 @@ import {
 } from '../../state/actions/preview';
 import { useDispatch } from 'react-redux';
 import SearchBar from '../Controls/SearchBar';
-import EmptyState from '../SystemStatus/EmptyState';
-import TablePagination from '@material-ui/core/TablePagination';
 import { getHostToGuestBus } from '../../modules/Preview/previewContext';
 import Select from '@material-ui/core/Select';
 import MenuItem from '@material-ui/core/MenuItem';
 import ContentType from '../../models/ContentType';
-import { Resource } from '../../models/Resource';
 import { useSelection } from '../../utils/hooks/useSelection';
 import { useActiveSiteId } from '../../utils/hooks/useActiveSiteId';
 import { useSelectorResource } from '../../utils/hooks/useSelectorResource';
 import { useDebouncedInput } from '../../utils/hooks/useDebouncedInput';
 import translations from './translations';
 import useStyles from './styles';
+import PreviewBrowseComponentsPanelUI from './PreviewBrowseComponentsPanelUI';
 
 interface ComponentResource {
   count: number;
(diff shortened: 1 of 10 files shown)


ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
```diff
@@ -179,7 +87,10 @@ export default function PreviewBrowseComponentsPanel() {
         const items =
           source.page[source.pageNumber]
             ?.map((id: string) => source.byId[id])
-            .filter((item: ContentInstance) => item.craftercms.contentTypeId === source.contentTypeFilter) || [];
+            .filter(
+              (item: ContentInstance) =>
+                source.contentTypeFilter === 'all' || item.craftercms.contentTypeId === source.contentTypeFilter
```
rart: - too inefficient to iterate through the whole array when you know in advanced that you needn't apply the filter at all
- also, the `||` is misplaced; it doesn't avoid an exception if the .map returns undefined
```js
let items = source.page[source.pageNumber]?.map(...) ?? []
if (source.contenTypeFilter !== 'all') 
  items = items.filter(...)

File: ui/app/src/state/epics/contentTypes.ts [hunk 2]
```diff
@@ -28,29 +28,30 @@ import { fetchItemsByContentType } from '../../services/content';
...
-        state.sites.active,
-        state.preview.components.contentTypeFilter,
-        state.contentTypes.byId,
-        state.preview.components.query
-      ).pipe(map(fetchComponentsByContentTypeComplete), catchAjaxError(fetchComponentsByContentTypeFailed))
-    )
-  );
-
-export default [fetch, fetchComponentsByContentType] as Epic[];
+] as Epic[];
```
rart:
```suggestion
] as CrafterCMSEpic[];
```

ui/app/src/state/actions/preview.ts
```diff
@@ -247,15 +247,9 @@ export const fetchAssetsPanelItemsComplete = /*#__PURE__*/ createAction<SearchRe
...
-  contentTypeFilter?: string,
-  options?: Partial<ComponentsContentTypeParams>
-): StandardAction {
-  return {
-    type: FETCH_COMPONENTS_BY_CONTENT_TYPE,
-    payload: { contentTypeFilter, options }
-  };
-}
+export const fetchComponentsByContentType = /*#__PURE__*/ createAction<Partial<ComponentsContentTypeParams>>(
+  FETCH_COMPONENTS_BY_CONTENT_TYPE
```
rart: we could/should discard these constants

ui/app/src/services/content.ts
```diff
@@ -462,31 +462,37 @@ export function fetchItemsByContentType(
...
   }
+
   return postJSON(`/studio/api/2/search/search.json?siteId=${site}`, {
     ...reversePluckProps(options, 'type'),
     filters: { 'content-type': contentTypes }
   }).pipe(
     map<AjaxResponse, { count: number; paths: string[] }>(({ response }) => ({
       count: response.result.total,
-      paths: response.result.items.filter((item) => item.type === options.type).map((item) => item.path)
+      paths: response.result.items.map((item) => item.path)
```
rart, line 472: Looks like you're dropping support for the type filter. why?
if we do it, then you need to drop it from the options of this method too.

ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
```diff
@@ -296,60 +201,46 @@ interface BrowsePanelUIProps {
 }
 
 function BrowsePanelUI(props: BrowsePanelUIProps) {
-  const { componentsResource, classes, onPageChanged, onDragStart, onDragEnd, chooseContentTypeImageUrl } = props;
+  const { componentsResource, classes, onPageChanged, onDragStart, onDragEnd } = props;
```
rart: Let's break the individual components into files

ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
```diff
@@ -296,60 +201,46 @@ interface BrowsePanelUIProps {
...
+            onDragStart={() => onDragStart(item)}
+            onDragEnd={onDragEnd}
+          />
+        ))}
+      </List>
+      {count === 0 && (
+        <EmptyState
+          title={formatMessage(translations.noResults)}
+          classes={{ image: classes.noResultsImage, title: classes.noResultsTitle }}
         />
```
rart: When there aren't any results, is there an empty message elsewhere? 
I'd also like to add a note, that only shared components are shown here. Let's discuss.

rart (changes requested):

ui/app/src/services/content.ts
aachavarria: I was doing some testing and the result from the fetch is only including components, so looks like the filter is not doing anything, i think in the past the fetch result included groovy or something similar.

ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
aachavarria: there is an  `*<EmptyState/>`, let's discuss tomorrow

rart (approved):

PR #2536, opened 2021-09-06
Update query selector to use ICEToolsPanel

jvega190:


rart (approved):
Please merge together with the BPs.

This will break ICE panels for existing 4 sites.
XML on existing sites needs to be updated from `craftercms.components.PageBuilderPanel` to `craftercms.components.ICEToolsPanel`.

sumerjabri (approved):

PR #2549, opened 2021-09-10
[4857] Rename RTE 5 to RTE

jvega190:
https://github.com/craftercms/craftercms/issues/4857


jvega190:
Blueprints need to be merged with this one to avoid breaking RTEs

rart (approved):

PR #2552, opened 2021-09-15
[4124][3.1.x] Bulk upload dialog becomes unresponsive with large volume of files

jvega190:
https://github.com/craftercms/craftercms/issues/4124

### DIFF
--- a/ui/app/src/components/BulkUpload.tsx
+++ b/ui/app/src/components/BulkUpload.tsx
@@ -72,6 +72,10 @@ const translations = defineMessages({
     id: 'bulkUpload.dropHere',
     defaultMessage: 'Drop files here or <span>browse</span>'
   },
+  maxFiles: {
+    id: 'bulkUpload.maxFiles',
+    defaultMessage: '{maxFiles} max.'
+  },
   browse: {
     id: 'words.browse',
     defaultMessage: 'Browse'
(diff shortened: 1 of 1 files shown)


ui/app/src/components/BulkUpload.tsx
```diff
@@ -88,6 +95,10 @@ const translations = defineMessages({
     id: 'bulkUpload.uploadInProgress',
     defaultMessage:
       'Uploads are still in progress. Leaving this page would stop the pending uploads. Are you sure you wish to leave?'
+  },
+  maxActiveUploadsReached: {
+    id: 'bulkUpload.maxActiveUploadsReached',
+    defaultMessage: 'Max active uploads reached.'
```
rart:
```suggestion
    defaultMessage: '{maxFiles} maximum active uploads reached. Excess has been discarded.'
```

rart: Also update 4

ui/app/src/components/BulkUpload.tsx
```diff
@@ -517,6 +548,11 @@ const DropZone = React.forwardRef((props: DropZoneProps, ref: any) => {
                   </span>
                 )
               })}
+              &nbsp; (
```
rart: Double check this. Pretty sure this will show the text.
```
              &nbsp; (
```

rart (changes requested):

ui/app/src/components/BulkUpload.tsx
jvega190: It works, but I changed it to be a span with styles for the spacing

ui/app/src/components/BulkUpload.tsx
jvega190: Updated, the maxFiles parameter in the translation was already being sent (forgot to remove before) so it doesn't need to be added (in 3.1)

rart (approved):

sumerjabri:
@russdanner do we want this in 3.1.18 or hold off to reduce risk?

PR #2558, opened 2021-09-21
[4854] - Add RTE config to page builder - Do not show full rte-text-editing in plain text controls

jvega190:
https://github.com/craftercms/craftercms/issues/4854

### DIFF
--- a/ui/guest/src/store/epics/root.ts
+++ b/ui/guest/src/store/epics/root.ts
@@ -342,7 +342,8 @@ const epic: Epic<GuestStandardAction, GuestStandardAction, GuestState> = combine
               } else if (not(validations?.readOnly?.value)) {
                 const setupId = field.properties?.rteConfiguration?.value ?? 'generic';
                 const setup = state.rteConfig[setupId] ?? Object.values(state.rteConfig)[0] ?? {};
-                return initTinyMCE(record, validations, setup);
+                // Only pass rte setup to html type, text/textarea (plaintext) controls won't show full rich-text-editing.
+                return initTinyMCE(record, validations, type === 'html' ? setup : {});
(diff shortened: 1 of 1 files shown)


ui/guest/src/store/epics/root.ts
```diff
@@ -342,7 +342,8 @@ const epic: Epic<GuestStandardAction, GuestStandardAction, GuestState> = combine
               } else if (not(validations?.readOnly?.value)) {
                 const setupId = field.properties?.rteConfiguration?.value ?? 'generic';
                 const setup = state.rteConfig[setupId] ?? Object.values(state.rteConfig)[0] ?? {};
-                return initTinyMCE(record, validations, setup);
+                // Only pass rte setup to html type, text/textarea (plaintext) controls won't show full rich-text-editing.
+                return initTinyMCE(record, validations, type === 'html' ? setup : {});
```
rart, line 346: This is not good enough. The point is that it's plain text. We need a toolbar that reflects that (or no toolbar at all).
With the default setup there's still stuff that doesn't apply:
image

rart (changes requested):

ui/guest/src/store/epics/root.ts
jvega190: Per discussion, I tested on a clean installation and it works properly. The issue where the label of the 'Hero Title' field is not showing `(Rich Text)` is due to a `@label` attribute set on the field (in the bp). I'll remove it from the repo.

jvega190: done in here https://github.com/craftercms/studio/pull/2747

rart (approved):

PR #2562, opened 2021-09-22
[5009] Text limits are not enforced on the RTE-2 Text Fields

jvega190:
https://github.com/craftercms/craftercms/issues/5009


rart (approved):
@jvega190 please send PR for 4

jvega190:
@rart we removed the old RTE control, and the new control doesn't have the text limits configuration.

rart:
> we removed the old RTE control, and the new control doesn't have the text limits configuration.

@russdanner please see above comment👆Is that a miss? Should we add it to the RTE.

@sumerjabri FYI

PR #2563, opened 2021-09-22
[4810][3.1.x] Top toolbar may not update on time when navigating via preview

jvega190:
- Disable nav on guest check out.

https://github.com/craftercms/craftercms/issues/4810

### DIFF
--- a/static-assets/scripts/guest.js
+++ b/static-assets/scripts/guest.js
@@ -232,6 +232,12 @@ crafterDefine('guest', ['crafter', 'jquery', 'communicator', 'ice-overlay'], fun
       clearSetTimeout(Constants.TIME_SCROLL);
     });
 
+    $(window).on('beforeunload', function () {
+      communicator.publish(Topics.GUEST_CHECKOUT, {
+        location: window.location.href
+      });
+    });
+
     loadCss(
       config.hostOrigin + crafter.join('/', config.studioContext, config.studioStaticAssets, 'styles', 'guest.css')
     );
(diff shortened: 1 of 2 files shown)


static-assets/scripts/guest.js
```diff
@@ -232,6 +232,12 @@ crafterDefine('guest', ['crafter', 'jquery', 'communicator', 'ice-overlay'], fun
       clearSetTimeout(Constants.TIME_SCROLL);
     });
 
+    $(window).on('beforeunload', function () {
+      communicator.publish(Topics.GUEST_CHECKOUT, {
+        location: window.location.href
+      });
```
rart:
```suggestion
      communicator.publish(Topics.GUEST_CHECKOUT);
```

static-assets/scripts/host.js
```diff
@@ -581,7 +581,6 @@
 
     path = path.replace('//', '/');
 
-    CStudioAuthoring.ContextualNav.WcmActiveContent?.disableNav(true);
```
rart, line 584: I think these may be covering other scenarios and we should leave them as long as they cause no trouble.

static-assets/scripts/host.js
```diff
@@ -634,7 +633,6 @@
   }
 
   function selectStudioContent(site, url) {
-    CStudioAuthoring.ContextualNav.WcmActiveContent?.disableNav(true);
```
rart, line 637: I think these may be covering other scenarios and we should leave them as long as they cause no trouble.

rart (changes requested):

rart (approved):

PR #2565, opened 2021-09-23
[5004] Invalid timezone causes Studio UI to crash

jvega190:
https://github.com/craftercms/craftercms/issues/5004

### DIFF
--- a/ui/app/src/state/epics/configuration.ts
+++ b/ui/app/src/state/epics/configuration.ts
@@ -15,7 +15,7 @@
  */
 
 import { ofType } from 'redux-observable';
-import { exhaustMap, filter, map, tap, withLatestFrom } from 'rxjs/operators';
+import { exhaustMap, filter, map, switchMap, withLatestFrom } from 'rxjs/operators';
 import { catchAjaxError } from '../../utils/ajax';
 import {
   fetchSiteConfig,
@@ -31,7 +31,6 @@ import {
 } from '../../services/configuration';
 import { CrafterCMSEpic } from '../store';
 import { showSystemNotification } from '../actions/system';
-import { getHostToHostBus } from '../../modules/Preview/previewContext';
 import { defineMessages } from 'react-intl';
 
 const configurationMessages = defineMessages({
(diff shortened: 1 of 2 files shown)


ui/app/src/state/epics/configuration.ts
```diff
@@ -46,13 +56,28 @@ export default [
...
+              const hostToHost$ = getHostToHostBus();
+              hostToHost$.next(
+                showSystemNotification({
+                  // @ts-ignore
+                  message: getIntl().formatMessage(configurationMessages.localeError, { message: e.message })
+                })
+              );
+            }
+          }),
           map((config) => fetchSiteConfigComplete(config)),
```
rart: Why ts-ignore? Should need to be one. **ALWAYS** explain any ts/eslint ignores you add.
```suggestion
          switchMap((config) => {
            try {
              let localeCode = config.locale?.localeCode || state.uiConfig.locale.localeCode;
              let options = config.locale?.dateTimeFormatOptions || state.uiConfig.locale.dateTimeFormatOptions;
              new Intl.DateTimeFormat(localeCode, options);
              return [fetchSiteConfigComplete(config)];
            } catch (e) {
                return [
                  fetchSiteConfigComplete({ ...config, /* use modified fallback values here and remove try from reducer */ }), 
                  showSystemNotification({
                    message: getIntl().formatMessage(configurationMessages.localeError, { message: e.message })
                  })
                ];
              );
            }
          }),
```

ui/app/src/state/reducers/uiConfig.ts
```diff
@@ -148,14 +148,25 @@ const reducer = createReducer<GlobalState['uiConfig']>(initialState, {
...
+    let localeCode = locale?.localeCode || initialState.locale.localeCode;
+    let options = locale?.dateTimeFormatOptions || initialState.locale.dateTimeFormatOptions;
+
+    // check for locale and options validity
+    try {
+      new Intl.DateTimeFormat(localeCode, options);
+    } catch (e) {
+      localeCode = initialState.locale.localeCode;
+      options = initialState.locale.dateTimeFormatOptions;
+    }
```
rart: I think we can now roll this back as the epic cleanses

rart (changes requested):

ui/app/src/state/epics/configuration.ts
```diff
@@ -46,14 +55,37 @@ export default [
...
+                  locale: {
+                    ...config.locale,
+                    localeCode: state.uiConfig.locale.localeCode,
+                    dateTimeFormatOptions: state.uiConfig.locale.dateTimeFormatOptions
+                  }
+                }),
+                showSystemNotification({
+                  // @ts-ignore - 'message' does not exist on type 'unknown', and catch variable type annotation must
+                  // be 'any' or 'unknown', but setting it to 'any' causes a prettier error.
+                  message: getIntl().formatMessage(configurationMessages.localeError, { message: e.message })
```
rart:
```suggestion
                  message: getIntl().formatMessage(configurationMessages.localeError, { message: (e as Error).message })
```
Oddly, it works even without the casting for me. I didn't see the `'message' does not exist on type 'unknown'` TS error via the IDE or the compilation even removing the TS ignore. Do you get the error when compiling?

rart (changes requested):

ui/app/src/state/epics/configuration.ts
jvega190: I just realized that the error is being displayed only on the IDE, not when compiling. Should I leave it with the casting?

rart (approved):

PR #2568, opened 2021-09-27
[5010] Content API's return "NULL" string instead of null object for some properties of folder content.

jvega190:
- When mimeType is null, do not display the - {mime} piece of the tooltip.

https://github.com/craftercms/craftercms/issues/5010

### DIFF
--- a/ui/app/src/components/ItemTypeIcon/ItemTypeIcon.tsx
+++ b/ui/app/src/components/ItemTypeIcon/ItemTypeIcon.tsx
@@ -44,7 +44,7 @@ export interface ItemTypeIconProps {
 }
 
 export function getItemTypeText(item: DetailedItem | SandboxItem) {
-  return `${capitalize(item.systemType ?? 'Unknown')} - ${item.mimeType}`;
+  return `${capitalize(item.systemType ?? 'Unknown')} ${item.mimeType ? `- ${item.mimeType}` : ''}`;
 }
 
 export default function ItemTypeIcon(props: ItemTypeIconProps) {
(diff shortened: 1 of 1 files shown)


ui/app/src/components/ItemTypeIcon/ItemTypeIcon.tsx
```diff
@@ -44,7 +44,7 @@ export interface ItemTypeIconProps {
 }
 
 export function getItemTypeText(item: DetailedItem | SandboxItem) {
-  return `${capitalize(item.systemType ?? 'Unknown')} - ${item.mimeType}`;
+  return `${capitalize(item.systemType ?? 'Unknown')} ${item.mimeType ? `- ${item.mimeType}` : ''}`;
```
rart: This was more of a debugging tool that something we wanted to actually show. I think the value of it still, is the fallback so let's do:
```suggestion
  return messages[item.systemType] ? formatMessage(messages[item.systemType]) : item.mimeType ? item.mimeType : formatMessage(messages.unknown);
```

rart (changes requested):

ui/app/src/components/ItemTypeIcon/ItemTypeIcon.tsx
```diff
@@ -44,7 +44,7 @@ export interface ItemTypeIconProps {
 }
 
 export function getItemTypeText(item: DetailedItem | SandboxItem) {
```
rart:
```suggestion
export function getItemTypeText(item: DetailedItem | SandboxItem, formatMessage: IntlFormatters['formatMessage']) {
```

rart (changes requested):

rart (approved):

PR #2576, opened 2021-10-04
[5016] File Browse datasource repository path only allows paths under /static-assets

aachavarria:
https://github.com/craftercms/craftercms/issues/5016

### DIFF
--- a/static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
+++ b/static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
@@ -119,10 +119,10 @@ YAHOO.extend(CStudioForms.Datasources.FileBrowseRepo, CStudioForms.CStudioFormDa
         label: CMgs.format(langBundle, 'repositoryPath'),
         name: 'repoPath',
         type: 'content-path-input',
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/data-sources/file-browse-repo.js
```diff
@@ -119,10 +119,10 @@ YAHOO.extend(CStudioForms.Datasources.FileBrowseRepo, CStudioForms.CStudioFormDa
         label: CMgs.format(langBundle, 'repositoryPath'),
         name: 'repoPath',
         type: 'content-path-input',
-        defaultValue: '/static-assets/',
-        rootPath: '/static-assets',
+        defaultValue: '/site/',
+        rootPath: '/site',
         validations: {
-          regex: /^\/static-assets(\/.*)?$/
+          regex: /^\/site(\/.*)?$/
```
rart: Doing this, then we wouldn't be able to select static assets. Looks like we have to leave with a blank root path.

rart (changes requested):

rart (approved):

PR #2584, opened 2021-10-06
[5024] Deleted items "My Last Edit" date in Dashboard dashlet displays epoch time

aachavarria:
https://github.com/craftercms/craftercms/issues/5024

### DIFF
--- a/ui/app/src/components/RecentActivityDashlet/RecentActivityDashlet.tsx
+++ b/ui/app/src/components/RecentActivityDashlet/RecentActivityDashlet.tsx
@@ -91,24 +91,29 @@ export default function RecentActivityDashlet() {
   const { itemsByPath, isFetching } = useDetailedItems(Object.keys(selectedLookup));
 
   const isAllChecked = useMemo(
-    () => !items.some((item) => !item.stateMap.deleted && !selectedLookup[item.path]),
+    () =>
+      items.length > 1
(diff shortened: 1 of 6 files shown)


ui/app/src/components/RecentActivityDashlet/RecentActivityDashlet.tsx
```diff
@@ -91,24 +91,29 @@ export default function RecentActivityDashlet() {
   const { itemsByPath, isFetching } = useDetailedItems(Object.keys(selectedLookup));
 
   const isAllChecked = useMemo(
-    () => !items.some((item) => !item.stateMap.deleted && !selectedLookup[item.path]),
+    () =>
+      items.length > 1
+        ? !items.some((item) => !item.stateMap.deleted && !selectedLookup[item.path])
+        : items[0]
+        ? selectedLookup[items[0].path]
+        : false,
```
rart, line 99: What if the single item that's present is also deleted? Out of the `items.some(...)`, there's no validation of `.deleted`. How come?

rart (changes requested):

ui/app/src/components/RecentActivityDashlet/RecentActivityDashlet.tsx
aachavarria: if the single item is deleted, it can't be on the selectedLookup, because the checkbox is disabled

rart (approved):

PR #2599, opened 2021-10-15
[studio-ui] Guest page builder move mode

aachavarria:


ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -15,29 +15,76 @@
...
+  }, [index]);
+
+  const dropzoneRecord = useMemo(
+    () => fromElement(getParentElementFromICEProps(modelId, fieldId, index)[0] as Element),
+    [fieldId, index, modelId]
+  );
+
+  const dropzoneChildrenLength = useMemo(
+    () => compileDropZone(dropzoneRecord.iceIds[0]).children.length,
+    [dropzoneRecord.iceIds]
```
rart:
```suggestion
    [dropzoneRecord.iceIds[0]]
```

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -56,16 +104,20 @@ export function MoveModeZoneMenu(props: MoveModeZoneMenuProps) {
...
-        </UltraStyledIconButton>
-      </Tooltip>
+      {index !== 0 && (
+        <Tooltip title="Move up/left (← or ↑)">
+          <UltraStyledIconButton size="small" onClick={onMoveUp}>
+            <ArrowUpwardRoundedIcon />
+          </UltraStyledIconButton>
+        </Tooltip>
+      )}
+      {index < dropzoneChildrenLength - 1 && (
```
rart: For clarity, please extract these conditions to vars above (`isFirstItem`, `isLastItem`). We should also have `isOnlyItem` and hide both buttons if the item is there by itself.

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -15,29 +15,76 @@
...
+    return index;
+  }, [index]);
+
+  const dropzoneRecord = useMemo(
+    () => fromElement(getParentElementFromICEProps(modelId, fieldId, index)[0] as Element),
+    [fieldId, index, modelId]
+  );
+
+  const dropzoneChildrenLength = useMemo(
+    () => compileDropZone(dropzoneRecord.iceIds[0]).children.length,
```
rart: If all you're getting from this is the number of children, it'd be best to not use the DOM and only use the ice registry to get this information.

ui/app/src/modules/Preview/PreviewConcierge.tsx
```diff
@@ -509,7 +509,7 @@ export function PreviewConcierge(props: any) {
             () => {
               issueDescriptorRequest({
                 site: siteId,
-                path,
+                path: path ?? models[parentModelId].craftercms.path,
```
rart, line 506: In **each** case, looks like there are several expressions to get the path. Can we consolidate?

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -48,23 +108,58 @@ export function MoveModeZoneMenu(props: MoveModeZoneMenuProps) {
...
+          onTrash();
+          break;
+        }
+      }
+    };
+    document.addEventListener('keydown', onKeyDown);
+    return () => {
+      document.removeEventListener('keydown', onKeyDown);
+    };
+  }, [isFirstItem, isLastItem, onMoveDown, onMoveUp, onTrash]);
```
rart: `onMoveDown`, `onMoveUp`, `onTrash` will trigger this effect every render.

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -48,23 +108,58 @@ export function MoveModeZoneMenu(props: MoveModeZoneMenuProps) {
     });
   };
   // endregion
+
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      switch (e.key) {
+        case 'ArrowUp': {
```
rart, line 123: Add left arrow to this case

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -48,23 +108,58 @@ export function MoveModeZoneMenu(props: MoveModeZoneMenuProps) {
...
+          if (!isFirstItem) {
+            e.preventDefault();
+            onMoveUp();
+          }
+          break;
+        }
+        case 'ArrowDown': {
```
rart, line 130: add right arrow to this case

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -48,23 +108,58 @@ export function MoveModeZoneMenu(props: MoveModeZoneMenuProps) {
...
+          if (!isLastItem) {
+            e.preventDefault();
+            onMoveDown();
+          }
+          break;
+        }
+        case 'Backspace': {
```
rart, line 136: Add confirmation to this case

ui/guest/src/store/reducers/root.ts
```diff
@@ -130,12 +130,13 @@ const reducer = createReducer(initialState, {
...
         const movableRecordId = iceRegistry.getMovableParentRecord(iceId);
-        if (notNullOrUndefined(movableRecordId)) {
+        const movableRecord = fromICEId(movableRecordId);
+        if (notNullOrUndefined(movableRecord)) {
           const draggable = getDraggable(movableRecordId);
           const highlight = getHoverData(
             // If (iceId == movableRecordId) the current record is already
             // the one to show the highlight on.
-            iceId === movableRecordId ? record.id : fromICEId(movableRecordId).id
+            iceId === movableRecordId ? record.id : movableRecord.id
```
rart: Please explain these changes. What's wrong with checking movableRecordId without pulling the element record yet?

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -15,29 +15,89 @@
...
+    fieldId: [fieldId],
+    index
+  } = record;
+
+  const elementIndex = useMemo(() => {
+    if (typeof index === 'string') {
+      return parseInt(index.substr(index.lastIndexOf('.') + 1), 10);
+    }
+    return index;
+  }, [index]);
```
rart:
```suggestion
  const elementIndex = useMemo(() => typeof index === 'string' ? parseInt(popPiece(index), 10) : index, [index]);
```

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -15,29 +15,89 @@
...
+
+  const dropzoneRecord = useMemo(
+    () => fromElement(getParentElementFromICEProps(modelId, fieldId, index)[0] as Element),
+    [fieldId, index, modelId]
+  );
```
rart: Likely not required

ui/guest/src/react/MoveModeZoneMenu.tsx
rart: Try
```suggestion
    () => extractCollection(getCachedModel(modelId), fieldId, index).length,
```

ui/guest/src/react/MoveModeZoneMenu.tsx
```diff
@@ -15,29 +15,89 @@
...
+
+  const onMoveDown = () => {
+    contentController.sortItem(
+      dropzoneRecord.modelId,
+      dropzoneRecord.fieldId[0],
+      dropzoneRecord.fieldId[0].includes('.') ? `${dropzoneRecord.index}.${elementIndex}` : elementIndex,
+      dropzoneRecord.fieldId[0].includes('.') ? `${dropzoneRecord.index}.${elementIndex + 1}` : elementIndex + 1
+    );
+    clearAndStartListening();
+  };
```
rart: I think we should abstract this on the content controller: create `moveItemUp` & `moveItemDown` methods there and just invoke here. Method should take into account only items, index 0, last index and ignore updates in those cases.

ui/guest/src/store/reducers/root.ts
aachavarria: we can revert this change, i was trying to fix the issue when the items are moved and the registration is wrong

ui/app/src/modules/Preview/PreviewConcierge.tsx
aachavarria: almost all use the same, but there are a couple of exception

PR #2601, opened 2021-10-18
[4856] Add paste from MS Word to RTE 5

jvega190:
https://github.com/craftercms/craftercms/issues/4856


rart (changes requested):
- Add `ul`
- Remove jQuery

rart (approved):

PR #2617, opened 2021-10-26
Update guest to load shared utils from app

jvega190:


ui/app/src/utils/contentType.ts
```diff
@@ -65,3 +94,39 @@ export function getFields(type: ContentType, ...ids: string[]): ContentTypeField
...
+        return field.validations.minValue?.value ?? 1;
+      }
+      case 'boolean':
+        return 'false';
+      case 'date-time':
+        return new Date().toISOString();
+    }
+  } else {
+    return null;
+  }
```
rart:
```suggestion
  } else {
    switch (field.type) {
      case 'image':
        const width = field.validations.width?.value ?? field.validations.minWidth?.value ?? 150;
        const height = field.validations.height?.value ?? field.validations.minHeight?.value ?? width;
        return `https://via.placeholder.com/${width}x${height}`;
      case 'textarea':
      case 'text': {
        let maxLength = parseInt(field.validations.maxLength?.value);
        let lorem = new LoremIpsum({
          wordsPerSentence: {
            max: 4,
            min: 4
          }
        });
        return maxLength ? lorem.generateSentences(1).substring(0, maxLength / 2) : lorem.generateWords(1);
      }
      case 'html':
        let lorem = new LoremIpsum();
        return lorem.generateParagraphs(1);
      case 'numeric-input': {
        return field.validations.minValue?.value ?? 1;
      }
      case 'boolean':
        return 'false';
      case 'date-time':
        return new Date().toISOString();
      default:
        return null;
    }
  }
```

rart (approved):

PR #2622, opened 2021-10-28
[5051][3.1.x] Switch to write-content API and stop using asset-upload

jvega190:
https://github.com/craftercms/craftercms/issues/5051


rart (commented):
Why are there less files removed here than in the analogue pr?

jvega190:
The ticket mentions that in 3.1.x asset upload will be deprecated, and removed in 4

rart (approved):

PR #2623, opened 2021-10-28
[5051][next] Switch to write-content API and stop using asset-upload

jvega190:
https://github.com/craftercms/craftercms/issues/5051


rart (approved):
@jvega190  please double check that the deleted dialog is not used anywhere in legacy

PR #2630, opened 2021-10-29
[4717] Don't close form dialogs with data entry when the user clicks out of a dialog

aachavarria:
https://github.com/craftercms/craftercms/issues/4717


ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
```diff
@@ -153,7 +157,10 @@ export default function GlobalConfigManagement() {
   const onConfirmOk = () => {
     setHasChanges(false);
     setShowConfirmDialog(false);
-    history.push(nextRoute);
+    // timeout need to avoid running the useEffect on linea:64 with hasChanges on true
```
rart:
```suggestion
    // timeout needed to avoid running the useEffect on line:64 with hasChanges on true
```

rart (approved):

PR #2634, opened 2021-11-02
[5061][5062] - Shared content DS issues

jvega190:
When search is enabled for a DS and on the browse path the parentPath macro is used, an exception is thrown #5061 - https://github.com/craftercms/craftercms/issues/5061
Studio opens both DS when they're setup for search #5062 - https://github.com/craftercms/craftercms/issues/5062


rart (approved):
@jvega190 PR to Develop

PR #2636, opened 2021-11-03
[5061][5062][next] - Shared content DS issues

jvega190:
When search is enabled for a DS and on the browse path the parentPath macro is used, an exception is thrown #5061 - craftercms/craftercms#5061
Studio opens both DS when they're setup for search #5062 - craftercms/craftercms#5062

### DIFF
--- a/ui/legacy/src/components/cstudio-forms/data-sources/dropTargets.js
+++ b/ui/legacy/src/components/cstudio-forms/data-sources/dropTargets.js
@@ -150,8 +150,14 @@
       return [];
     },
 
+    _processPathsForMacros: function (path) {
+      const model = this.form.model;
+      return CStudioAuthoring.Operations.processPathsForMacros(path, model);
+    },
+
     _openBrowse: function (contentType, control) {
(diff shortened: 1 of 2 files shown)


ui/legacy/src/components/cstudio-forms/data-sources/dropTargets.js
```diff
@@ -150,8 +150,14 @@
...
     },
 
+    _processPathsForMacros: function (path) {
+      const model = this.form.model;
+      return CStudioAuthoring.Operations.processPathsForMacros(path, model);
+    },
+
     _openBrowse: function (contentType, control) {
-      const path = `${this.baseBrowsePath}/${contentType.replace(/\//g, '_').substr(1)}`;
+      let path = this._processPathsForMacros(this.baseBrowsePath);
```
jvega190, line 159: @rart Need to discuss a couple of things here

rart:
Is this ready now?

jvega190:
@rart yes it is, I resolved the comment that I added before

rart (approved):

PR #2641, opened 2021-11-04
[5065][next] Site Dashboard stuck in processing when a scheduled publish is published in GMT Timezone

jvega190:
- Updated api calls that were not matching APIs,
- Fixed no filterBy validation

https://github.com/craftercms/craftercms/issues/5065

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -3705,18 +3705,18 @@ var nodeOpen = false,
       getScheduledItems: function (site, sortBy, sortAscDesc, filterBy, callback) {
         callback.beforeServiceCall();
 
-        if (filterBy === undefined && filterBy === null) {
+        if (filterBy === undefined || filterBy === null) {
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -3705,39 +3705,37 @@ var nodeOpen = false,
       getScheduledItems: function (site, sortBy, sortAscDesc, filterBy, callback) {
         callback.beforeServiceCall();
 
-        if (filterBy === undefined && filterBy === null) {
+        if (filterBy === undefined || filterBy === null) {
```
rart, line 3708: Where is this change coming from? Don't remember the 3.1.x PR having it.

jvega190: This is the updated API call using CrafterCMSNext (that's why it is different than 3.1.x), and there was an error validating filterBy

rart (approved):

PR #2647, opened 2021-11-09
[5082][3.1.x] Studio search breaks when updating the search term while in a > 1 page

aachavarria:
https://github.com/craftercms/craftercms/issues/5082


rart (commented):
@aachavarria is `currentPage` a 0-based index? Shouldn't these added statements reset to 0?

aachavarria:
@rart no, the base page for search on 3.1 is 1

rart (approved):

PR #2648, opened 2021-11-09
[5083][3.1.x] When making an edit using ICE, the timezone field of a date field is removed

jvega190:
https://github.com/craftercms/craftercms/issues/5083

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -551,13 +551,7 @@ var CStudioForms =
       this.id = name;
       this.style = style;
       this.definition = formDefinition;
-      // Initialize dynamicFields with all current keys in model.
-      // This is because if we're opening a form with only some fields and there were already values in the model
-      // set by registering dynamic fields, then the controls are not being initialized (so dynamic fields wont be
-      // registered), and when saving the content id won't save the dynamic fields.
-      // This way we can assure that all original fields (when opening the form) will be saved and no other fields will
-      // unless registering them via `registerDynamicField`.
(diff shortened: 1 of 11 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -551,7 +551,13 @@ var CStudioForms =
...
       this.style = style;
       this.definition = formDefinition;
-      this.dynamicFields = [];
+      // Initialize dynamicFields with all current keys in model.
+      // This is because if we're opening a form with only some fields and there were already values in the model
+      // set by registering dynamic fields, then the controls are not being initialized (so dynamic fields wont be
+      // registered), and when saving the content id won't save the dynamic fields.
+      // This way we can assure that all original fields (when opening the form) will be saved and no other fields will
+      // unless registering them via `registerDynamicField`.
+      this.dynamicFields = Object.keys(model) ?? [];
```
rart: is model the content itself or the content type definition?
What if the content itself has stale values that the content type no longer has? Will they end up getting carried over?

jvega190: The detail is that the dynamic fields are not in the content type definition because they're dynamically created on controls, for example in date-time control, the `{variable}_dt` is the one defined in the content-type definition, and an extra variable `{variable}_dt_tz` is dynamically created in the control, but it doesn't exist in the definition.

rart: Aaron will rollback changes made/merged in `develop`. We need to find another solution. We've found the first bug already (which should have gotten caught in testing).

PR #2661, opened 2021-11-10
[3752] Content Type: Insert expression for Destination Path Pattern and Postfix are broken

jvega190:
https://github.com/craftercms/craftercms/issues/3752

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-types.js
+++ b/static-assets/components/cstudio-admin/mods/content-types.js
@@ -3742,11 +3742,7 @@
 
     CStudioAdminConsole.helpInsert = function (button, identifier) {
       var $button = $(button);
-      // Some prop-types wrap the input within a div
-      const $input =
-        $(identifier).siblings('div').length > 0
-          ? $(identifier).siblings('div').find('input')
-          : $(identifier).siblings('input');
+      const $input = $(identifier).parent().find('input');
       $input.val($input.val() + $button.attr('data-insert'));
 
       $input.change();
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -3742,7 +3742,11 @@
 
     CStudioAdminConsole.helpInsert = function (button, identifier) {
       var $button = $(button);
-      var $input = $(identifier).siblings('input');
+      // Some prop-types wrap the input within a div
+      const $input =
+        $(identifier).siblings('div').length > 0
+          ? $(identifier).siblings('div').find('input')
+          : $(identifier).siblings('input');
```
rart: Too wired. Go to the parent of the row and find the child there. Preferably through some specific class or attribute if available.

rart (changes requested):

static-assets/components/cstudio-admin/mods/content-types.js
jvega190: Right now theres no class to find directly the input (there are different types of controls with different classes). By using `identifier` (is a selector), getting the parent and finding the input it does work. Should we add a common class to the controls (all that are inputs and could use the `helpInsert`) for a more specific selection?

jvega190: Tested with current usages (postfixes and macros) and it is working

rart (approved):

PR #2665, opened 2021-11-12
Add transition none style to Popper to avoid unwanted possible transitions set by sites #5098

jvega190:
https://github.com/craftercms/craftercms/issues/5098


rart:
Please explain

jvega190:
@rart if there's a transition style set to a `role="tooltip"` element, it will apply to our `ZoneMarkers`, which I don't think we want. For example look at https://www.loom.com/share/5dfcb828862448ebb5305fee56f6b244

ui/guest/src/react/ZoneMarker.tsx
```diff
@@ -89,6 +89,9 @@ export function ZoneMarker(props: ZoneMarkerProps) {
           anchorEl={elRef.current}
           placement="top-start"
           onClick={onPopperClick}
+          style={{
+            transition: 'none'
+          }}
```
rart: Add it to the sx and type it appropriately.

rart (changes requested):

rart (approved):
@jvega190 popper does have a transition prop. does that not serve the purpose?

rart (approved):

PR #2671, opened 2021-11-16
[5078] UI display issues

jvega190:
https://github.com/craftercms/craftercms/issues/5078


jvega190:
Tested in other controls like `node-selector` and `video-picker` and the issue is not happening there

rart (approved):

PR #2672, opened 2021-11-16
[5072] Remove error graphic from PathNavigator path not found case

jvega190:
https://github.com/craftercms/craftercms/issues/5072


rart (approved):
@jvega190 draft

rart (approved):

PR #2676, opened 2021-11-17
[5095] Using the RTE code editor in in-context-editing fails

jvega190:
https://github.com/craftercms/craftercms/issues/5095


jvega190:
@rart I added a `code_editor_inline` property to disable it from the preview rte control, so it will display the dialog there, and the inline editor (+ dialog switch) in forms-engine

rart (approved):

PR #2688, opened 2021-11-22
[5074] Add "maximize" to code editor dialog

jvega190:
https://github.com/craftercms/craftercms/issues/5074

### DIFF
--- a/ui/app/src/components/CodeEditorDialog/CodeEditorDialogContainer.tsx
+++ b/ui/app/src/components/CodeEditorDialog/CodeEditorDialogContainer.tsx
@@ -50,7 +50,7 @@ import { CodeEditorDialogContainerProps, getContentModelSnippets } from './utils
 import { batchActions } from '../../state/actions/misc';
 
 export function CodeEditorDialogContainer(props: CodeEditorDialogContainerProps) {
-  const { path, onMinimize, onClose, onSaveClose, mode, isSubmitting, readonly, contentType } = props;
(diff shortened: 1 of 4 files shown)


ui/app/src/components/CodeEditorDialog/CodeEditorDialogContainer.tsx
```diff
@@ -209,6 +209,10 @@ export function CodeEditorDialogContainer(props: CodeEditorDialogContainerProps)
             {
               icon: 'MinimizeIcon',
               onClick: onMinimize
+            },
+            {
+              icon: 'MaximizeIcon',
+              onClick: toggleMaximized
             }
```
rart: This whole block is conditional to only onMinimize.

ui/app/src/components/CodeEditorDialog/utils.ts
```diff
@@ -44,6 +44,7 @@ export interface CodeEditorDialogContainerProps
   extends CodeEditorDialogBaseProps,
     Pick<CodeEditorDialogProps, 'onMinimize' | 'isSubmitting' | 'onSuccess' | 'onClose'> {
   title: string;
+  toggleMaximized(): void;
```
rart:
```suggestion
  onFullScreen(): void;
```

rart (changes requested):

ui/app/src/components/CodeEditorDialog/utils.ts
rart: 👆

rart (changes requested):
👆 see above

rart (approved):

PR #2692, opened 2021-11-24
[5081] When an item is selected in move mode and the user switches to edit mode, the item still behaves as movable

aachavarria:
https://github.com/craftercms/craftercms/issues/5081

### DIFF
--- a/ui/guest/src/store/reducers/root.ts
+++ b/ui/guest/src/store/reducers/root.ts
@@ -409,7 +409,6 @@ const reducer = createReducer(initialState, {
     state.highlightMode !== payload.highlightMode
       ? {
           ...resetState(state),
-          highlighted: {},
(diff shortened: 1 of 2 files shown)


ui/guest/src/store/reducers/root.ts
```diff
@@ -408,7 +408,7 @@ const reducer = createReducer(initialState, {
   [highlightModeChanged.type]: (state, { payload }) =>
     state.highlightMode !== payload.highlightMode
       ? {
-          ...state,
+          ...resetState(state),
           highlighted: {},
```
rart: resetState already resets highlighted
```suggestion
```

rart (changes requested):

rart (approved):

PR #2704, opened 2021-12-02
[5095] using the RTE code editor in in-context-editing fails

jvega190:
https://github.com/craftercms/craftercms/issues/5095


static-assets/js/tinymce-plugins/ace/plugin.js
```diff
@@ -184,7 +184,12 @@ tinymce.PluginManager.add(PLUGIN_NAME, function (editor, url) {
     aceEditor.on('change', function () {
       editorTextareaEl.value = aceEditor.getValue();
       editor.setContent(aceEditor.getValue());
-      aceModes.inline.getSession().setValue(aceEditor.getValue());
+
+      if (inlineMode) {
+        aceModes.inline.getSession().setValue(aceEditor.getValue());
```
rart, line 189: Would this path not require the `external_change` to be fired too?

jvega190: Inline mode does not need the event, since that's used on the form, and when saving it retrieves the value and save. For dialog mode it needs to be triggered so the preview control know that there's a change for it to save when focusing out. Though we may leave that event for anyone outside that needs to know about an external change.
Will make the change.

static-assets/js/tinymce-plugins/ace/plugin.min.js
```diff
@@ -1 +1,117 @@
...
\ No newline at end of file
+const PLUGIN_NAME = 'acecode',
```
rart: Not minified

rart (changes requested):

rart (approved):

PR #2705, opened 2021-12-02
[5127][next] Cannot save form with a required RTE input with only image

jvega190:
https://github.com/craftercms/craftercms/issues/5127

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -5696,8 +5696,11 @@ var nodeOpen = false,
         const textarea = document.createElement('textarea');
         const div = document.createElement('div');
         textarea.innerHTML = html;
-        div.innerHTML = textarea.value;
-        return div.innerText.trim() === '';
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -5696,8 +5696,11 @@ var nodeOpen = false,
         const textarea = document.createElement('textarea');
         const div = document.createElement('div');
         textarea.innerHTML = html;
-        div.innerHTML = textarea.value;
-        return div.innerText.trim() === '';
```
rart, line 5700: Just curious, how come this doesn’t work?

rart (approved):

static-assets/components/cstudio-common/common-api.js
jvega190: This validation was just retrieving the text of the element, not markup like in this case `img`, so if we had a content like
```<p>&nbsp;</p>```
innerText would have returned an empty string.

PR #2717, opened 2021-12-14
[5139] Dragging an image from the page onto the page in XB on Chrome on Linux causes problems for XB

aachavarria:
https://github.com/craftercms/craftercms/issues/5139


ui/guest/src/react/Guest.tsx
```diff
@@ -420,7 +420,9 @@ function Guest(props: GuestProps) {
         .subscribe((e) => {
           e.preventDefault();
           e.stopPropagation();
-          dispatch(desktopAssetDragStarted({ asset: e.dataTransfer.items[0] }));
+          if (e.dataTransfer.items[0].type.startsWith('image') && e.dataTransfer.items[0].kind === 'file') {
+            dispatch(desktopAssetDragStarted({ asset: e.dataTransfer.items[0] }));
+          }
```
rart, line 425: Desktop asset **and** assets panel drag & drop continues to work with this change applied?

aachavarria: this only affect drag for desktop, i tested on my machine and it is working, but we need to test on your laptop to see if it continue working as expected on mac

rart (approved):

PR #2728, opened 2021-12-28
[5182] Component ID control wraps when too long (doesn't look good)

aachavarria:
https://github.com/craftercms/craftercms/issues/5182


rart (changes requested):
Could we make the thing auto-grow? A hard limit is fine but making it flexible so that at least most cases look as they should would be best.

aachavarria:
> Could we make the thing auto-grow? A hard limit is fine but making it flexible so that at least most cases look as they should would be best.

With the last update the left part of the path should always be visible, and the right part will take the remaining space

rart (approved):

PR #2733, opened 2021-12-30
[5179] Paginate by 9 instead of 10 for 3 column listings

aachavarria:
https://github.com/craftercms/craftercms/issues/5179


rart (approved):
Can something be done on create site dialog or is that one dynamic in its number of columns?

aachavarria:
> Can something be done on create site dialog or is that one dynamic in its number of columns?

create site dialog does not have pagination

PR #2738, opened 2022-01-03
[5183] Warning message about saving invalid XML files is missing in version 4.

aachavarria:
https://github.com/craftercms/craftercms/issues/5174
https://github.com/craftercms/craftercms/issues/5183

### DIFF
--- a/ui/app/src/components/PluginConfigDialog/PluginConfigDialogContainer.tsx
+++ b/ui/app/src/components/PluginConfigDialog/PluginConfigDialogContainer.tsx
@@ -24,7 +24,7 @@ import SecondaryButton from '../SecondaryButton';
 import { FormattedMessage, useIntl } from 'react-intl';
 import PrimaryButton from '../PrimaryButton';
 import { getPluginConfiguration, setPluginConfiguration } from '../../services/marketplace';
-import { useActiveSiteId } from '../../hooks';
+import { useActiveSiteId, useUpdateRefs } from '../../hooks';
 import { useDispatch } from 'react-redux';
 import { showErrorDialog } from '../../state/reducers/dialogs/error';
 import { showSystemNotification } from '../../state/actions/system';
@@ -40,6 +40,9 @@ export function PluginConfigDialogContainer(props: PluginConfigDialogContainerPr
   const dispatch = useDispatch();
   const [disabledSaveButton, setDisabledSaveButton] = useState(true);
   const { formatMessage } = useIntl();
+  const functionRefs = useUpdateRefs({
+    onSubmittingAndOrPendingChange
+  });
 
   useEffect(() => {
     setLoading(true);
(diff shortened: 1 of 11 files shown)


ui/app/src/components/SiteConfigurationManagement/SiteConfigurationManagement.tsx
```diff
@@ -184,7 +185,21 @@ export default function SiteConfigurationManagement(props: SiteConfigurationMana
...
+          message: formatMessage(translations.xmlContainsErrors, {
+            errors: doc
+          }),
+          options: {
+            variant: 'error'
+          }
+        })
+      );
+      return;
+    }
```
rart: extract to local function to use in both places

ui/app/src/index.tsx
```diff
@@ -28,3 +28,4 @@ if (process.env.NODE_ENV === 'production') {
   const elem = document.getElementById('root');
   ReactDOM.render(<App />, elem);
 }
+export { parseValidateDocument } from './utils/xml';
```
rart: why? rollback

ui/app/src/components/PluginConfigDialog/PluginConfigDialogContainer.tsx
```diff
@@ -93,7 +108,7 @@ export function PluginConfigDialogContainer(props: PluginConfigDialogContainerPr
           onSaved();
         },
         error: ({ response }) => {
-          onSubmittingAndOrPendingChange({ isSubmitting: false });
+          onSubmittingAndOrPendingChange({ isSubmitting: false, hasPendingChanges: true });
```
rart: It feels like by this point it should already be true. Why do you need to set it here?

rart (changes requested):

ui/app/src/hooks/useEnhancedDialogState.ts
```diff
@@ -75,5 +85,5 @@ export function useEnhancedDialogState(initialState?: Partial<EnhancedDialogStat
       onHasPendingChange,
       onSubmittingAndOrPendingChange
     };
-  }, [setState, state]);
+  }, [initialState, setState, state]);
```
rart: This is dangerous to include here as users of the effect most likely won’t memo the initial state. Ignore it and add a comment to justify the eslint ignore

ui/app/src/utils/xml.ts
```diff
@@ -181,3 +181,14 @@ export function deserialize(xml: string | Node, options?: X2jOptionsOptional): a
 export function cdataWrap(value: string): string {
   return `<![CDATA[${value}]]>`;
 }
+
+export function parseValidateDocument(content: string): XMLDocument | string {
```
rart, line 185: Is the parsed document ever used for something? I’m thinking the function should have a single responsibility.

rart (changes requested):

ui/app/src/utils/xml.ts
aachavarria: yes, the doc is used

rart (approved):

PR #2755, opened 2022-01-12
[3.1] Add style_formats config to RTE5

phuongnq:
https://github.com/craftercms/craftercms/issues/5237

Sample configuration:

```
        <styleFormats>
                <![CDATA[
            [
    {title : 'Bold text', inline : 'b'},
    {title : 'Red text', inline : 'span', styles : {color : '#ff0000'}},
    {title : 'Red header', block : 'h1', styles : {color : '#ff0000'}},
    {title : 'Example 1', inline : 'span', classes : 'example1'},
    {title : 'Example 2', inline : 'span', classes : 'example2'},
    {title : 'Table styles'},
    {title : 'Table row 1', selector : 'tr', classes : 'tablerow1'}
]
]]>
        </styleFormats>
        <styleFormatsMerge>true</styleFormatsMerge>
```

### DIFF
--- a/static-assets/components/cstudio-forms/controls/rte-tinymce5.js
+++ b/static-assets/components/cstudio-forms/controls/rte-tinymce5.js
@@ -229,6 +229,8 @@ CStudioAuthoring.Module.requireModule(
             toolbarConfig2,
             toolbarConfig3,
             toolbarConfig4,
+            styleFormats,
+            styleFormatsMerge,
             templates;
 
           containerEl.id = this.id;
@@ -307,6 +309,14 @@ CStudioAuthoring.Module.requireModule(
 
           rteStyleOverride = rteConfig.rteStyleOverride ? rteConfig.rteStyleOverride : null;
 
+          // use tinymce default if not set
+          styleFormats =
+            rteConfig.styleFormats && rteConfig.styleFormats.length != 0
+            ? Function(`"use strict"; return (${rteConfig.styleFormats})`)()
+            : undefined;
+
+          styleFormatsMerge = rteConfig.styleFormatsMerge === 'true'
+
           const codeEditorWrap = rteConfig.codeEditorWrap ? rteConfig.codeEditorWrap === 'true' : false;
 
           const external = {
@@ -369,6 +379,10 @@ CStudioAuthoring.Module.requireModule(
             code_editor_wrap: codeEditorWrap,
 
             external_plugins: external,
+
+            style_formats: styleFormats,
+
+            style_formats_merge: styleFormatsMerge,
 
             setup: function (editor) {
               var addPadding = function () {


static-assets/components/cstudio-forms/controls/rte-tinymce5.js
```diff
@@ -370,6 +380,11 @@ CStudioAuthoring.Module.requireModule(
 
             external_plugins: external,
 
+            style_formats: styleFormats,
+
+            // default to merge if no value is set
+            style_formats_merge: (styleFormats.length === 0 && styleFormatsMerge === null) || styleFormatsMerge,
```
rart: This is a change in current default behaviours. Why default it to try if nothing is set? The safe default seems to be `false`, which is [tinymce's default](https://www.tiny.cloud/docs/configure/editor-appearance/#style_formats_merge).

Are we trying to change the default for every body? @sumerjabri @russdanner 

A bit cleaner to initialize the value —with all/whatever logic above— and here just assign the prop.
```suggestion
            style_formats_merge: styleFormatsMerge,
```

static-assets/components/cstudio-forms/controls/rte-tinymce5.js
```diff
@@ -307,6 +309,14 @@ CStudioAuthoring.Module.requireModule(
 
           rteStyleOverride = rteConfig.rteStyleOverride ? rteConfig.rteStyleOverride : null;
 
+          styleFormats =
+            rteConfig.styleFormats && rteConfig.styleFormats.length != 0
+            ? Function(`"use strict"; return (${rteConfig.styleFormats})`)()
+            : [];
+
+          styleFormatsMerge = rteConfig.styleFormatsMerge && rteConfig.styleFormatsMerge !== 0
+            ? rteConfig.styleFormatsMerge === 'true' : null;
```
rart: `rteConfig.styleFormatsMerge && rteConfig.styleFormatsMerge !== 0` looks redundant. If `rteConfig.styleFormatsMerge` is true, `rteConfig.styleFormatsMerge !== 0` will always be true too.

As said on the comment above, initialize here using whatever logic and just use below.

This would be sufficient to keep the current defaults:
```suggestion
          styleFormatsMerge = rteConfig.styleFormatsMerge === 'true';
```

rart (changes requested):

static-assets/components/cstudio-forms/controls/rte-tinymce5.js
phuongnq: @rart We don't want to change the default. Actually, it was the opposite, I want to keep the default.
In the previous source code, we have `style_formats` default to `[]`. This caused the Format list to be empty and override TinyMCE default.

But I agreed that the code for `style_formats_merge` is not good. I changed back to the value as you suggested. And if there is no value set for `styleFormats`, the variable will be `undefined`. This should fix the default issue.

static-assets/components/cstudio-forms/controls/rte-tinymce5.js
phuongnq: Thanks. I fixed!

rart (approved):

PR #2758, opened 2022-01-12
[5157] Edit Template/Controllers when no template is defined in content type

aachavarria:
https://github.com/craftercms/craftercms/issues/5157


ui/app/src/services/contentTypes.ts
```diff
@@ -421,6 +421,25 @@ export function deleteContentType(site: string, contentTypeId: string): Observab
...
+      const properties = doc.querySelectorAll('properties > property');
+      const property = Array.from(properties).find(
+        (node) => node.querySelector('name').innerHTML.trim() === 'display-template'
+      );
+      if (property) {
+        property.querySelector('value').innerHTML = displayTemplate;
+        return writeConfiguration(site, path, module, beautify(serialize(doc)));
+      } else {
+        return of(false);
+      }
```
rart: On this case, if it's not there yet, you need to create it.

rart (changes requested):

rart (approved):

PR #2767, opened 2022-01-17
[next] Add rte paste word plugin

phuongnq:
https://github.com/craftercms/craftercms/issues/4035

### DIFF
--- a/static-assets/components/cstudio-forms/controls/rte.js
+++ b/static-assets/components/cstudio-forms/controls/rte.js
@@ -218,6 +218,7 @@ CStudioAuthoring.Module.requireModule(
             pluginList,
             extendedElements,
             rteStylesheets,
+            rtePasteWordElements,
             rteStyleOverride,
             toolbarConfig1,
             toolbarConfig2,
@@ -301,6 +302,12 @@ CStudioAuthoring.Module.requireModule(
               ? rteConfig.rteStylesheets.link
               : null;
 
+          rtePasteWordElements = rteConfig.rtePasteWordElements
+            ? rteConfig.rtePasteWordElements
+            : '-strong/b,-em/i,-u,-span,-p,-ol,-ul,-li,-h1,-h2,-h3,-h4,-h5,-h6,-h7,-h8,-pre,-p/div,' +
+              '-a[href|name],sub,sup,pre,blockquote,strike,br,del,' +
+              'table[width],tr,td[colspan|rowspan|width],th[colspan|rowspan|width],thead,tfoot,tbody';
+
           rteStyleOverride = rteConfig.rteStyleOverride ? rteConfig.rteStyleOverride : null;
 
           try {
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -341,7 +347,8 @@ CStudioAuthoring.Module.requireModule(
 
           const external = {
             acecode: '/studio/static-assets/js/tinymce-plugins/ace/plugin.min.js',
-            paste_cleanup: '/studio/static-assets/js/tinymce-plugins/paste_cleanup/plugin.js'
+            paste_cleanup: '/studio/static-assets/js/tinymce-plugins/paste_cleanup/plugin.js',
+            paste_text: '/studio/static-assets/js/tinymce-plugins/paste_text/plugin.js',
```
rart: Shouldn't we call this `past_from_word`? 

Is this specific to word or general paste clean up? If it works for anything, what's the difference to our `paste_cleanup` plugin? Could/should this one replace `paste_cleanup` all together?

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -301,6 +302,11 @@ CStudioAuthoring.Module.requireModule(
               ? rteConfig.rteStylesheets.link
               : null;
 
+          rtePasteWordElements =
+            rteConfig.rtePasteWordElements && rteConfig.rtePasteWordElements.length != 0
+              ? rteConfig.rtePasteWordElements
+              : '';
```
rart: Looks like this will be a string. The first part of the condition is sufficient.
```suggestion
          rtePasteWordElements = rteConfig.rtePasteWordElements ? rteConfig.rtePasteWordElements : '';
```

static-assets/components/cstudio-forms/controls/rte.js
rart: Please make sure you're running prettier to format code on any modified files. We don't use trailing commas.
```suggestion
            paste_text: '/studio/static-assets/js/tinymce-plugins/paste_text/plugin.js'
```

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -429,6 +436,8 @@ CStudioAuthoring.Module.requireModule(
 
             style_formats_merge: styleFormatsMerge,
 
+            paste_word_valid_elements: rtePasteWordElements,
```
rart, line 459: Per [tinymce docs](https://www.tiny.cloud/docs/plugins/opensource/paste/#paste_word_valid_elements), "this property is deprecated and will be removed in v6". You are using this for **your** plugin, correct? Not relying on tinymce's paste plugin?

For tinymce's plugin, `paste_word_valid_elements` needs to be `false`. Do you need to do something of the sorts?

rart (changes requested):

static-assets/components/cstudio-forms/controls/rte.js
phuongnq: Hi Roy, yes, we are using this for our own plugin. We don't rely on the default `paste` plugin. But our plugin has some logic borrowed from `paste`.  As of checking with @russdanner, we want to have an option to update valid elements which should be kept while pasting a Word document. Only valid elements are kept.

phuongnq:
Seem like `paste_cleanup` works well with Word documents. I will close this for now.

PR #2776, opened 2022-01-18
[5193] Close content confirm form dialog does not support dark theme

jvega190:
https://github.com/craftercms/craftercms/issues/5193


rart (commented):
Won't be taking this for now. Keep the branch around in case we need to revisit.

PR #2780, opened 2022-01-19
[5138] Remove legacy template-editor code

jvega190:
https://github.com/craftercms/craftercms/issues/5138


rart (changes requested):
- Clean up constants that aren't used anymore
- Consolidate the legacy opening of the next template editor

rart (approved):

PR #2784, opened 2022-01-20
[5151] Switch RTE to Code editor in full-screen mode does not work

jvega190:
https://github.com/craftercms/craftercms/issues/5151


static-assets/js/tinymce-plugins/ace/plugin.js
```diff
@@ -39,27 +39,28 @@ tinymce.PluginManager.add(PLUGIN_NAME, function (editor, url) {
...
   };
 
+  const onAction = (context, editor) => {
+    const isFullscreen = editor.plugins.fullscreen.isFullscreen();
+    if (isFullscreen || !inlineMode) {
```
jvega190, line 44: @rart the 4.0.x plugin has an inlineMode parameter that is needed for ICE rte control, 3.1.x doesn't need it.

rart (approved):

PR #2790, opened 2022-01-24
[3.1] Add tinymce init extended options

phuongnq:
https://github.com/craftercms/craftercms/issues/5261


rart (changes requested):
Please use the same approach used in v4 for `extendedOptions`.
See https://github.com/craftercms/studio-ui/blob/develop/ui/guest/src/controls/rte.ts#L225

phuongnq:
@rart Thanks. I updated PR to have read-only options as v4

rart (approved):

PR #2797, opened 2022-01-26
[5275] Slugify page URL fields

rart:
Added npm:slugify@1.6.5

craftercms/craftercms#5275

### DIFF
--- a/static-assets/components/cstudio-forms/controls/file-name.js
+++ b/static-assets/components/cstudio-forms/controls/file-name.js
@@ -80,8 +80,6 @@ var slugify = (function () {
   return replace;
 })();
 
-slugify.extend({ '&': '' });
-
 CStudioForms.Controls.FileName =
   CStudioForms.Controls.FileName ||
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/controls/file-name.js
```diff
@@ -14,6 +14,74 @@
...
+  }
+
+  replace.extend = function (customMap) {
+    Object.assign(charMap, customMap);
+  };
+
+  return replace;
+})();
+
+slugify.extend({ '&': '' });
```
avasquez614: @rart  is this change so that `&` is just removed, instead of being replaced for the word `and`?

avasquez614 (changes requested):

static-assets/components/cstudio-forms/controls/file-name.js
rart: yes, @avasquez614. in accordance to the specs, & should be removed. we're not giving config options at this stage so it's currently baked behaviour.

avasquez614 (approved):

PR #2799, opened 2022-01-26
[5272] Content Types: Preview Image is not set in config when an Image is uploaded

aachavarria:
https://github.com/craftercms/craftercms/issues/5272


static-assets/components/cstudio-admin/mods/content-type-propsheet/image.js
```diff
@@ -160,7 +159,8 @@ YAHOO.extend(
                   }
                 }
                 image.addEventListener('load', imageLoaded, false);
-                image.addEventListener('error', function () {
+                image.addEventListener('error', function (e) {
+                  console.log(e);
```
rart: x
```suggestion
```

rart (changes requested):

rart (approved):

PR #2804, opened 2022-01-28
[5266] Make it easy to edit viewed files

aachavarria:
https://github.com/craftercms/craftercms/issues/5266


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -1917,6 +1917,36 @@ var CStudioForms =
             YAHOO.util.Event.addListener(window, 'beforeunload', unloadFn, me);
             YAHOO.util.Event.addListener(closeButtonEl, 'click', cancelFn, me);
 
+            const canEdit = CrafterCMSNext.system.store
+              .getState()
+              .user.permissionsBySite[CrafterCMSNext.system.store.getState().sites.active].includes('content_write');
```
sumerjabri: Don't we have a set of constants for these strings? What happens when we change these strings in the back?

ui/app/src/utils/i18n-legacy.ts
```diff
@@ -723,6 +723,10 @@ export const formEngineMessages = defineMessages({
   useMVS: {
     id: 'formEngine.useMVS',
     defaultMessage: 'Use _mvs postfix (backward compat)'
+  },
+  contentWritePermission: {
+    id: 'formEngine.contentWritePermission',
+    defaultMessage: "Current user don't have edit permission"
```
sumerjabri: The current user doesn't have the necessary permissions to edit this content item"

sumerjabri (changes requested):
What about localization?
Does this cover all the use-cases of editing: FTL, Groovy, text files, etc.?
See other notes.

static-assets/components/cstudio-forms/forms-engine.js
aachavarria: I think we don´t have constants for these strings, another option is use the item available actions

sumerjabri: That's even better, let's use available actions.

aachavarria:
> What about localization? Does this cover all the use-cases of editing: FTL, Groovy, text files, etc.? See other notes.

that cases are covered on the codeEditor, however I'm doing a update to use available actions for both cases

rart (approved):

PR #2814, opened 2022-02-03
[5283] Fix loading state of group buttons

aachavarria:
https://github.com/craftercms/craftercms/issues/5283

### DIFF
--- a/ui/app/src/components/SplitButton/SplitButton.tsx
+++ b/ui/app/src/components/SplitButton/SplitButton.tsx
@@ -24,6 +24,7 @@ import {
   setStoredSaveButtonSubAction
 } from '../../utils/state';
 import { useActiveUser } from '../../hooks';
+import { FormattedMessage } from 'react-intl';
 
 export function SplitButton(props: SplitButtonProps) {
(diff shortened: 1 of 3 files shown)


ui/app/src/components/SplitButton/SplitButton.tsx
```diff
@@ -81,7 +82,9 @@ export function SplitButton(props: SplitButtonProps) {
   return (
     <>
       {loading ? (
-        <PrimaryButton loading disabled />
+        <PrimaryButton loading disabled>
```
rart: Fix this at the SplitButtonUI level to use the MUI `loading` prop of button and freeze the buttons at that level.

rart (changes requested):

rart (changes requested):
Again, the idea is to employ the `loading` prop of MUI's `Button` component. Is that not possible when used in a button group?

rart (changes requested):
The idea is for the button to look the same and no layout changes (no change of the split to the regular button). If the loading prop doesn't work in the button group fake it out by disabling the button group and replacing the text with a spinner. Use some styles for the width of the button not to change as it switches between states.

rart (approved):

PR #2816, opened 2022-02-04
[5286] Change Template Name: to Content Type:

jvega190:
https://github.com/craftercms/craftercms/issues/5286


jvega190:
Added some translations that were missing in korean, also, the translation for german was correct.

rart (approved):

PR #2819, opened 2022-02-07
[5287] Code editor doesn't focus on the edit panel

aachavarria:
https://github.com/craftercms/craftercms/issues/5287
https://github.com/craftercms/craftercms/issues/5266


rart (commented):
PRs got mixed. Please remove the LogConsole work from this PR.

ui/app/src/components/CodeEditorDialog/CodeEditorDialogContainer.tsx
```diff
@@ -233,6 +233,7 @@ export function CodeEditorDialogContainer(props: CodeEditorDialogContainerProps)
         <ConditionalLoadingState isLoading={loading} classes={{ root: classes.loadingState }}>
           <AceEditor
             ref={editorRef}
+            autoFocus
```
rart: 1. What happens to this flag when opening readonly?
2. When you go from ready only to edit, does the autoFocus pick up?

rart (changes requested):
Let's try adding a `default` background to the editor when in readonly; make it more distinct from the editable editor

ui/app/src/components/CodeEditorDialog/CodeEditorDialogContainer.tsx
aachavarria: checking

rart (approved):

PR #2851, opened 2022-02-22
[5332] Preview Image upload failed 

jvega190:
https://github.com/craftercms/craftercms/issues/5332

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-type-propsheet/image.js
+++ b/static-assets/components/cstudio-admin/mods/content-type-propsheet/image.js
@@ -223,8 +223,7 @@ YAHOO.extend(
 
     createImageData: function (imageData, path) {
       var url = this.createPreviewUrl(
-        CStudioAuthoringContext.baseUri +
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-admin/mods/content-type-propsheet/image.js
```diff
@@ -223,8 +223,7 @@ YAHOO.extend(
 
     createImageData: function (imageData, path) {
       var url = this.createPreviewUrl(
-        CStudioAuthoringContext.baseUri +
```
rart, line 226: @jvega190 why is baseUri not correct?

rart (approved):

static-assets/components/cstudio-admin/mods/content-type-propsheet/image.js
jvega190: createPreviewUrl already adds the baseUri, so it was being added twice

PR #2862, opened 2022-03-02
[5347] Cut-Paste of a page as a child of another page does not update the Sidebar

jvega190:
https://github.com/craftercms/craftercms/issues/5347


ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -301,21 +301,21 @@ export function PathNavigator(props: PathNavigatorProps) {
...
             }
           }
+          if (state.leaves.some((path) => withoutIndex(path) === withoutIndex(payload.target))) {
+            dispatch(
+              pathNavigatorUpdate({
+                id,
+                leaves: state.leaves.filter((path) => withoutIndex(path) !== withoutIndex(payload.target))
+              })
+            );
+          }
```
rart: Ping me to discusss.

rart (changes requested):
This fix looks like a race condition, or a timing-based fix which could end up being still a race condition. I don't think we need the leaves state anymore since the api now provides `childCount`. Fix might be to remove the leaves state.

rart (approved):

PR #2874, opened 2022-03-09
Add check if node content element is null while append child

phuongnq:
https://github.com/craftercms/craftercms/issues/4209


phuongnq:
closed as this need modify `yui2`

PR #2886, opened 2022-03-18
[5384] The browse window for pickers cut's out the tree when the tree is deep

jvega190:
https://github.com/craftercms/craftercms/issues/5384

### DIFF
--- a/ui/scss/browse.scss
+++ b/ui/scss/browse.scss
@@ -115,7 +115,7 @@ body {
 			padding: 20px 20px 20px 0;
 			border-radius: 5px;
 			float: left;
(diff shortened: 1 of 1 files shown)


ui/scss/browse.scss
```diff
@@ -115,7 +115,7 @@ body {
 			padding: 20px 20px 20px 0;
 			border-radius: 5px;
 			float: left;
-			height: calc(100vh - 95px);
+			height: calc(100vh - 117px);
```
jvega190, line 118: Scrollbar was there but was not showing

rart (approved):

PR #2887, opened 2022-03-18
[5384] The browse window for pickers cut's out the tree when the tree is deep

jvega190:
https://github.com/craftercms/craftercms/issues/5384


ui/scss/src/browse.scss
```diff
@@ -67,7 +67,7 @@ body {
       padding: 20px 20px 20px 0;
       border-radius: 5px;
       float: left;
-      height: calc(100vh - 95px);
+      height: calc(100vh - 160px);
```
jvega190, line 70: Differente between 4.0 and 3.1.x is because the dialog style updates in 4

rart (approved):

PR #2900, opened 2022-03-30
[5417] Available actions are not refreshed for current item upon a change event

jvega190:
https://github.com/craftercms/craftercms/issues/5417

### DIFF
--- a/ui/app/src/state/epics/content.ts
+++ b/ui/app/src/state/epics/content.ts
@@ -66,7 +66,6 @@ import {
 import { isEditableAsset } from '../../utils/content';
 import {
   blockUI,
-  emitSystemEvent,
   lockContentEvent,
   showDeleteItemSuccessNotification,
   showDuplicatedItemSuccessNotification,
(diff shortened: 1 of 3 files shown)


ui/app/src/state/epics/content.ts
```diff
@@ -508,8 +510,21 @@ const content: CrafterCMSEpic[] = [
...
       })
-    )
+    ),
   // endregion
+  // region Emit System Event
+  (action$, state$) =>
+    action$.pipe(
+      ofType(emitSystemEvent.type),
+      withLatestFrom(state$),
+      filter(([{ payload }]) => payload.type === lockContentEvent.type),
```
rart: Lets go ahead and dispatch system events to the store (on epics/system) so we don't have to do these filterings like we've had to do here and in the content reducer. On the emitSystemEvent epic, after the tap, map(action => action) and change here and on the reducer to simply wait for the actual action rather than having to inspect the system events.

rart (changes requested):

ui/app/src/state/epics/content.ts
```diff
@@ -508,8 +509,20 @@ const content: CrafterCMSEpic[] = [
...
+    action$.pipe(
+      ofType(lockContentEvent.type),
+      withLatestFrom(state$),
+      switchMap(([{ payload }, state]) =>
+        fetchSandboxItemService(state.sites.active, payload.targetPath).pipe(
+          map((item) => fetchSandboxItemComplete({ item })),
+          catchAjaxError(fetchSandboxItemFailed)
+        )
+      )
+    )
```
rart, line 526:
```suggestion
    )
    // endregion
```

ui/app/src/state/epics/content.ts
```diff
@@ -508,8 +509,20 @@ const content: CrafterCMSEpic[] = [
...
       })
-    )
+    ),
   // endregion
+  // region Lock Content Event
+  (action$, state$) =>
+    action$.pipe(
+      ofType(lockContentEvent.type),
+      withLatestFrom(state$),
+      switchMap(([{ payload }, state]) =>
```
rart:
```suggestion
      withLatestFrom(state$),
      filter(state has item already so we don't end up with things that aren't in use)
      switchMap(([{ payload }, state]) =>
```

rart (approved):

rart (changes requested):
See comments above

rart (approved):

PR #2902, opened 2022-03-31
[5421] When duplicating items or content that open 'groovy' or 'ftl' editors and then closing the editor, UI should call the item_unlock API

jvega190:
https://github.com/craftercms/craftercms/issues/5421

### DIFF
--- a/ui/app/src/state/epics/content.ts
+++ b/ui/app/src/state/epics/content.ts
@@ -56,6 +56,7 @@ import {
 } from '../../services/content';
 import { merge, NEVER, Observable, of } from 'rxjs';
 import {
+  closeCodeEditorDialog,
   closeConfirmDialog,
   closeDeleteDialog,
   showCodeEditorDialog,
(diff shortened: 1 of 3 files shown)


ui/app/src/state/epics/content.ts
```diff
@@ -283,12 +286,14 @@ const content: CrafterCMSEpic[] = [
...
             const editableAsset = isEditableAsset(payload.path);
             if (editableAsset) {
+              const extension = getFileExtensionFromPath(path);
               return showCodeEditorDialog({
                 authoringBase: state.env.authoringBase,
                 site: state.sites.active,
                 path,
-                type: 'asset',
-                onSuccess: payload.onSuccess
+                mode: getEditorModeFromExtension(extension),
```
jvega190, line 294: @rart this is not directly related with the original issue, I need to discuss this changes with you

ui/app/src/state/epics/content.ts
```diff
@@ -281,14 +282,16 @@ const content: CrafterCMSEpic[] = [
...
+            const mode = getEditorMode(state.content.itemsByPath[payload.path].mimeType);
             const editableAsset = isEditableAsset(payload.path);
             if (editableAsset) {
               return showCodeEditorDialog({
                 authoringBase: state.env.authoringBase,
                 site: state.sites.active,
                 path,
-                type: 'asset',
-                onSuccess: payload.onSuccess
+                mode,
```
jvega190, line 292: `type` was being sent instead of `mode`, so the editor mode wasn't properly set

rart (approved):

PR #2907, opened 2022-04-04
[5433] After fetching DetailedItem, if there's a change on a request publishing date it won't update in another browser/user

jvega190:
https://github.com/craftercms/craftercms/issues/5433


ui/app/src/state/epics/content.ts
```diff
@@ -526,6 +527,14 @@ const content: CrafterCMSEpic[] = [
...
         )
       )
+    ),
+  // endregion
+  // region Show Item Mega Menu
+  (action$, state$) =>
+    action$.pipe(
+      ofType(showItemMegaMenu.type),
+      withLatestFrom(state$),
+      switchMap(([{ payload }]) => [completeDetailedItem({ path: payload.path, force: true })])
```
rart: `map` instead of `switchMap` returning an array

rart (changes requested):
I'm thinking we should rather request the detailed items when the publish option is clicked. 
This doesn't cover multiple items sent to the dialog like it can happen on the dashboard widgets, for instance.
Because DetailedItems are teaching on the backend, we should do this only when needed (i.e. when state is submitted or scheduled?).

ui/app/src/state/epics/content.ts
```diff
@@ -178,6 +182,17 @@ const content: CrafterCMSEpic[] = [
         )
       )
     ),
+  (action$, state$) =>
+    action$.pipe(
+      ofType(fetchDetailedItems.type),
+      withLatestFrom(state$),
+      switchMap(([{ payload }, state]) =>
+        forkJoin(payload.paths.map((path) => fetchDetailedItemService(state.sites.active, path))).pipe(
```
rart: Let's make a `fetchDetailedItems(site, paths)` service that does this instead

ui/app/src/state/epics/dialogs.ts
```diff
@@ -217,6 +219,13 @@ const dialogEpics: CrafterCMSEpic[] = [
         hostToGuest$.next(action);
       }),
       ignoreElements()
+    ),
+  // endregion
+  // region Show Publish Dialog
+  (action$) =>
+    action$.pipe(
+      ofType(showPublishDialog.type),
+      map(({ payload }) => fetchDetailedItems({ paths: payload.items.map((item) => item.path) }))
```
rart, line 229: `filter(paths.length)`

rart (changes requested):

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -45,9 +45,11 @@ import { pluckProps } from '../../utils/object';
 import moment from 'moment-timezone';
 import { updatePublishDialog } from '../../state/actions/dialogs';
 import { approve, publish, requestPublish } from '../../services/workflow';
+import { useDetailedItems } from '../../hooks';
```
rart:
```suggestion
import useDetailedItems from '../../hooks/useDetailedItems';
```

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -173,13 +175,13 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
     () => ({
-      items,
+      items: !detailedItems.isFetching ? Object.values(detailedItems.itemsByPath) : null,
       error: state.error,
       submitting: isSubmitting,
       publishingTargets,
       myPermissions
     }),
-    [items, state.error, isSubmitting, publishingTargets, myPermissions]
+    [detailedItems, state.error, isSubmitting, publishingTargets, myPermissions]
```
rart: The return value of `useDetailedItems` is a new object reference every time. This memo will renew every render. The internal values `items` and `isFetching` are ok to use, though. Use these directly instead.

rart, line 184: The return value of `useDetailedItems` is a new object reference every time. This memo will renew every render.

rart (changes requested):

ui/app/src/state/epics/dialogs.ts
```diff
@@ -217,6 +219,17 @@ const dialogEpics: CrafterCMSEpic[] = [
...
       }),
       ignoreElements()
+    ),
+  // endregion
+  // region Show Publish Dialog
+  (action$) =>
+    action$.pipe(
+      ofType(showPublishDialog.type),
+      filter(({ payload }) => {
+        const length = payload.items.map((item) => item.path).length;
```
rart: Why are you mapping just to get the length? Seems like useless processing.

rart (changes requested):

ui/app/src/state/epics/dialogs.ts
```diff
@@ -217,6 +219,17 @@ const dialogEpics: CrafterCMSEpic[] = [
...
+    ),
+  // endregion
+  // region Show Publish Dialog
+  (action$) =>
+    action$.pipe(
+      ofType(showPublishDialog.type),
+      filter(({ payload }) => {
+        const length = payload.items.length;
+        return length && length > 0;
+      }),
```
rart:
```suggestion
      filter(({ payload }) => Boolean(payload.items?.length)),
```

rart (approved):

PR #2918, opened 2022-04-08
[5438] Path Navigator refreshes twice with skeletonization unlike Tree Navigator

jvega190:
https://github.com/craftercms/craftercms/issues/5438


ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -247,7 +247,7 @@ export function PathNavigator(props: PathNavigatorProps) {
 
           if (withoutIndex(targetPath) === withoutIndex(path)) {
             // if path being deleted is the rootPath
-            dispatch(pathNavigatorRefresh({ id }));
+            dispatch(pathNavigatorBackgroundRefresh({ id }));
```
rart: Please retest how it behaves in this case. If the root path is deleted, essentially the whole content should change to the "Path doesn't exist error". In this case, skeleton may be in order.

rart (changes requested):

rart (approved):

PR #2923, opened 2022-04-08
[5469] Deleting the last item on a page in sidebar doesn't refresh the sidebar

jvega190:
https://github.com/craftercms/craftercms/issues/5469

### DIFF
--- a/ui/app/src/state/reducers/pathNavigator.ts
+++ b/ui/app/src/state/reducers/pathNavigator.ts
@@ -134,21 +134,26 @@ const reducer = createReducer<LookupTable<PathNavigatorStateProps>>(
     }),
     [pathNavigatorFetchPathComplete.type]: (state, { payload: { id, children, parent } }) => {
       const path = parent?.path ?? state[id].currentPath;
-      return {
-        ...state,
-        [id]: {
-          ...state[id],
-          currentPath: path,
-          breadcrumb: getIndividualPaths(withoutIndex(path), withoutIndex(state[id].rootPath)),
-          itemsInPath: children.length === 0 ? [] : children.map((item) => item.path),
-          levelDescriptor: children.levelDescriptor?.path,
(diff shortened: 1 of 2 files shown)


ui/app/src/state/reducers/pathNavigator.ts
```diff
@@ -134,21 +134,26 @@ const reducer = createReducer<LookupTable<PathNavigatorStateProps>>(
...
-          offset: children.offset,
-          limit: children.limit,
-          isFetching: false,
-          error: null
-        }
-      };
+
+      if (children.offset >= children.limit && children.length === 0) {
+        return state;
+      } else {
```
rart: First line is not used on the if statement nor in it's body. Move to the else.
```ts
       if (children.offset >= children.limit && children.length === 0) {
         return state;
       } else {
         const path = parent?.path ?? state[id].currentPath;
```

ui/app/src/state/reducers/pathNavigator.ts
```diff
@@ -134,21 +134,26 @@ const reducer = createReducer<LookupTable<PathNavigatorStateProps>>(
...
-          levelDescriptor: children.levelDescriptor?.path,
-          total: children.total,
-          offset: children.offset,
-          limit: children.limit,
-          isFetching: false,
-          error: null
-        }
-      };
+
+      if (children.offset >= children.limit && children.length === 0) {
```
rart, line 137: What is this trying to accomplish? Add comment/description.
When/why does this happen?

rart (changes requested):

ui/app/src/state/reducers/pathNavigator.ts
jvega190: Comment added according to our conversation

rart (approved):

PR #2932, opened 2022-04-12
[5471] Form engine cancel or dialog close buttons do not close the form

jvega190:
https://github.com/craftercms/craftercms/issues/5471

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -1832,7 +1832,7 @@ var CStudioForms =
               if (acnDraftContent) {
                 unlockBeforeCancel(path);
               } else {
-                if (!form.readOnly && path && path.indexOf('.xml') != -1) {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -1832,7 +1832,7 @@ var CStudioForms =
               if (acnDraftContent) {
                 unlockBeforeCancel(path);
               } else {
-                if (!form.readOnly && path && path.indexOf('.xml') != -1) {
+                if (!form.readOnly && path && path.indexOf('.xml') != -1 && !me.config.isInclude) {
```
rart, line 1835: Confirm tested with embedded, shared, parent and nested levels of embedded and shared.

jvega190: Confirmed tested with embedded, shared and multiple levels of embedded and shared

rart (approved):

PR #2938, opened 2022-04-13
[5487] Some XB actions (like sort) cause components to stay locked.

jvega190:
https://github.com/craftercms/craftercms/issues/5487

### DIFF
--- a/ui/guest/src/store/epics/root.ts
+++ b/ui/guest/src/store/epics/root.ts
@@ -269,9 +269,9 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
         };
 
         const models = getCachedModels();
-        const modelId = action.payload.record.modelId;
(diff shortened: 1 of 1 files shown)


ui/guest/src/store/epics/root.ts
```diff
@@ -269,9 +269,9 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
         };
 
         const models = getCachedModels();
-        const modelId = action.payload.record.modelId;
-        const parentModelId = getParentModelId(modelId, models, modelHierarchyMap);
-        const path = models[parentModelId ?? modelId].craftercms.path;
+        const dropZone = dragContext.dropZone;
```
jvega190, line 272: I tested in all cases (placing detached assets, upload assets, sort components, placing new/existing components) and also edition on content inside a component (like updating an image of a component) and the path to be locked is correct.

rart (approved):

PR #2943, opened 2022-04-19
[5493] Increase Audit Report User drop down to display all users

jvega190:
https://github.com/craftercms/craftercms/issues/5493

### DIFF
--- a/static-assets/scripts/admin.js
+++ b/static-assets/scripts/admin.js
@@ -435,8 +435,13 @@
       };
 
       var getUsers = function (site) {
+        const params = {
+          limit: 60,
+          ...(site ? { site } : {})
+        };
+
         adminService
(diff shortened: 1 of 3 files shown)


static-assets/scripts/admin.js
```diff
@@ -435,8 +435,13 @@
       };
 
       var getUsers = function (site) {
+        const params = {
+          limit: 60,
+          ...(site ? { site } : {})
+        };
+
         adminService
-          .getUsers(site)
```
jvega190, line 439: This was not being properly sent, since the internal service receives the params object instead of only the site. That was not causing an issue though (still was incorrect), but with the addition of the limit param the change was needed.

rart (approved):

PR #2948, opened 2022-04-20
[5505] Switch the Form Engine to use the TinyMCE configuration in ui.xml

jvega190:
https://github.com/craftercms/craftercms/issues/5505

### DIFF
--- a/static-assets/components/cstudio-forms/controls/rte.js
+++ b/static-assets/components/cstudio-forms/controls/rte.js
@@ -64,7 +64,7 @@ CStudioAuthoring.Module.requireModule(
             configuration = 'generic';
 
           const { operators } = CrafterCMSNext.rxjs;
-          const { filter, take } = operators;
+          const { take } = operators;
 
           for (var i = 0; i < config.properties.length; i++) {
             var prop = config.properties[i];
@@ -78,32 +78,39 @@ CStudioAuthoring.Module.requireModule(
             }
           }
 
-          CrafterCMSNext.system.getStore().subscribe((store) => {
-            if (!Boolean(store?.getState().preview.richTextEditor)) {
-              // If textEditorConfig is not loaded
-              const unsubscribe = store.subscribe(() => {
-                if (Boolean(store?.getState().preview.richTextEditor)) {
-                  _thisControl._initializeRte(
-                    config,
-                    store.getState().preview.richTextEditor[configuration],
-                    containerEl
-                  );
-                  unsubscribe();
-                }
-              });
+          CrafterCMSNext.system
+            .getStore()
+            .pipe(take(1))
+            .subscribe((store) => {
+              if (!Boolean(store.getState().preview.richTextEditor)) {
+                // If textEditorConfig is not loaded
+                const unsubscribe = store.subscribe(() => {
+                  if (Boolean(store.getState().preview.richTextEditor)) {
+                    _thisControl._initializeRte(
+                      config,
+                      store.getState().preview.richTextEditor[configuration],
+                      containerEl
+                    );
+                    unsubscribe();
+                  }
+                });
 
-              store.dispatch({
-                type: 'INIT_RICH_TEXT_EDITOR_CONFIG',
-                payload: {
-                  configXml: CrafterCMSNext.system.store.getState().uiConfig.xml,
-                  siteId: CStudioAuthoringContext.site
-                }
-              });
-            } else {
-              // If textEditorConfig is already loaded
-              _thisControl._initializeRte(config, store.getState().preview.richTextEditor[configuration], containerEl);
-            }
-          });
+                store.dispatch({
+                  type: 'INIT_RICH_TEXT_EDITOR_CONFIG',
+                  payload: {
+                    configXml: CrafterCMSNext.system.store.getState().uiConfig.xml,
+                    siteId: CStudioAuthoringContext.site
+                  }
+                });
+              } else {
+                // If textEditorConfig is already loaded
+                _thisControl._initializeRte(
+                  config,
+                  store.getState().preview.richTextEditor[configuration],
+                  containerEl
+                );
+              }
+            });
         },
 
         /**
@@ -284,7 +291,7 @@ CStudioAuthoring.Module.requireModule(
           // https://www.tiny.cloud/docs/plugins/
           // paste plugin is hardcoded in order to enable drag and drop functionality (and avoid it being removed from
           // configuration file).
-          pluginList = [rteConfig?.tinymceOptions?.plugins, 'paste', this.autoGrow && 'autoresize']
+          pluginList = [rteConfig.tinymceOptions?.plugins, 'paste', this.autoGrow && 'autoresize']
             .filter(Boolean)
             .join(' ');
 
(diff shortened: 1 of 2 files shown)


static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -75,17 +78,36 @@ CStudioAuthoring.Module.requireModule(
...
+                configXml: CrafterCMSNext.system.store.getState().uiConfig.xml,
+                siteId: CStudioAuthoringContext.site
+              }
+            });
+            CrafterCMSNext.system
+              .getStore()
+              .pipe(
+                filter((store) => Boolean(store.getState().preview.richTextEditor)),
+                take(1)
+              )
```
rart: `getStore` only emits once when the store is created; if this call misses the RTE config initialization it will never go through; you need to subscribe to the store itself.

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -517,9 +449,33 @@ CStudioAuthoring.Module.requireModule(
...
+                'images_upload_handler', // Images upload integration with our data sources is already implemented in the control
+                'code_editor_inline', // Code editor will always be inline in forms-engine.
+                'plugins', // Considered/used above, mixed with our options
+                'toolbar_sticky', // Toolbar is configured and styled to be sticky in forms-engine
+                'relative_urls' // To avoid allowing convertion of urls to be relative to the document_base_url
+              )
+            })
           });
-          _thisControl.editor = editor;
+          // _thisControl.editor = editor;
```
rart: ?

rart (changes requested):

static-assets/components/cstudio-forms/controls/rte.js
jvega190: Removed, this was incorrectly set (editor variable was wrong) and overridden after the tinymce initialization

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -75,17 +78,32 @@ CStudioAuthoring.Module.requireModule(
...
-            'no-role-support',
-            {
-              success: function (rteConfig) {
-                _thisControl._initializeRte(config, rteConfig, containerEl);
-              },
-              failure: function () {}
-            },
-            '/form-control-config/rte/rte-config.xml'
-          );
+          CrafterCMSNext.system.getStore().subscribe((store) => {
```
rart: You should `take(1)` here

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -63,6 +63,9 @@ CStudioAuthoring.Module.requireModule(
           var _thisControl = this,
             configuration = 'generic';
 
+          const { operators } = CrafterCMSNext.rxjs;
+          const { filter, take } = operators;
```
rart:
```suggestion
          const { take } = CrafterCMSNext.rxjs.operators;
```

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -75,17 +78,32 @@ CStudioAuthoring.Module.requireModule(
...
-              },
-              failure: function () {}
-            },
-            '/form-control-config/rte/rte-config.xml'
-          );
+          CrafterCMSNext.system.getStore().subscribe((store) => {
+            if (!Boolean(store?.getState().preview.richTextEditor)) {
+              // If textEditorConfig is not loaded
+              const unsubscribe = store.subscribe(() => {
+                if (Boolean(store?.getState().preview.richTextEditor)) {
```
rart: Store is never nullish here — that's the whole point of use the `getStore`
```suggestion
            if (!Boolean(store.getState().preview.richTextEditor)) {
              // If textEditorConfig is not loaded
              const unsubscribe = store.subscribe(() => {
                if (Boolean(store.getState().preview.richTextEditor)) {
```

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -354,20 +314,13 @@ CStudioAuthoring.Module.requireModule(
               imageDatasources.includes(datasource.id)
           );
 
-          const codeEditorWrap = rteConfig.codeEditorWrap ? rteConfig.codeEditorWrap === 'true' : false;
-
           const external = {
+            ...rteConfig?.tinymceOptions?.external_plugins,
```
rart: Why would rteConfig ever be nullish? This shouldn't run prior to the config initialization.

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -376,70 +329,46 @@ CStudioAuthoring.Module.requireModule(
...
-            image_uploadtab: this.editorImageDatasources.length > 0,
-            craftercms_paste_cleanup: rteConfig.craftercmsPasteCleanup?.trim() !== 'false',
-
-            menu: {
-              tools: { title: 'Tools', items: 'tinymcespellchecker code acecode wordcount' }
-            },
-
+            contextmenu: !this.enableSpellCheck, // comes from control props (not xml config)
+            image_uploadtab: this.editorImageDatasources.length > 0, // comes from control props (not xml config)
+            craftercms_paste_cleanup: rteConfig?.tinymceOptions?.craftercmsPasteCleanup ?? true, // If doesn't exist or if true => true
```
rart: This should always be a boolean. Is the `craftercmsPasteCleanup` pre-parsed to boolean?
Also, I don't think we're applied this changes to XB's rte. Please add the `craftercms_paste_extension` and this config.

rart (changes requested):

static-assets/components/cstudio-forms/controls/rte.js
jvega190: Yeah those values are boolean.
I'll add it to XB's rte.

static-assets/components/cstudio-forms/controls/rte.js
jvega190: It's never going to be null, removing the optional chaining. This function is called after initialization so we're good with that.

ui/guest/src/controls/rte.ts
```diff
@@ -118,6 +118,7 @@ export function initTinyMCE(
     skin: window.matchMedia('(prefers-color-scheme: dark)').matches ? 'oxide-dark' : 'oxide',
     media_live_embeds: true,
     file_picker_types: 'image media',
+    craftercms_paste_cleanup: rteSetup?.tinymceOptions?.craftercms_paste_cleanup ?? true, // If doesn't exist or if true => true
```
jvega190, line 121: @rart in this case (XB's RTE) rteSetup is optional so condition is needed.

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -376,70 +336,46 @@ CStudioAuthoring.Module.requireModule(
...
             encoding: 'xml',
             relative_urls: false,
             remove_script_host: false,
             convert_urls: false,
-            readonly: _thisControl.readonly,
-            force_br_newlines: forceBRTags,
-            forced_root_block: forceRootBlockPTag,
+            readonly: _thisControl.readonly, // comes from control props (not xml config)
+            force_br_newlines: forceBRTags, // comes from control props (not xml config)
+            forced_root_block: forceRootBlockPTag, // comes from control props (not xml config)
```
rart, line 347: Anything that comes from control props should be disallowed to be overridden by ui.xml config or merged, if applicable.

rart (changes requested):
Remove "editform" from plugins config.

rart (approved):
@sumerjabri this will need a lot of coordinated and creative testing and poking by @russdanner and QA (@yacdaniel)

PR #2950, opened 2022-04-21
[5487] Some XB actions (like sort) cause components to stay locked.

jvega190:
https://github.com/craftercms/craftercms/issues/5487

### DIFF
--- a/ui/guest/src/store/epics/root.ts
+++ b/ui/guest/src/store/epics/root.ts
@@ -849,6 +849,13 @@ const moveComponent = (dragContext) => {
           containerRecord.fieldId.includes('.') ? `${containerRecord.index}.${targetIndex}` : targetIndex
         );
       });
+    } else {
+      // if draggedElementIndex equals targetIndex it means that item wasn't moved, but it was locked, so it needs
+      // to be unlocked.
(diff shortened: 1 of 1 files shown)


ui/guest/src/store/epics/root.ts
```diff
@@ -849,6 +849,13 @@ const moveComponent = (dragContext) => {
           containerRecord.fieldId.includes('.') ? `${containerRecord.index}.${targetIndex}` : targetIndex
         );
       });
+    } else {
+      // if draggedElementIndex equals targetIndex it means that item wasn't moved, but it was locked, so it needs
+      // to be unlocked.
+      const models = contentController.getCachedModels();
+      const id = dragged.modelId;
+      const path = models[id].craftercms.path;
+      post(unlockItem({ path }));
```
rart, line 858: Check this for embedded components too. You may need to re-retrieve the "parent id" to get to the path in such cases.

rart (changes requested):

ui/guest/src/store/epics/root.ts
jvega190: It works properly for both shared and embedded components

rart (approved):

PR #2969, opened 2022-04-26
[5550] There's no feedback when trying to open a folder with a lot of images in the image picker

jvega190:
https://github.com/craftercms/craftercms/issues/5550


static-assets/components/cstudio-browse/browse.js
```diff
@@ -557,9 +577,6 @@
       }
     }
 
-    $resultsContainer.empty();
```
jvega190, line 560: This was incorrect, causing the loading state to be cleared before loading content. Moved inside `setResults` function.

rart (approved):

PR #2976, opened 2022-04-28
[5567] Save & Close option is showing up in Choose Content Type window after saving and closing Content Type Modelling

jvega190:
https://github.com/craftercms/craftercms/issues/5567

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-types.js
+++ b/static-assets/components/cstudio-admin/mods/content-types.js
@@ -85,11 +85,6 @@
         $('#createNewContentTypeButton').click(() => {
           this.onNewClick();
         });
-
-        CStudioAuthoring.ContextualNav.AdminConsoleNav.initActions([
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -85,11 +85,6 @@
         $('#createNewContentTypeButton').click(() => {
           this.onNewClick();
         });
-
-        CStudioAuthoring.ContextualNav.AdminConsoleNav.initActions([
```
jvega190, line 89: Now that the contextual-nav.xml configuration is not loaded anymore, `AdminConsoleNav` module is not loaded, so this initialization is not needed (and the object doesn't exist). Checked for other usages and they're dead code (we should remove it)

rart: go ahead

rart (approved):

PR #2989, opened 2022-05-04
[5581] Dashboard issues with the contextual actions

jvega190:
https://github.com/craftercms/craftercms/issues/5581

### DIFF
--- a/ui/app/src/state/epics/content.ts
+++ b/ui/app/src/state/epics/content.ts
@@ -182,7 +182,7 @@ const content: CrafterCMSEpic[] = [
           // request (via reloadDetailedItem action)
           !state.content.itemsByPath[payload.path] || type === reloadDetailedItem.type
       ),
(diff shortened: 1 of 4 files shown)


ui/app/src/state/epics/content.ts
```diff
@@ -182,7 +182,7 @@ const content: CrafterCMSEpic[] = [
           // request (via reloadDetailedItem action)
           !state.content.itemsByPath[payload.path] || type === reloadDetailedItem.type
       ),
-      switchMap(([{ payload }, state]) =>
+      mergeMap(([{ payload }, state]) =>
```
jvega190, line 185: @rart the issue was that after publishing there was a `reloadDetailedItem` for each published item, and the switchMap was causing a cancelling of the previous reload, so the state wasn't updated properly.

rart (approved):

PR #2994, opened 2022-05-05
[5560] Change Type service and dialog should not present the existing type as an option

jvega190:
https://github.com/craftercms/craftercms/issues/5560


jvega190:
Ready. Hold in draft for 4.0.1

rart (approved):

PR #3001, opened 2022-05-11
[5552] Duplication/deletion using XB allowed event though there are min/max validations

jvega190:
https://github.com/craftercms/craftercms/issues/5552


jvega190:
Ready. Hold in draft for 4.0.1

rart (approved):

PR #3009, opened 2022-05-20
[5542] The Add Component spinner in XB should be centered vertically and horizontally

jvega190:
https://github.com/craftercms/craftercms/issues/5542


ui/guest/src/store/epics/root.ts
```diff
@@ -275,7 +275,10 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
         // so there should be no lock or other actions.
         if (dropZone) {
           const { modelId } = iceRegistry.getById(dropZone.iceId);
-          const path = models[modelId].craftercms.path;
+          // get parentModelId in case the current dropZone is an embedded component
```
jvega190, line 183: This was an issue detected while testing on bootstrap blueprint, if current dropzone is an embedded, we need to get the parentModel to lock the proper item

rart (approved):

PR #3011, opened 2022-05-24
[5604] Make the site cookie domain configurable as base or sub

jvega190:
https://github.com/craftercms/craftercms/issues/5604

### DIFF
--- a/ui/app/src/utils/auth.ts
+++ b/ui/app/src/utils/auth.ts
@@ -47,13 +47,26 @@ export function getXSRFToken(): string {
   return Cookies.get(XSRF_TOKEN_COOKIE_NAME);
 }
 
-export function getCookieDomain(): string {
-  return window.location.hostname.includes('.') ? window.location.hostname : '';
+export function getCookieDomain(useBaseDomain: boolean = true): string {
+  let hostname = window.location.hostname;
+  let domain = hostname.includes('.') ? hostname : '';
+
+  if (useBaseDomain) {
+    const ipRegex = /\b(?:[0-9]{1,3}\.){3}[0-9]{1,3}\b/gm;
+    const isHostnameIp = Boolean(hostname.match(ipRegex));
+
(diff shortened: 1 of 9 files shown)


rart (approved):

ui/app/src/utils/auth.ts
```diff
@@ -47,13 +47,26 @@ export function getXSRFToken(): string {
...
+    const ipRegex = /\b(?:[0-9]{1,3}\.){3}[0-9]{1,3}\b/gm;
+    const isHostnameIp = Boolean(hostname.match(ipRegex));
+
+    if (hostname.includes('.') && !isHostnameIp) {
+      domain = hostname.replace(/^(.*?)\./, '');
+      domain = `.${domain.includes('.') ? domain : hostname}`;
+    }
+  }
+
+  return domain;
```
rart:
```suggestion
export function getCookieDomain(useBaseDomain: boolean = false): string {
  let domainName = window.location.hostname;
  if (useBaseDomain) {
    let segments = domainName.split('.');
    return segments.length <= 2 ? domainName : segments.slice(-2).join('.');
  } else {
    return domainName;
  }
```

rart (changes requested):

ui/app/src/utils/auth.ts
```diff
@@ -47,13 +47,19 @@ export function getXSRFToken(): string {
...
+  if (useBaseDomain) {
+    let segments = domainName.split('.');
+    return segments.length <= 2 ? domainName : segments.slice(-2).join('.');
+  } else {
+    return domainName;
+  }
 }
 
-export function setSiteCookie(value: string): void {
+export function setSiteCookie(value: string, useBaseDomain: boolean = true): void {
```
rart:
```suggestion
export function setSiteCookie(value: string, useBaseDomain: boolean = false): void {
```

rart (changes requested):

rart (approved):

PR #3013, opened 2022-05-26
[5585] When a user inputs data in Project Tools and press ESC or click outside the form, the dialog closes without confirmation

jvega190:
https://github.com/craftercms/craftercms/issues/5585


static-assets/components/cstudio-dialogs/new-content-type.js
```diff
@@ -164,7 +175,10 @@ CStudioAuthoring.Dialogs.NewContentType = CStudioAuthoring.Dialogs.NewContentTyp
       objectTypeEl: document.getElementById('contentTypeObjectType')
     };
 
+    // 'Type' dropdown
     YEvent.on('contentTypeObjectType', 'change', function () {
+      self.onSetDirty(true);
```
rart, line 184: Should avoid setting this every keystroke. Set only when it changes from blank to not-blank and backwards and/or make sure react-side is also not setting state on every keystroke.

static-assets/components/cstudio-dialogs/new-content-type.js
```diff
@@ -173,13 +187,16 @@ CStudioAuthoring.Dialogs.NewContentType = CStudioAuthoring.Dialogs.NewContentTyp
...
 
+    // 'Display Label' input
     YEvent.on('contentTypeDisplayName', 'keyup', function () {
       YAHOO.Bubbling.fire('content-type.values.changed');
       value = document.getElementById('contentTypeDisplayName').value;
 
       value = value.replace(/[^a-z0-9]/gi, '');
       value = value.toLowerCase();
 
+      self.onSetDirty(Boolean(value));
```
rart, line 202: Should avoid setting this every keystroke. Set only when it changes from blank to not-blank and backwards and/or make sure react-side is also not setting state on every keystroke.

ui/app/src/components/GraphiQL/GraphiQL.tsx
```diff
@@ -39,8 +42,10 @@ function GraphiQL(props: GraphiQLProps) {
     showAppsButton,
     embedded = false,
     method = 'post',
-    url = `${window.location.origin}/api/1/site/graphql`
+    url = `${window.location.origin}/api/1/site/graphql`,
+    onSubmittingAndOrPendingChange
   } = props;
+  const [initialQuery] = useState(window.localStorage.getItem(`${storageKey}graphiql:query`));
   const [query, setQuery] = useState(() => window.localStorage.getItem(`${storageKey}graphiql:query`));
```
rart:
```suggestion
  const initialQuery = useMemo(() => window.localStorage.getItem(`${storageKey}graphiql:query`), []);
  const [query, setQuery] = useState(initialQuery);
```

ui/app/src/components/GraphiQL/GraphiQL.tsx
```diff
@@ -84,12 +89,18 @@ function GraphiQL(props: GraphiQLProps) {
       };
     }
   }, [url, method]);
+  const fnRefs = useUpdateRefs({ onSubmittingAndOrPendingChange });
+  const hasChanges = initialQuery !== query;
```
rart:
```suggestion
  const hasChanges = isBlank(query) ? false : initialQuery !== query;
```

ui/app/src/components/PublishOnDemandWidget/PublishOnDemandWidget.tsx
```diff
@@ -153,6 +157,13 @@ export function PublishOnDemandWidget(props: PublishOnDemandWidgetProps) {
     }
   };
 
+  useEffect(() => {
+    fnRefs.current.onSubmittingAndOrPendingChange?.({
+      hasPendingChanges: mode === 'studio' ? publishStudioFormValid : publishGitFormValid,
```
rart: Don't see how `valid` is suitable to identify if there are changes.

ui/app/src/components/WorkflowStateManagement/WorkflowStateManagement.tsx
```diff
@@ -123,6 +127,12 @@ export function WorkflowStateManagement(props: WorkflowStateManagementProps) {
     fetchStates();
   }, [fetchStates]);
 
+  useEffect(() => {
+    fnRefs.current.onSubmittingAndOrPendingChange?.({
+      hasPendingChanges: Boolean(selectedItemsLength)
+    });
+  }, [selectedItemsLength, fnRefs]);
```
rart: `selectedItemsLength` may change too much. All we would care about is when it changes from or to `0`. Doesn't `hasSelectedItems` give you that already?

rart (changes requested):

ui/app/src/components/PublishOnDemandWidget/PublishOnDemandWidget.tsx
jvega190: Valid is true when the form has content (so it can be published), if there is no content in the form (is blank) then valid becomes false

ui/app/src/components/PublishOnDemandWidget/PublishOnDemandWidget.tsx
```diff
@@ -118,30 +120,48 @@ const initialPublishGitFormData = {
...
   const [publishingTargetsError, setPublishingTargetsError] = useState(null);
   const [publishGitFormData, setPublishGitFormData] = useSpreadState<PublishFormData>(initialPublishGitFormData);
   const [publishGitFormValid, setPublishGitFormValid] = useState(false);
+  const publishGitFormChanged = useMemo(() => {
+    return (
+      publishGitFormData.commitIds !== initialPublishGitFormData.commitIds ||
+      publishGitFormData.comment !== initialPublishGitFormData.comment ||
+      publishGitFormData.environment !== initialPublishingTarget
+    );
+  }, [initialPublishingTarget, publishGitFormData]);
```
rart: These change on every keystroke so not worth the memo

ui/app/src/components/PublishOnDemandWidget/PublishOnDemandWidget.tsx
```diff
@@ -118,30 +120,48 @@ const initialPublishGitFormData = {
...
   const [publishStudioFormData, setPublishStudioFormData] =
     useSpreadState<PublishFormData>(initialPublishStudioFormData);
   const [publishStudioFormValid, setPublishStudioFormValid] = useState(false);
+  const publishStudioFormChanged = useMemo(() => {
+    return (
+      publishStudioFormData.path !== initialPublishStudioFormData.path ||
+      publishStudioFormData.comment !== initialPublishStudioFormData.comment ||
+      publishStudioFormData.environment !== initialPublishingTarget
+    );
+  }, [initialPublishingTarget, publishStudioFormData]);
```
rart: These change on every keystroke so not worth the memo

rart: you should check the mode and keep a single "hasChanges" that then use on the effect. Only run the calculations (without memo) for the applicable form.

rart (changes requested):

ui/app/src/components/PublishOnDemandWidget/PublishOnDemandWidget.tsx
```diff
@@ -153,6 +167,13 @@ export function PublishOnDemandWidget(props: PublishOnDemandWidgetProps) {
     }
   };
 
+  useEffect(() => {
+    fnRefs.current.onSubmittingAndOrPendingChange?.({
+      hasPendingChanges: hasChanges,
+      isSubmitting
+    });
+  }, [isSubmitting, mode, hasChanges, fnRefs]);
```
rart:
```suggestion
  }, [isSubmitting, hasChanges, fnRefs]);
```

rart (changes requested):

rart (approved):

PR #3014, opened 2022-05-26
[5569] Add option to image picker to disallow upload of SVG

jvega190:
https://github.com/craftercms/craftercms/issues/5569

### DIFF
--- a/static-assets/components/cstudio-forms/controls/image-picker.js
+++ b/static-assets/components/cstudio-forms/controls/image-picker.js
@@ -58,6 +58,8 @@ CStudioForms.Controls.ImagePicker =
     this.previewBoxWidth = 300;
     this.external = null;
     this.supportedPostFixes = ['_s'];
+    this.CMgs = CStudioAuthoring.Messages;
+    this.langBundle = CStudioAuthoring.Messages.getBundle('contentTypes', CStudioAuthoringContext.lang);
 
     return this;
   };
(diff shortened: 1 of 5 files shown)


static-assets/components/cstudio-forms/controls/image-picker.js
```diff
@@ -871,7 +877,8 @@ YAHOO.extend(CStudioForms.Controls.ImagePicker, CStudioForms.CStudioFormField, {
         name: 'imageManager',
         type: 'datasource:image'
       },
-      { label: CMgs.format(langBundle, 'readonly'), name: 'readonly', type: 'boolean' }
+      { label: CMgs.format(langBundle, 'readonly'), name: 'readonly', type: 'boolean' },
+      { label: CMgs.format(langBundle, 'allowSVG'), name: 'allowSVG', type: 'boolean', defaultValue: 'true' }
```
rart: Disallow by default. Should the defaultValue here be a string or a boolean?
```suggestion
      { label: CMgs.format(langBundle, 'allowSVG'), name: 'allowSVG', type: 'boolean', defaultValue: 'false' }
```

rart (changes requested):

static-assets/components/cstudio-forms/controls/image-picker.js
jvega190: Even though the value is type boolean it is set as a string and the content-type engine handles/parses it as a boolean

rart (approved):

PR #3015, opened 2022-05-27
[4077] Add publisher review workflow option

jvega190:
https://github.com/craftercms/craftercms/issues/4077

### DIFF
--- a/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
+++ b/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
@@ -26,7 +26,6 @@ import {
   PublishDialogResourceInput
 } from './utils';
 import { useActiveSiteId } from '../../hooks/useActiveSiteId';
-import { usePermissionsBySite } from '../../hooks/usePermissionsBySite';
 import { useDispatch } from 'react-redux';
 import { fetchPublishingTargets } from '../../services/publishing';
 import { getComputedPublishingTarget, getDateScheduled } from '../../utils/content';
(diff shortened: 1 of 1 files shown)


ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -76,9 +75,7 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
   const [submitDisabled, setSubmitDisabled] = useState(true);
 
   const siteId = useActiveSiteId();
-  const permissionsBySite = usePermissionsBySite();
-  const myPermissions = permissionsBySite[siteId];
-  const hasPublishPermission = myPermissions?.includes('publish');
+  const hasPublishPermission = items?.every((item) => item.availableActionsMap.publish);
```
rart: This runs on every render; `some` would be best: doesn't iterate through the whole array and doesn't return true with an empty array
```suggestion
  const hasPublishPermission = items?.some((item) => !item.availableActionsMap.publish);
```

rart (changes requested):

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
jvega190: It should be with a `!` at he beggining: `!items?.some((item) => !item.availableActionsMap.publish)` (if there are some items without publish permission => negate to get if it has publish permission)

rart (approved):

PR #3029, opened 2022-06-02
[5614][3.1.x] TinyMCE menu unselectable after adding a new repeat group item

jvega190:
https://github.com/craftercms/craftercms/issues/5614

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -2114,9 +2114,19 @@ var CStudioForms =
             });
           };
 
+          // If the repeating group that it's being re-rendered has RTE5 controls, remove all of them before cleanup up
+          // markup (so it'll remove all things related to those RTEs)
+          const $rte5Controls = $(controlEl).find('.rte-tinymce5-control')
+          if ($rte5Controls.length) {
+            const $rteInputs = $rte5Controls.find('.cstudio-form-control-input');
+            $rteInputs.each((index, element) => {
+              const rteId = element.getAttribute('id');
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2114,9 +2114,19 @@ var CStudioForms =
             });
           };
 
+          // If the repeating group that it's being re-rendered has RTE5 controls, remove all of them before cleanup up
```
jvega190, line 2117: @rart the issue was that when re-rendering the repeating groups all elements `.tox-silver-sink` were being removed (those are related with RTE menus), and those not related with the repeating group being re-rendered were also removed, causing initialized RTEs to miss the elements and not to behave properly.
Using the `.remove()` function of the RTE removes everything related with the editor being removed.

static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2182,6 +2192,7 @@ var CStudioForms =
             form.model[repeat.id][0] = [];
 
             this.parentNode.parentNode.reRender(this.parentNode.parentNode);
+            repeatEdited = true;
```
rart, line 2195: What's this for? Don't see it used anywhere in this PR.

jvega190: That was missing in the 'Add first item' case for the repeating group to indicate that there was a change in the form (and for it to show the confirmation dialog when trying to close it after adding a first item in a repeating group.

rart (approved):

PR #3032, opened 2022-06-06
[5615][3.1.x] Remove Sandbox Branch from the site creation dialog

jvega190:
https://github.com/craftercms/craftercms/issues/5615

### DIFF
--- a/ui/app/src/components/BlueprintForm.tsx
+++ b/ui/app/src/components/BlueprintForm.tsx
@@ -87,9 +87,9 @@ const messages = defineMessages({
     id: 'createSiteDialog.cantStart',
     defaultMessage: 'Site names may not start with zeros, dashes (-) or underscores (_).'
   },
-  sandboxBranch: {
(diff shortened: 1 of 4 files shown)


ui/app/src/components/BlueprintForm.tsx
```diff
@@ -87,9 +87,9 @@ const messages = defineMessages({
     id: 'createSiteDialog.cantStart',
     defaultMessage: 'Site names may not start with zeros, dashes (-) or underscores (_).'
   },
-  sandboxBranch: {
-    id: 'createSiteDialog.sandboxBranch',
-    defaultMessage: 'Sandbox Branch'
+  gitBranch: {
+    id: 'createSiteDialog.gitBranch',
+    defaultMessage: 'Git Branch'
```
rart, line 92: There's an existing message with this same string. Use that key so we don't need to add new translations.

rart (changes requested):
See other changes in the PR for 4.

rart (approved):

PR #3046, opened 2022-06-17
[5597] Opening Dependencies -> History on an image or template without changes, shows blank page in UI

jvega190:
https://github.com/craftercms/craftercms/issues/5597

### DIFF
--- a/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
+++ b/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
@@ -57,12 +57,14 @@ import DialogFooter from '../DialogFooter/DialogFooter';
 import { HistoryDialogPagination } from './HistoryDialogPagination';
 import { historyStyles } from './HistoryDialog';
 import useSelection from '../../hooks/useSelection';
+import useFetchSandboxItems from '../../hooks/useFetchSandboxItems';
 
 export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
   const { versionsBranch } = props;
   const { count, page, limit, current, rootPath, isConfig } = versionsBranch;
+  useFetchSandboxItems([versionsBranch.item.path]);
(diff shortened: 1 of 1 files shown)


ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -57,12 +57,14 @@ import DialogFooter from '../DialogFooter/DialogFooter';
...
 import { historyStyles } from './HistoryDialog';
 import useSelection from '../../hooks/useSelection';
+import useFetchSandboxItems from '../../hooks/useFetchSandboxItems';
 
 export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
   const { versionsBranch } = props;
   const { count, page, limit, current, rootPath, isConfig } = versionsBranch;
+  useFetchSandboxItems([versionsBranch.item.path]);
   // TODO: It'd be best for the dialog to directly receive a live item. Must change versions branch to only hold the path.
-  const item = useSelection((state) => state.content.itemsByPath[versionsBranch.item.path]) ?? versionsBranch.item;
```
jvega190, line 65: versionsBranch.item is a parsedLegacyItem, so it doesn't have available actions, causing the component to fail. So I'm using the `useFetchSandboxItems` to fetch if it is not in state.

rart: how are parsedLegacyItems getting to the content state? we should avoid that and do something similar to what was done on the dashboards: upon getting the legacy items, retrieving the sandbox items. then, push that on to state

jvega190: In this case the parsedLegacyItem wasn't getting to the content state, it was being used directly in the component. The item was being retrieved from the state and if it wasn't there it was using the one from the versions service. So that's what I did, using useFetchSandboxItems to retrieve and push to state, then use only from state.

rart (approved):

PR #3051, opened 2022-06-20
[5646] UI error when inputting data with keyboard inside Audit Logs -> Timestamp Filter

jvega190:
https://github.com/craftercms/craftercms/issues/5646

### DIFF
--- a/ui/app/src/components/AuditGridFilterPopover/AuditGridFilterPopoverBody.tsx
+++ b/ui/app/src/components/AuditGridFilterPopover/AuditGridFilterPopoverBody.tsx
@@ -17,10 +17,8 @@
 import useStyles from './styles';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import React, { ChangeEvent, useCallback, useMemo, useState } from 'react';
-import Avatar from '@mui/material/Avatar';
 import ClearRoundedIcon from '@mui/icons-material/ClearRounded';
-import DatePicker from '@mui/lab/DatePicker';
-import TimePicker from '@mui/lab/TimePicker';
+import DateTimePicker from '@mui/lab/DateTimePicker';
 import AdapterDateFns from '@mui/lab/AdapterDateFns';
 import LocalizationProvider from '@mui/lab/LocalizationProvider';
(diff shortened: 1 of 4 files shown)


ui/app/src/components/AuditGridFilterPopover/AuditGridFilterPopoverBody.tsx
```diff
@@ -179,15 +179,19 @@ export function AuditGridFilterPopoverBody(props: AuditGridFilterPopoverProps) {
                 inputFormat="MM/dd/yyyy"
                 value={fromDate}
                 onChange={onFromDateSelected}
-                renderInput={(props) => <TextField {...props} />}
+                renderInput={(props) => (
+                  <TextField {...props} inputProps={{ ...props.inputProps, onKeyDown: (e) => e.preventDefault() }} />
+                )}
               />
               <TimePicker
```
jvega190: @rart shouldn't this be a datePicker too?

rart (changes requested):
Please apply changes as discussed during our call.

ui/app/src/components/AuditGridFilterPopover/styles.ts
```diff
@@ -18,26 +18,18 @@ import { makeStyles } from 'tss-react/mui';
...
-    top: '-20px',
-    right: '-20px',
-    cursor: 'pointer',
-    '&:hover': {
-      opacity: '0.9'
+  timestampFiltersContainer: {
+    display: 'flex',
+    alignItems: 'center',
+    marginBottom: theme.spacing(2),
+    '& .MuiTextField-root:first-child': {
```
jvega190: There seems to be an issue in the `DateTime/date/time` pickers and the className doesn't apply (neither under PaperProps, or classes)

rart: Use the field classes export

rart: https://mui.com/x/react-date-pickers/getting-started/#MaterialUIPickers.js

rart (approved):

PR #3067, opened 2022-06-24
[5654] Studio search result is not displayed in correct sort order of response data

jvega190:
https://github.com/craftercms/craftercms/issues/5654

### DIFF
--- a/static-assets/components/cstudio-search/search.js
+++ b/static-assets/components/cstudio-search/search.js
@@ -525,6 +525,9 @@
       ? CStudioSearch.typesMap[result.type].icon
       : CStudioSearch.typesMap['Other'].icon;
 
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-search/search.js
```diff
@@ -525,6 +525,9 @@
       ? CStudioSearch.typesMap[result.type].icon
       : CStudioSearch.typesMap['Other'].icon;
 
+    const $resultContainer = $('<div class="result-container"/>')
```
jvega190, line 528: @rart the issue was that there's an async call before rendering the result card, so I'm rendering first the container (to put it in the correct position) then render the rest of the card

rart (approved):

russdanner:
Confirmed

PR #3068, opened 2022-06-24
[5653] Certain file types not being handled by UI, empty screen when opening them from Item History -> View

jvega190:
https://github.com/craftercms/craftercms/issues/5653

### DIFF
--- a/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
+++ b/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
@@ -58,6 +58,7 @@ import { HistoryDialogPagination } from './HistoryDialogPagination';
 import { historyStyles } from './HistoryDialog';
 import useSelection from '../../hooks/useSelection';
 import useFetchSandboxItems from '../../hooks/useFetchSandboxItems';
+import { UNDEFINED } from '../../utils/constants';
 
 export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
   const { versionsBranch } = props;
@@ -165,7 +166,7 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
             type: image ? 'image' : video ? 'video' : 'editor',
             title: item.label,
             [image || video ? 'url' : 'content']: content,
-            mode: image || video ? void 0 : getEditorMode(item),
+            mode: image || video ? UNDEFINED : getEditorMode(item),
             subtitle: `v.${version.versionNumber}`,
             ...(video ? { mimeType: item.mimeType } : {})
           })
--- a/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
+++ b/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
@@ -123,6 +123,7 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
     },
     [item, count, setMenu, formatMessage, isConfig]
   );
+  const hasMenuOptions = isPreviewable(item) || count > 1;
 
   const compareVersionDialogWithActions = () =>
     showCompareVersionsDialog({
@@ -315,7 +316,7 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
         <SuspenseWithEmptyState resource={versionsResource}>
           <VersionList
             versions={versionsResource}
-            onOpenMenu={handleOpenMenu}
+            onOpenMenu={hasMenuOptions ? handleOpenMenu : null}
             onItemClick={handleViewItem}
             current={current}
           />
--- a/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
+++ b/ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
@@ -37,7 +37,7 @@ import translations from './translations';
 import { batchActions } from '../../state/actions/misc';
 import { fetchContentTypes } from '../../state/actions/preview';
 import { fetchContentByCommitId } from '../../services/content';
-import { getEditorMode, isImage } from '../PathNavigator/utils';
+import { getEditorMode, isImage, isPreviewable } from '../PathNavigator/utils';
 import {
   compareBothVersions,
   compareToPreviousVersion,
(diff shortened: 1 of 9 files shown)


ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -155,15 +155,30 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
...
-          showPreviewDialog({
-            type: image ? 'image' : 'editor',
-            title: item.label,
-            [image ? 'url' : 'content']: content,
-            mode: image ? void 0 : getEditorMode(item),
-            subtitle: `v.${version.versionNumber}`
-          })
-        );
+
+        if (content instanceof Blob) {
```
jvega190, line 159: I though it would be good to have this validation here (even though I added fonts to the fetchContentByCommitId service), just in case there's a non-handled Blob.

rart (changes requested):
Let's create and implement a whitelist of previewable mime types and only show the "view" option for those that we know we can preview.

Make sure to update the `isPreviewable` so that non-previewables don't show as links on the sidebar.

Logic between available actions and the history dialog view should somehow be shared.

Add into a site of yours files of all of our typical files that we know we can preview to examine mime types for our whitelist (js, ts, css, html, txt, json, etc, be creative). We should whitelist `text/*`, `application/json`, `application/x-javascript`, etc..

ui/app/src/components/PathNavigator/utils.ts
```diff
@@ -37,8 +37,20 @@ export function isVideo(item: DetailedItem | SandboxItem): boolean {
...
+    /^image\//.test(mimeType)
+  );
+}
+
 export function isPreviewable(item: DetailedItem | SandboxItem): boolean {
-  return ['page', 'component', 'asset', 'renderingTemplate', 'script', 'taxonomy'].includes(item?.systemType);
+  if (item?.systemType === 'asset') {
+    return isAssetPreviewable(item.mimeType);
+  } else {
+    return ['page', 'component', 'asset', 'renderingTemplate', 'script', 'taxonomy'].includes(item?.systemType);
```
rart:
```suggestion
    return ['page', 'component', 'renderingTemplate', 'script', 'taxonomy'].includes(item?.systemType);
```

ui/app/src/components/PathNavigator/utils.ts
```diff
@@ -37,8 +37,20 @@ export function isVideo(item: DetailedItem | SandboxItem): boolean {
   return item?.mimeType.startsWith('video/');
 }
 
+export function isAssetPreviewable(mimeType: string): boolean {
```
rart: Separate text from binary (isTextContent, isMediaContent)

ui/app/src/services/content.ts
```diff
@@ -1361,13 +1362,10 @@ export function fetchContentByCommitId(site: string, path: string, commitId: str
...
       const type = ajax.xhr.getResponseHeader('content-type');
-      if (
-        /^text\//.test(type) ||
-        /^application\/(x-httpd-php|rtf|xhtml\+xml|xml|json|ld\+json|javascript|x-groovy|x-sh)$/.test(type)
-      ) {
-        return blob.text() as Promise<string>;
-      } else if (/^image\//.test(type)) {
+      if (/^image\//.test(type)) {
         return of(URL.createObjectURL(blob));
+      } else if (isAssetPreviewable(type)) {
```
rart: Use isTextContent

ui/app/src/utils/itemActions.ts
```diff
@@ -358,10 +358,12 @@ export function generateSingleItemOptions(
   if (hasReadAction(item.availableActions) && actionsToInclude.view) {
     if (['page', 'component', 'taxonomy', 'levelDescriptor'].includes(type)) {
       sectionA.push(menuOptions.view);
-    } else if (isImage) {
-      sectionA.push(menuOptions.viewImage);
-    } else {
-      sectionA.push(menuOptions.viewCode);
+    } else if (isPreviewable(item)) {
+      if (isImage) {
+        sectionA.push(menuOptions.viewImage);
```
rart: Test and handle video everywhere

rart (changes requested):

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -118,8 +121,9 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
         activeItem: version
       });
     },
-    [item?.systemType, item?.availableActionsMap.revert, count, setMenu, formatMessage, isConfig, item?.stateMap.locked]
+    [item, count, setMenu, formatMessage, isConfig]
```
rart: The item object can change frequently. Why not keep the individual properties instead of the instance as dependencies of the callback?

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -118,8 +121,9 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
...
   );
+  const hasMenuOptions = isPreviewable(item) || count > 1;
```
rart: Why would there be options if it's not previewable? Why does it not suffice to check the count?

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -152,16 +156,18 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
...
         dispatch(
           showPreviewDialog({
-            type: image ? 'image' : 'editor',
+            type: image ? 'image' : video ? 'video' : 'editor',
             title: item.label,
-            [image ? 'url' : 'content']: content,
-            mode: image ? void 0 : getEditorMode(item),
-            subtitle: `v.${version.versionNumber}`
+            [image || video ? 'url' : 'content']: content,
+            mode: image || video ? void 0 : getEditorMode(item),
```
rart:
```suggestion
            mode: image || video ? UNDEFINED : getEditorMode(item),
```

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
jvega190: There may be options like revert, that don't depend on the item being previewable or not

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
jvega190: I need the whole object a couple of lines before when I do `isPreviewable(item)`

rart (approved):

PR #3071, opened 2022-06-29
[5658] Insert component via XB drag-drop with & in the default value of a field fails

jvega190:
- https://github.com/craftercms/craftercms/issues/5658
- https://github.com/craftercms/craftercms/issues/5661

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -1497,7 +1497,7 @@ var CStudioForms =
               const saveContent = () => {
                 CrafterCMSNext.util.ajax.post(CStudioAuthoring.Service.createServiceUri(serviceUrl), xml).subscribe(
                   function () {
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -1497,7 +1497,7 @@ var CStudioForms =
               const saveContent = () => {
                 CrafterCMSNext.util.ajax.post(CStudioAuthoring.Service.createServiceUri(serviceUrl), xml).subscribe(
                   function () {
-                    CStudioForms.currentValidFolder = CStudioForms.updatedModel['folder-name'];
+                    CStudioForms.currentValidFolder = CStudioForms.updatedModel?.['folder-name'];
```
jvega190, line 1500: While testing this fix I found an exception in here (updateModel may not exist)

rart (approved):

PR #3085, opened 2022-07-11
[5679] Improve dark mode border treatment on image picker

jvega190:
https://github.com/craftercms/craftercms/issues/5679


static-assets/themes/cstudioTheme/css/forms-default.css
```diff
@@ -1044,7 +1044,7 @@ div.yui-calcontainer.cstudio-form-field-calendar {
 }
 
 .cstudio-form-control-asset-picker-preview-content {
-  height: 250px;
+  height: 248px;
```
jvega190, line 1047: Image container was the same size as the image, so it was overlapping. I reduced it 2px to leave space for the border (that is 1px)

rart (approved):

PR #3092, opened 2022-07-15
[5500] Adding a Smile Feature Item from Browse Components, causes issues with XB

jvega190:
https://github.com/craftercms/craftercms/issues/5500

### DIFF
--- a/ui/guest/src/iceRegistry.ts
+++ b/ui/guest/src/iceRegistry.ts
@@ -216,10 +216,10 @@ export function getRecordDropTargets(id: number): ICERecord[] {
 
     flattenChildren(id, allChildren);
 
-    return getContentTypeDropTargets(contentType).filter((rec) => {
-      // Not the current item nor a descendant of it (i.e. can't
+    return getContentTypeDropTargets(contentType, (rec) => {
+      // Exclude if it's the current item or a descendant of it (i.e. can't
       // move an item deeper inside itself).
-      return rec.modelId !== id && !allChildren.includes(rec.modelId);
+      return !(rec.modelId !== id && !allChildren.includes(rec.modelId));
     });
   } else if (field.type === 'repeat') {
     return getRepeatGroupItemDropTargets(record);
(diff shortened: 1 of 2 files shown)


ui/guest/src/store/reducers/root.ts
```diff
@@ -110,6 +111,14 @@ const resetState: (state: GuestState) => GuestState = (state: GuestState) => ({
   dragContext: null
 });
 
+const filterDropZones = (dropZones: DropZone[], instance) => {
+  return dropZones.filter((dropZone) => {
+    const instanceId = instance.craftercms.id;
+    const dropZoneId = iceRegistry.getById(dropZone.iceId).modelId;
+    return !contentController.modelHierarchyMap[dropZoneId].children.includes(instanceId);
+  });
+};
```
rart: We already have some reducers which aren't pure but we don't want to keep adding to them. These should be done earlier, before dispatching the action; whether it is at the epic or handler level.

rart (changes requested):

rart (changes requested):
Ping to discuss

ui/guest/src/iceRegistry.ts
```diff
@@ -215,10 +216,10 @@ export function getRecordDropTargets(id: number): ICERecord[] {
...
     flattenChildren(id, allChildren);
 
-    return getContentTypeDropTargets(contentType).filter((rec) => {
-      // Not the current item nor a descendant of it (i.e. can't
-      // move a item deeper inside itself).
-      return rec.modelId !== id && !allChildren.includes(rec.modelId);
+    return getContentTypeDropTargets(contentType, (rec) => {
+      // Exclude if it's the current item or a descendant of it (i.e. can't
+      // move an item deeper inside itself).
+      return !(rec.modelId !== id && !allChildren.includes(rec.modelId));
```
```suggestion
      return rec.modelId === id || allChildren.includes(rec.modelId);
```

rart (changes requested):

rart (approved):

PR #3099, opened 2022-07-19
[5556] Studio Search and Search Dialogs have Path filter that limits to current path only

jvega190:
Added lockBasePath to forbid the basePath to be changed (sub folders can be selected in the path selector)

craftercms/craftercms#5556

### DIFF
--- a/ui/app/src/components/SiteSearchPathSelector/PathSelector.tsx
+++ b/ui/app/src/components/SiteSearchPathSelector/PathSelector.tsx
@@ -151,8 +151,8 @@ export function PathSelector(props: PathSelectorProps) {
   const onClean = (e) => {
     e.stopPropagation();
     e.preventDefault();
-    setPath('');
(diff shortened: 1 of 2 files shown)


ui/app/src/components/SiteSearchPathSelector/PathSelector.tsx
```diff
@@ -232,7 +238,7 @@ export function PathSelector(props: PathSelectorProps) {
             placeholder={formatMessage(messages.searchIn)}
             startAdornment={!disabled ? <SearchIcon className={classes.pathSelectorSearchIcon} /> : null}
             endAdornment={
-              !disabled && value ? (
+              !disabled && !lockBasePath && value ? (
```
rart: One should be able to dismiss the filtering

rart (changes requested):

PR #3108, opened 2022-07-27
Add cstudio-rte inlinepopups window.css to fix console error

phuongnq:
https://github.com/craftercms/craftercms/issues/5702


rart (commented):
@phuongnq who's trying to load this file? Shouldn't the fix be at that level, stopping the file from trying to get loaded instead of adding a dummy file?

phuongnq:
Hi @rart 
Yes, that's a good point.
The situation is that we have both `inlinepopups` and `cs_inlinepopups` plugins for RTE2.

* `inlinepopup` is the default plugin without any modification
* `cs_inlinepopups` is the one modified by Crafter

In both, there is a condition to load `window.css` from a skin:

```
init : function(ed, url) {
			// Replace window manager
			ed.onBeforeRenderUI.add(function() {
				ed.windowManager = new tinymce2.InlineWindowManager(ed);
				DOM.loadCSS(url + '/skins/' + (ed.settings.inlinepopups_skin || 'clearlooks2') + "/window.css");
			});
		},
```

As you can see, by default, it has `clearlooks2` skin, but in our case, we want to use `cstudio-rte` skin. In our customized version, we have already loaded this skin `cstudio-rte/window.css` so there is no need for the original plugin to load the same file. That's why I created an empty one to clear the error in console. Updating the original plugin is also possible, but not sure if we want to do it. Other consideration is to remove the original plugin but we may still have some functionality which is needed.

rart (approved):

PR #3109, opened 2022-07-27
XB stability update

jvega190:
- Retrieve parent model (recursive) when the current model is embedded to lock proper item
- Update all matching elements of an updated field instead of only first match
- Fetch level-descriptors

### DIFF
--- a/ui/guest/src/react/GuestProxy.tsx
+++ b/ui/guest/src/react/GuestProxy.tsx
@@ -419,9 +419,9 @@ export function GuestProxy() {
           if (fieldType === 'image') {
             const tagName = updatedField.eq(index).prop('tagName').toLowerCase();
             if (tagName === 'img') {
-              updatedField.eq(index).attr('src', value);
+              updatedField.attr('src', value);
             } else {
-              updatedField.eq(index).css('background-image', `url(${value})`);
+              updatedField.css('background-image', `url(${value})`);
             }
           } else if (fieldType === 'video-picker') {
             updatedField.eq(index).find('source').attr('src', value);
--- a/ui/guest/src/react/GuestProxy.tsx
+++ b/ui/guest/src/react/GuestProxy.tsx
@@ -408,24 +408,31 @@ export function GuestProxy() {
           break;
         }
         case updateFieldValueOperation.type:
-          const { modelId, fieldId, index = 0, value } = op.args;
-          const updatedField: JQuery<any> = $(
-            `[data-craftercms-model-id="${modelId}"][data-craftercms-field-id="${fieldId}"]`
+          const { modelId, fieldId, index, value } = op.args;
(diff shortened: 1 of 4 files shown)


ui/guest/src/store/util.ts
```diff
@@ -72,7 +72,7 @@ export function beforeWrite$<T extends any = 'continue', S extends any = never>(
           tap(({ payload }) => payload.type !== 'continue' && post(unlockItem({ path }))),
           switchMap(({ payload }) => (payload.type === 'continue' ? continue$ : stop$))
         );
-      } else if (item.commitId !== localItem.commitId && item.lockOwner !== username) {
+      } else if (item.commitId !== localItem?.commitId && item.lockOwner !== username) {
```
rart: Do we actually want an out of sync if the local item is null? Doesn't sound right.

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -419,9 +419,9 @@ export function GuestProxy() {
           if (fieldType === 'image') {
             const tagName = updatedField.eq(index).prop('tagName').toLowerCase();
             if (tagName === 'img') {
-              updatedField.eq(index).attr('src', value);
```
rart, line 422: What was going on here?
Please retest this in other places too.

rart (changes requested):

ui/guest/src/react/GuestProxy.tsx
jvega190: updatedField is a results of a query of all elements matching a modelId+fieldId, and it was only updating the src of one element (at `index`). So if we have one component in multiple parts (for example wordify, in the slider and in the latest content cards) it won't update all of the instances

rart: The index needs to be **considered** for repeat-group-nested fields/item.

rart (changes requested):

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -408,24 +408,31 @@ export function GuestProxy() {
...
+              updatedField.attr('src', value);
             } else {
-              updatedField.eq(index).css('background-image', `url(${value})`);
+              updatedField.css('background-image', `url(${value})`);
             }
           } else if (fieldType === 'video-picker') {
-            updatedField.eq(index).find('source').attr('src', value);
-            updatedField.eq(index)[0].load();
+            updatedField.find('source').attr('src', value);
+            $(updatedField).each((index, element) => {
```
rart:
```suggestion
            updatedField.each((index, element) => {
```

rart (changes requested):

rart (approved):

PR #3111, opened 2022-07-28
[5696] Add tool tip or similar visual cue with path to item in item selector form control

jvega190:
https://github.com/craftercms/craftercms/issues/5696


ui/app/src/env/i18n-legacy.ts
```diff
@@ -735,6 +735,10 @@ export const formEngineMessages = defineMessages({
   edit: {
     id: 'words.edit',
     defaultMessage: 'Edit'
+  },
+  embeddedComponent: {
+    id: 'formEngine.embeddedComponent',
```
rart:
```suggestion
    id: 'common.embeddedComponent',
```

rart (approved):

PR #3129, opened 2022-08-25
[5733] Renaming items or editing Page URL via UI, returns HTTP 404 on /content/item_by_path call

jvega190:
https://github.com/craftercms/craftercms/issues/5733

### DIFF
--- a/ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
+++ b/ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
@@ -18,10 +18,9 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import { useActiveSiteId } from '../../hooks/useActiveSiteId';
 import { FormattedMessage, useIntl } from 'react-intl';
-import { useDetailedItem } from '../../hooks/useDetailedItem';
 import { DetailedItem, SandboxItem } from '../../models/Item';
 import { getParentPath, getRootPath, withoutIndex } from '../../utils/path';
-import { createFolder, fetchSandboxItem, renameFolder } from '../../services/content';
+import { createFolder, fetchDetailedItem, fetchSandboxItem, renameFolder } from '../../services/content';
 import { batchActions } from '../../state/actions/misc';
 import { updateCreateFolderDialog } from '../../state/actions/dialogs';
 import { showErrorDialog } from '../../state/reducers/dialogs/error';
@@ -51,7 +50,8 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
   const dispatch = useDispatch();
   const site = useActiveSiteId();
   const { formatMessage } = useIntl();
-  const item = useDetailedItem(props.path);
+  const siteId = useActiveSiteId();
+  const [item, setItem] = useState(null);
   const [openSelector, setOpenSelector] = useState(false);
   const [selectedItem, setSelectedItem] = useState<DetailedItem>(null);
   const path = useMemo(() => {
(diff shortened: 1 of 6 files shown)


ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -64,6 +64,12 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
     : itemLookupTable[newFolderPath] !== UNDEFINED;
   const isValid = !isBlank(name) && !folderExists && (!rename || name !== value);
 
+  // If folder name changes, useDetailedItem will try to re-fetch a non-existing item (old path), so we will only re-fetch
+  // if path changes
```
rart: This comment is a bit confusing. If the folder name changes the path changes so whether it is useDetailed or this effect, they would both react to that no? What was the idea here? The comment needs to state 1. not to use useDetailedItem and why 2. why/how the effect solves.

ui/app/src/components/LegacyFormDialog/EmbeddedLegacyContainer.tsx
```diff
@@ -79,9 +80,16 @@ export const EmbeddedLegacyContainer = React.forwardRef(function EmbeddedLegacyE
   const iframeRef = useRef(null);
   const dispatch = useDispatch();
   const [error, setError] = useState<ApiResponse>(null);
-  const item = useDetailedItem(path);
+  const [item, setItem] = useState(null);
+  const siteId = useActiveSiteId();
   const availableActions = item?.availableActions;
 
+  // If filename changes, useDetailedItem will try to re-fetch a non-existing item (old path), so we will only re-fetch
+  // if path changes
```
rart: As above. 

Also, is it worth creating a hook for both of these instances?

rart (changes requested):

jvega190:
@rart I updated the comments, and created a hook called useDetailedItemNoState

ui/app/src/hooks/useDetailedItemNoState.ts
```diff
@@ -0,0 +1,31 @@
...
+export function useDetailedItemNoState(path: string): DetailedItem {
+  const [item, setItem] = useState(null);
+  const siteId = useActiveSiteId();
+
+  useEffect(() => {
+    fetchDetailedItem(siteId, path).subscribe(setItem);
+  }, [siteId, path]);
+
+  return item;
+}
```
rart:
```suggestion
}

export default useDetailedItemNoState;

```

rart (changes requested):
Do we actually need a **detailed** item here? It's a bit more expensive to fetch so if Sandbox does, we should go for that.

As far as the new hook, lets try something like....
```ts
export function useFetchItem(path: string, detailed: boolean = false): DetailedItem {
   const [item, setItem] = useState(null);
   useEffect(() => {
     if (detailed) 
       fetchDetailedItem(siteId, path).subscribe(setItem);
      else
       fetchSandboxItem(sideId, path, { caseAsDetailedItem: true }).subscribe(setItem)
   }, [siteId, path, detailed]);
   return item;
 }
```

ui/app/src/components/RenameAssetDialog/RenameAssetDialogContainer.tsx
```diff
@@ -144,7 +144,7 @@ export function RenameAssetDialogContainer(props: RenameAssetContainerProps) {
         <form
           onSubmit={(e) => {
             e.preventDefault();
-            if (renameDisabled) {
+            if (!renameDisabled) {
```
jvega190, line 147: Pushed this as a small fix (issue preventing submit on enter)

ui/app/src/hooks/useFetchItem.ts
```diff
@@ -0,0 +1,33 @@
...
+  const siteId = useActiveSiteId();
+  useEffect(() => {
+    if (detailed) {
+      fetchDetailedItem(siteId, path).subscribe(setItem);
+    } else {
+      fetchSandboxItem(siteId, path, { castAsDetailedItem: true }).subscribe(setItem);
+    }
+  }, [siteId, path, detailed]);
+  return item;
+}
```
rart, line 33:
```suggestion
}

export default useFetchItem;

```

ui/app/src/hooks/useFetchItem.ts
```diff
@@ -0,0 +1,36 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect, useState } from 'react';
+import useActiveSiteId from './useActiveSiteId';
+import { fetchDetailedItem, fetchSandboxItem } from '../services/content';
+import { DetailedItem } from '../models';
```
rart:
```suggestion
import { DetailedItem } from '../models/Item';
```

rart (approved):

PR #3130, opened 2022-08-25
[5724] Remove/consolidate marketplace tabs on create site dialog

jvega190:
https://github.com/craftercms/craftercms/issues/5724

### DIFF
--- a/ui/app/src/components/CreateSiteDialog/CreateSiteDialog.tsx
+++ b/ui/app/src/components/CreateSiteDialog/CreateSiteDialog.tsx
@@ -64,9 +64,9 @@ const messages = defineMessages({
     id: 'common.marketplace',
     defaultMessage: 'Marketplace'
   },
-  publicMarketplace: {
-    id: 'createSiteDialog.publicMarketplace',
-    defaultMessage: 'Public Marketplace'
+  publicMarketplaceBlueprints: {
+    id: 'createSiteDialog.publicMarketplaceBlueprints',
+    defaultMessage: 'Public Marketplace Blueprints'
   },
   back: {
     id: 'common.back',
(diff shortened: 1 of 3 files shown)


rart (changes requested):
See comment on ticket.

rart (changes requested):
- Switching the show incompatible bps check box on and off momentarily shows the no bps found state. When loading there should be a loading state instead. Same thing applies for the initial load.
- Let's add a divider above the "Marketplace" title
- Let's try overline text styles for the "Maketplace" title. Or table header styles.
- Spacing below the "marketplace" title whilst the search is not visible is too large
- The dialog header spacing is messed up (vertically). Is it using our DialogHeader component? Does it have strange/unnecessary overrides?
- The dialog body also behaves strangely. The scroll begins around 20px below where the body starts. Please check. Make sure the other views don't get messed up.
- The search button doesn't seem to be an actual button. It should be an IconButton
- Backdrop click of the dialog doesn't work

rart (changes requested):
Now, the second line of the header is misaligned.

ui/app/src/components/CreateSiteDialog/CreateSiteDialog.tsx
```diff
@@ -57,6 +54,10 @@ import { useEnv } from '../../hooks/useEnv';
 import { useSpreadState } from '../../hooks/useSpreadState';
 import { getSystemLink } from '../../utils/system';
 import { keyframes } from 'tss-react';
+import { SignalWifiBadRounded } from '@mui/icons-material';
```
rart:
```suggestion
import SignalWifiBadRounded from '@mui/icons-material/SignalWifiBadRounded';
```

rart (approved):

rart (approved):

PR #3133, opened 2022-08-30
[5738] Do not show dialog on create new content if only one type is available

jvega190:
https://github.com/craftercms/craftercms/issues/5738

### DIFF
--- a/ui/app/src/components/NewContentDialog/NewContentDialogContainer.tsx
+++ b/ui/app/src/components/NewContentDialog/NewContentDialogContainer.tsx
@@ -20,8 +20,8 @@ import { FormattedMessage, useIntl } from 'react-intl';
 import { useDispatch } from 'react-redux';
 import useStyles from './styles';
 import { useSelection } from '../../hooks/useSelection';
-import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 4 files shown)


ui/app/src/components/NewContentDialog/NewContentDialogContainer.tsx
```diff
@@ -96,16 +99,21 @@ export function NewContentDialogContainer(props: NewContentDialogContainerProps)
           ? `${selectedItem.path}/`
           : selectedItem.path;
 
-      fetchLegacyContentTypes(site, path).subscribe(
-        (response) => {
-          setContentTypes(response);
+      fetchLegacyContentTypes(site, path).subscribe({
+        next(response) {
+          if (response.length === 1) {
+            dispatch(closeNewContentDialog());
```
rart, line 105: Who closes the dialog when there are multiple content types?
I feel closing the dialog should be done by the consumer of the dialog.

jvega190: The original behavior (and now the behavior when there are more than 1 content-type available) is that it closes after the form has been saved, so it stays open while a form in open (in case it's cancelled you can select another content-type). That behavior is done in this component though.

rart (approved):

PR #3141, opened 2022-09-06
[5764] Audit: When the get parameters call fails, there's no feedback

jvega190:
https://github.com/craftercms/craftercms/issues/5764


jvega190:
pending detail about roles permissions (to get parameters)

rart (approved):

PR #3142, opened 2022-09-07
[5771] Variation in the position of elements of Navigation Menu when accessing in any other role that isn't admin

jvega190:
https://github.com/craftercms/craftercms/issues/5771


rart (changes requested):
Please review as discussed

jvega190:
@rart I updated both skeleton and content container to be the same `Grid` and styles for them to behave consistently. Tested with sitesRail at both sides, and responsive and it's good.

rart (approved):

PR #3150, opened 2022-09-20
[370E] DeleteDialog error handling

jvega190:
https://github.com/craftersoftware/craftercms/issues/370

### DIFF
--- a/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
+++ b/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
@@ -15,7 +15,6 @@
  */
 
 import React, { useEffect, useState } from 'react';
-import { useSpreadState } from '../../hooks/useSpreadState';
 import { useActiveSiteId } from '../../hooks/useActiveSiteId';
 import { useDispatch } from 'react-redux';
 import {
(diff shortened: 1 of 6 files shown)


ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
```diff
@@ -69,6 +69,12 @@ export function DeleteDialogContainer(props: DeleteDialogContainerProps) {
   const [confirmChecked, setConfirmChecked] = useState(false);
   const authoringBase = useSelection((state) => state.env.authoringBase);
 
+  useEffect(() => {
+    if (error) {
+      setApiState({ error });
+    }
+  }, [error, setApiState]);
+
```
rart: Why would you do this instead of just using the error prop directly?

rart (changes requested):

ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
jvega190: We handle two different errors in this dialog, the one when submitting, and the one (the newly added) when the `get_delete_package` fails. Before the change, when submitting we call the submit service and if it fails, we do `setApiState({ error })`. So now that I added the `get_delete_package` error I set the error in the same apiState to use the same mechanism

rart: Make apiState into a simple useState and rename to submitError
Use error prop directly in place with submitError (ie `submitError || error`)
Remove this effect

rart (approved):

PR #3151, opened 2022-09-21
[5772] Migrate to Vite

jvega190:


sumerjabri:
Will not pursue Vite.

PR #3154, opened 2022-09-27
[5754] On content type builder, pasting on the variable name field prepends gibberish

jvega190:
https://github.com/craftercms/craftercms/issues/5754

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-type-propsheet/variable.js
+++ b/static-assets/components/cstudio-admin/mods/content-type-propsheet/variable.js
@@ -34,12 +34,16 @@ YAHOO.extend(
       valueEl.fieldName = this.fieldName;
 
       if (updateFn) {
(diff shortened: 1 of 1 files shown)


rart (changes requested):
Fixes pending

jvega190:
Latest fixes in [0e734d8](https://github.com/craftercms/studio-ui/pull/3154/commits/0e734d8a97f66de66ae53e80bed6425a3adbda04).

rart (approved):
What was the cause of the prepended gibberish on paste?

jvega190:
The paste event was causing the issue, since the event is triggered before the input value is updated we were cleaning the value from the control (still not sure why exactly the gibberish). But after updating the event to `input` now we clean/update the actual new value directly from the input, and the issue is not happening anymore

PR #3167, opened 2022-10-11
Stability

rart:
Display XB errors with error styling (different from success styles), avoid crash in non-secure contexts when copying to clipboard.

### DIFF
--- a/ui/app/src/utils/system.ts
+++ b/ui/app/src/utils/system.ts
@@ -48,7 +48,14 @@ export function getSystemLink({
 }
 
 export function copyToClipboard(textToCopy: string): Promise<void> {
-  return navigator.clipboard.writeText(textToCopy);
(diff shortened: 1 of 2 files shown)


ui/app/src/utils/system.ts
```diff
@@ -48,7 +48,14 @@ export function getSystemLink({
 }
 
 export function copyToClipboard(textToCopy: string): Promise<void> {
-  return navigator.clipboard.writeText(textToCopy);
+  // Clipboard is only available on user-initiated callbacks so this may fail when
+  // attempting to auto-copy over non-secure contexts (e.g. not https).
```
rart:
```suggestion
  // Clipboard is only available on user-initiated callbacks over non-secure contexts (e.g. not https).
```

sumerjabri (approved):

PR #3169, opened 2022-10-11
[5817][next] Malformed Page-URL's can pass through Studio's validation check by using the browsers auto-correct

jvega190:
https://github.com/craftercms/craftercms/issues/5817


rart (changes requested):
turn off autocorrect, spellcheck and autocapitalize on these inputs

rart (approved):

PR #3187, opened 2022-10-31
[5773] Requests to API 2 /audit/:id as developer, returns HTTP 403 instead of 200

jvega190:
https://github.com/craftercms/craftercms/issues/5773

### DIFF
--- a/ui/app/src/components/AuditGrid/AuditGridUI.tsx
+++ b/ui/app/src/components/AuditGrid/AuditGridUI.tsx
@@ -371,7 +371,7 @@ export function AuditGridUI(props: AuditGridUIProps) {
         sortingOrder={['desc', 'asc']}
         sortModel={sortModel}
         sortingMode="server"
-        autoHeight
+        autoHeight={Boolean(auditLogs.length)}
         disableColumnFilter
         className={classes.gridRoot}
         components={{
--- a/ui/app/src/components/AuditManagement/AuditManagement.tsx
+++ b/ui/app/src/components/AuditManagement/AuditManagement.tsx
@@ -132,7 +132,7 @@ export function AuditManagement(props: AuditManagementProps) {
       setDialogParams(parametersLookup[id]);
       auditLogEntryParametersDialogState.onOpen();
     } else {
-      fetchAuditLogEntry(id).subscribe({
+      fetchAuditLogEntry(id, site).subscribe({
         next(response) {
           setParametersLookup({ [id]: response.parameters });
           if (response.parameters.length) {
--- a/ui/app/src/services/audit.ts
+++ b/ui/app/src/services/audit.ts
@@ -60,7 +60,8 @@ export function fetchAuditLog(options: AuditOptions): Observable<PagedArray<Audi
 }
 
 export function fetchAuditLogEntry(id: number, siteId?: string): Observable<AuditLogEntry> {
-  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(
-    `/studio/api/2/audit/${id}${siteId ? `?siteId=${siteId}` : ''}`
-  ).pipe(pluck('response', 'auditLog'));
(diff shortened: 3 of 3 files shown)


ui/app/src/components/AuditGrid/AuditGridUI.tsx
```diff
@@ -371,7 +371,7 @@ export function AuditGridUI(props: AuditGridUIProps) {
         sortingOrder={['desc', 'asc']}
         sortModel={sortModel}
         sortingMode="server"
-        autoHeight
+        autoHeight={Boolean(auditLogs.length)}
```
jvega190, line 374: There was a visual issue when there were no logs (and Empty State shows) and the autoHeight was true. When there are rows the autoHeight works properly.

ui/app/src/components/AuditManagement/AuditManagement.tsx
```diff
@@ -132,7 +132,7 @@ export function AuditManagement(props: AuditManagementProps) {
       setDialogParams(parametersLookup[id]);
       auditLogEntryParametersDialogState.onOpen();
     } else {
-      fetchAuditLogEntry(id).subscribe({
+      fetchAuditLogEntry(id, site).subscribe({
```
rart, line 135: What about global?

rart (changes requested):

ui/app/src/components/AuditManagement/AuditManagement.tsx
jvega190: Site prop is optional (null in global), and the service validates siteId, if null, the param is not sent

ui/app/src/services/audit.ts
```diff
@@ -59,8 +59,8 @@ export function fetchAuditLog(options: AuditOptions): Observable<PagedArray<Audi
   );
 }
 
-export function fetchAuditLogEntry(id: number): Observable<AuditLogEntry> {
-  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(`/studio/api/2/audit/${id}`).pipe(
-    pluck('response', 'auditLog')
-  );
+export function fetchAuditLogEntry(id: number, siteId?: string): Observable<AuditLogEntry> {
+  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(
+    `/studio/api/2/audit/${id}${siteId ? `?siteId=${siteId}` : ''}`
```
rart: use `toQueryString` util

rart (changes requested):

ui/app/src/services/audit.ts
```diff
@@ -59,8 +59,9 @@ export function fetchAuditLog(options: AuditOptions): Observable<PagedArray<Audi
   );
 }
 
-export function fetchAuditLogEntry(id: number): Observable<AuditLogEntry> {
-  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(`/studio/api/2/audit/${id}`).pipe(
+export function fetchAuditLogEntry(id: number, siteId?: string): Observable<AuditLogEntry> {
+  const qs = toQueryString({ siteId });
+  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(`/studio/api/2/audit/${id}${qs}`).pipe(
```
rart: Did you test this?? Without the `skipNull` siteId argument will be sent empty. Global shouldn't have a siteId
```suggestion
  const qs = toQueryString({ siteId }, { skipNull: true });
  return get<Api2ResponseFormat<{ auditLog: AuditLogEntry }>>(`/studio/api/2/audit/${id}${qs}`).pipe(
```

rart (changes requested):

ui/app/src/services/audit.ts
jvega190: I did, when null the queryString ends up like `?`, no siteId parameter

rart (approved):

PR #3198, opened 2022-11-04
[5819] Searching for an image in the form and then clicking on it causes it to expand instead of getting selected

jvega190:
https://github.com/craftercms/craftercms/issues/5819

### DIFF
--- a/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogUI.tsx
+++ b/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogUI.tsx
@@ -49,6 +49,7 @@ export function BrowseFilesDialogUI(props: BrowseFilesDialogUIProps) {
     total,
     numOfLoaderItems = 12,
     onCardSelected,
+    onPreviewImage,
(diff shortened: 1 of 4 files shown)


ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogUI.tsx
```diff
@@ -98,7 +97,7 @@ export function BrowseFilesDialogUI(props: BrowseFilesDialogUIProps) {
                       item={item}
                       selected={multiSelect ? selectedArray : null}
                       onSelect={multiSelect ? onCheckboxChecked : null}
-                      onPreview={() => onPreviewImage(item)}
+                      onPreview={() => onCardSelected(item)}
```
rart: Leave the prop on the _UI_ but don't supply from the _Container_ dialog and default to onCardSelected.
```ts
const { ... onCardSelected, onPreviewImage = onCardSelected } = props
```

rart (changes requested):

rart (approved):

PR #3203, opened 2022-11-09
[451E]

jvega190:
https://github.com/craftersoftware/craftercms/issues/451


rart:
https://github.com/craftercms/studio-ui/pull/3206

PR #3220, opened 2022-12-16
[5889] The Crop Image functionality does not work for existing images

jvega190:
https://github.com/craftercms/craftercms/issues/5889


rart (changes requested):
We probably need to run name policy validation here

rart (approved):

PR #3225, opened 2022-12-28
[5888] Export SearchUI component so that it can easily be used by plugins

jvega190:
https://github.com/craftercms/craftercms/issues/5888


ui/app/src/components/Search/Search.tsx
```diff
@@ -26,18 +26,22 @@ import { debounceTime, distinctUntilChanged } from 'rxjs/operators';
...
 export function Search(props: SearchProps) {
-  const { mode = 'default', onSelect, embedded = false, onAcceptSelection, onClose } = props;
+  const { mode = 'default', onSelect, embedded = false, onAcceptSelection, onClose, parameters } = props;
 
   // region State
   const [filters, setFilters] = useSpreadState({
-    sortBy: undefined,
-    sortOrder: undefined
+    sortBy: parameters?.sortBy ?? undefined,
+    sortOrder: parameters?.sortOrder ?? undefined
```
rart: I do wonder if this is even necessary. If these props aren't supplied, they will already be undefined. If only the sub-props are not supplied, they will also be undefined.
 ```suggestion
    sortBy: parameters?.sortBy ?? UNDEFINED,
    sortOrder: parameters?.sortOrder ?? UNDEFINED
```

File: ui/app/src/components/Search/utils.ts [hunk 2]
```diff
@@ -72,9 +72,14 @@ export interface URLDrivenSearchProps {
...
 }
 
+export interface SearchParameters extends Partial<ElasticParams> {
+  path?: string;
+}
+
 export interface SearchProps {
   mode?: 'default' | 'select';
   embedded?: boolean;
+  parameters?: SearchParameters;
```
rart:
```suggestion
  initialFilters?: SearchParameters;
```

ui/app/src/components/Search/utils.ts
```diff
@@ -72,9 +72,14 @@ export interface URLDrivenSearchProps {
   onAcceptSelection?(items: DetailedItem[]): any;
 }
 
+export interface SearchParameters extends Partial<ElasticParams> {
```
rart, line 75:
```suggestion
export interface SearchFilters extends Partial<ElasticParams> {
```

ui/app/src/components/Search/Search.tsx
```diff
@@ -70,6 +74,23 @@ export function Search(props: SearchProps) {
...
+      const newCheckedFilters = {};
+      Object.entries(parameters.filters).forEach(([key, filter]: [string, any]) => {
+        if (Array.isArray(filter)) {
+          newCheckedFilters[key] = filter.reduce((checked, key) => ({ ...checked, [key]: true }), {});
+        } else if (filter.date) {
+          newCheckedFilters[key] = `${filter.min ?? ''}TODATE${filter.max ?? ''}ID${filter.id}`;
+        } else {
+          newCheckedFilters[key] = `${filter.min ?? ''}TO${filter.max ?? ''}`;
+        }
+      });
```
rart: Preferably this should be in a single place. Where does this concatenation and such usually occurs? I would have thought this all should be on the useSearchState hook.

rart (changes requested):

ui/app/src/components/Search/utils.ts
jvega190: per conversation, it'll be called 'initialParameters'

ui/app/src/components/Search/utils.ts
jvega190: per conversation, it'll remain called 'SearchParameters'

rart (approved):

PR #3226, opened 2022-12-28
[5895] Rename asset dialog dependency edit button doesn't open the right mode for FTL

jvega190:
https://github.com/craftercms/craftercms/issues/5895

### DIFF
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -997,6 +997,15 @@ export const openItemEditor = (
   if (type === 'form') {
     dispatch(showEditDialog({ path: item.path, authoringBase, site: siteId, onSaveSuccess }));
   } else {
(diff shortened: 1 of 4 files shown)


ui/app/src/utils/content.ts
```diff
@@ -997,6 +997,15 @@ export const openItemEditor = (
...
     dispatch(showEditDialog({ path: item.path, authoringBase, site: siteId, onSaveSuccess }));
   } else {
-    dispatch(showCodeEditorDialog({ site: siteId, authoringBase, path: item.path, type, onSuccess: onSaveSuccess }));
+    dispatch(
+      showCodeEditorDialog({
+        site: siteId,
+        authoringBase,
+        path: item.path,
+        type,
+        mode: getEditorMode(item.mimeType),
```
rart, line 1006: Should also fix the mode-null thing. If mode isn't provided it should fallback to text without failure.

rart (changes requested):

rart (approved):

PR #3227, opened 2022-12-29
[5886] Upgrade Fast XML Parser

jvega190:
https://github.com/craftercms/craftercms/issues/5886


ui/app/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -114,15 +114,7 @@ function DateTimePicker(props: DateTimePickerProps) {
   const internalDate = useMemo(() => {
     const date = value ? new Date(value) : new Date();
     const localOffset = moment().format().substr(-6);
-    let dateWithoutOffset;
```
jvega190, line 117: Reverted validation since converting is needed because the date is created in local timezone

rart (approved):

PR #3263, opened 2023-02-17
[5940] Consider repository path in components datasource when creating components

jvega190:
https://github.com/craftercms/craftercms/issues/5940

### DIFF
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -1009,3 +1009,10 @@ export const openItemEditor = (
     );
   }
 };
+
+export function generateComponentBasePath(contentType: string) {
+  return `/site/components/${contentType.replace('/component/', '')}s/`.replace(/\/{1,}$/m, '');
+}
+export function generateComponentPath(modelId: string, contentType: string) {
+  return `${generateComponentBasePath(contentType)}/${modelId}.xml`;
+}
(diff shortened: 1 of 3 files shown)


ui/app/src/utils/content.ts
```diff
@@ -1009,3 +1009,10 @@ export const openItemEditor = (
     );
   }
 };
+
+export function generateComponentBasePath(contentType: string) {
+  return `/site/components/${contentType.replace('/component/', '')}s/`.replace(/\/{1,}$/m, '');
+}
```
rart, line 1015:
```suggestion
}

```

ui/legacy/src/components/cstudio-forms/data-sources/components.js
```diff
@@ -30,7 +30,8 @@
     this.allowEmbedded = false;
     this.enableSearch = false;
     this.enableBrowse = false;
-    this.baseRepoPath = '/site/components';
+    this.defaultBaseRepoPath = '/site/components';
+    this.baseRepoPath = null;
     this.baseBrowsePath = '/site/components';
```
rart, line 35: Does the base browse path work correctly?

rart (changes requested):

ui/legacy/src/components/cstudio-forms/data-sources/components.js
jvega190: yes it does, it gets overridden later in the code (if set in content-type editor prop)

rart:
Test form with more than one components data source on the same item selector making sure the embedded and shared rules work correctly (review ticket desc), paths for browse/search and repo path all work correctly

rart (approved):

PR #3266, opened 2023-02-21
[5946] Validate XB version against studio version

jvega190:
https://github.com/craftercms/craftercms/issues/5946

### DIFF
--- a/ui/app/src/assets/guestMessages.ts
+++ b/ui/app/src/assets/guestMessages.ts
@@ -161,5 +161,10 @@ export const guestMessages = defineMessages({
   uploadError: {
     id: 'desktopAssetUpload.uploadError',
     defaultMessage: 'Upload error. File was not uploaded'
+  },
+  invalidExpBuilderVersion: {
+    id: 'validations.invalidExpBuilderVersion',
+    defaultMessage:
+      'Your application’s Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues.'
   }
 });
--- a/ui/app/src/assets/guestMessages.ts
+++ b/ui/app/src/assets/guestMessages.ts
@@ -165,6 +165,6 @@ export const guestMessages = defineMessages({
   outdatedExpBuilderVersion: {
     id: 'validations.outdatedExpBuilderVersion',
     defaultMessage:
-      'Your application’s Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues.'
+      "Your application's Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues."
   }
 });
--- a/ui/app/src/assets/guestMessages.ts
+++ b/ui/app/src/assets/guestMessages.ts
@@ -165,6 +165,6 @@ export const guestMessages = defineMessages({
   outdatedExpBuilderVersion: {
     id: 'validations.outdatedExpBuilderVersion',
     defaultMessage:
-      "Your application's Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues."
+      "Your application's Experience Builder package version is out of date. Please notify the development team to perform the necessary updates to avoid any possible issues."
   }
 });
--- a/ui/app/src/assets/guestMessages.ts
+++ b/ui/app/src/assets/guestMessages.ts
@@ -162,8 +162,8 @@ export const guestMessages = defineMessages({
     id: 'desktopAssetUpload.uploadError',
     defaultMessage: 'Upload error. File was not uploaded'
   },
-  invalidExpBuilderVersion: {
-    id: 'validations.invalidExpBuilderVersion',
+  outdatedExpBuilderVersion: {
+    id: 'validations.outdatedExpBuilderVersion',
     defaultMessage:
       'Your application’s Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues.'
   }
--- a/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
+++ b/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
@@ -325,6 +325,7 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
   const [dataSourceActionsListState, setDataSourceActionsListState] = useSpreadState<DataSourcesActionsListProps>(
     dataSourceActionsListInitialState
   );
+  const env = useEnv();
   const conditionallyToggleEditMode = (nextHighlightMode?: HighlightMode) => {
     if (item && !isItemLockedForMe(item, user.username) && hasEditAction(item.availableActions)) {
       dispatch(
@@ -422,7 +423,8 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
           }
           break;
       }
-    }
+    },
+    env
   });
 
   const onRtePickerResult = (payload?: { path: string; name: string }) => {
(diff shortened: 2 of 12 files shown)


ui/app/src/assets/guestMessages.ts
```diff
@@ -161,5 +161,10 @@ export const guestMessages = defineMessages({
   uploadError: {
     id: 'desktopAssetUpload.uploadError',
     defaultMessage: 'Upload error. File was not uploaded'
+  },
+  outdatedExpBuilderVersion: {
+    id: 'validations.outdatedExpBuilderVersion',
+    defaultMessage:
+      'Your application’s Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues.'
```
sumerjabri: Use `'` not `

Use this message instead:
```
Your application's Experience Builder package version is out of date. Please notify the development team to perform the necessary updates to avoid any possible issues.
```

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -550,12 +552,20 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
...
+        env
       } = upToDateRefs.current;
       const { type, payload } = action;
       switch (type) {
         case guestSiteLoad.type:
         case guestCheckIn.type:
+          const { version: guestVersion } = payload;
+          const studioVersion = env.version;
+
+          if (type === guestCheckIn.type && (nou(guestVersion) || guestVersion.substr(0, 5) !== studioVersion)) {
```
sumerjabri: Where is the check against a list of valid versions (or ranges)? Did I miss it?

sumerjabri (changes requested):

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
jvega190: @sumerjabri it will be checking 'base' versions, so if XB is for example `4.0.3.2` and studio is `4.0.3` it will match `4.0.3`, if not, it is considered outdated

ui/app/src/assets/guestMessages.ts
```diff
@@ -161,5 +161,10 @@ export const guestMessages = defineMessages({
   uploadError: {
     id: 'desktopAssetUpload.uploadError',
     defaultMessage: 'Upload error. File was not uploaded'
+  },
+  outdatedExpBuilderVersion: {
+    id: 'validations.outdatedExpBuilderVersion',
+    defaultMessage:
+      "Your application's Experience Builder package version is out of date. Please notify the development team to perform the update to avoid issues."
```
sumerjabri: Please update the string to what I indicated in the prior comment.

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
sumerjabri: I don't like this. It should be checked against a list (or ranges) of compatible versions.

ui/app/src/assets/guestMessages.ts
sumerjabri: This is the string:

```
Your application's Experience Builder package version is out of date. Please notify the development team to perform the necessary updates to avoid any possible issues.
```

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
rart: This code is not meant to be forever. This is meant as a "_crawl_" and on our next release, implement something more robust and comprehensive, test it appropriately and better understand what the checklist should be.

Note this logic is **not** embedded in client apps, it's in Studio so anyone who upgrades gets the update — even if their app embeds the SDK.

There is no list to be checked against just yet, and we don't know what the next version that breaks compatibility will be. Right now, is the first time we're implementing XB reporting its version and Studio checking it, hence, we're going to get either nothing (if the package is older), or `4.0.3` or eventually `4.0.3.{something}` if there's a patch. By the time there's a `4.0.4`, it means there's a new version of CrafterCMS out too with new improved compatibility checking logic. Makes sense?

If you still disagree, let's meet and decide on what the list or range will be. I don't have full grasp on what that should be which is why I wanted a _crawl_ start as we approach `4.0.3` release.

ui/guest/scripts/replaceEnvVersion.js
```diff
@@ -0,0 +1,29 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const pkg = require('../../app/package.json');
```
rart: Why use app's package.json?
```suggestion
const pkg = require('../package.json');
```

rart (changes requested):

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
rart:
```suggestion
          if (type === guestCheckIn.type && guestVersion?.substr(0, 5) !== studioVersion) {
```

rart: Reviewed with Sumer. Will start with this approach.

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -565,7 +575,7 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
         case guestSiteLoad.type: {
           const { url, location } = payload;
           const path = getPathFromPreviewURL(url);
-          dispatch(guestCheckIn({ location, site: siteId, path }));
+          dispatch(guestCheckIn({ location, site: siteId, path, version: '' }));
```
rart:
```suggestion
          dispatch(guestCheckIn({ location, site: siteId, path }));
```

ui/app/src/state/actions/preview.ts
```diff
@@ -54,8 +54,7 @@ export const guestCheckIn = /*#__PURE__*/ createAction<{
   path: string;
   site: string;
   documentDomain?: string;
-  // TODO:
-  // version: string;
+  version: string;
```
rart:
```suggestion
  version?: string;
```

rart (changes requested):

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -550,12 +552,20 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
...
       } = upToDateRefs.current;
       const { type, payload } = action;
       switch (type) {
         case guestSiteLoad.type:
         case guestCheckIn.type:
+          const { version: guestVersion } = payload;
+          const studioVersion = env.version;
+
+          if (type === guestCheckIn.type && guestVersion?.substr(0, 5) !== studioVersion) {
+            enqueueSnackbar(formatMessage(guestMessages.outdatedExpBuilderVersion));
```
rart: Just thought this would be too intrusive as it would snack on every page navigation, refresh, edit, etc. Show it once per day.

rart (changes requested):

rart (approved):

PR #3267, opened 2023-02-22
[5952] Workflow States errors when selecting items

jvega190:
https://github.com/craftercms/craftercms/issues/5952

### DIFF
--- a/ui/app/src/components/WorkflowStateManagement/WorkflowStateManagement.tsx
+++ b/ui/app/src/components/WorkflowStateManagement/WorkflowStateManagement.tsx
@@ -89,7 +89,6 @@ export function WorkflowStateManagement(props: WorkflowStateManagementProps) {
   const { embedded, showAppsButton = !embedded, onSubmittingAndOrPendingChange } = props;
   const [fetching, setFetching] = useState(false);
   const [items, setItems] = useState<PagedArray<SandboxItem>>(null);
-  const [totalItems, setTotalItems] = useState(0);
   const [error, setError] = useState<ApiResponse>();
   const siteId = useActiveSiteId();
   const [openSetStateDialog, setOpenSetStateDialog] = useState(false);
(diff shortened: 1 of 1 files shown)


ui/app/src/components/WorkflowStateManagement/WorkflowStateManagement.tsx
```diff
@@ -89,6 +89,7 @@ export function WorkflowStateManagement(props: WorkflowStateManagementProps) {
   const { embedded, showAppsButton = !embedded, onSubmittingAndOrPendingChange } = props;
   const [fetching, setFetching] = useState(false);
   const [items, setItems] = useState<PagedArray<SandboxItem>>(null);
+  const [totalItems, setTotalItems] = useState(0);
```
rart: Feels a bit like slipping the error under the rug. Why is it trying to render things before they are set? Is it maybe just missing a `.?` somewhere?

Is a new piece of state really necessary?

```suggestion
  const [items, setItems] = useState<PagedArray<SandboxItem>>(() => Object.assign([], { total: 0 }));
```

rart: Make sure you're actually testing the 2 reproduction paths and understanding what's happening.

rart (changes requested):

ui/app/src/components/WorkflowStateManagement/WorkflowStateManagement.tsx
```diff
@@ -364,7 +363,7 @@ export function WorkflowStateManagement(props: WorkflowStateManagementProps) {
             title={<FormattedMessage id="itemStates.emptyStateMessage" defaultMessage="No results found" />}
           />
         )}
-        {items?.length && (
+        {Boolean(items?.length) && (
```
jvega190, line 366: Added the `Boolean` to avoid a rendering of a 'zero' when there are no items

rart (approved):

PR #3271, opened 2023-02-28
[5961] Publishing queue - 'All' filter shouldn't uncheck options once they're all already checked.

jvega190:
https://github.com/craftercms/craftercms/issues/5961

### DIFF
--- a/ui/app/src/components/CreateSiteDialog/index.ts
+++ b/ui/app/src/components/CreateSiteDialog/index.ts
@@ -24,6 +24,4 @@ export * from './BlueprintForm';
 
 export * from './BlueprintReview';
 
-export * from './FilterDropdown';
-
 export * from './GitForm';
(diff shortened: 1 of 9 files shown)


rart (changes requested):
Ping me to discuss some additional updates

ui/app/src/components/CreateSiteDialog/index.ts
```diff
@@ -24,6 +24,4 @@ export * from './BlueprintForm';
 
 export * from './BlueprintReview';
 
-export * from './FilterDropdown';
```
jvega190, line 27: Moved this out of here since it is only being used by PublishingQueue

ui/app/src/models/Publishing.ts
```diff
@@ -26,9 +28,7 @@ export interface Package {
   comment: string;
 }
 
-export interface Selected {
-  [keys: string]: boolean;
-}
+export type Selected = LookupTable<boolean>;
```
jvega190, line 31: This is being used in multiple parts, but updated it since it is basically a LookupTable

ui/app/src/components/PublishingDashboard/PublishingDashboard.tsx
```diff
@@ -70,7 +70,7 @@ export function PublishingDashboard(props: PublishingDashboardProps) {
         </Grid>
         {userPermissions.includes('get_publishing_queue') && (
           <Grid item xs={12}>
-            <PublishingQueueWidget siteId={site} />
+            <PublishingQueueWidget siteId={site} readOnly={!userPermissions.includes('publish')} />
```
rart: Cache the possession of the publish permission to avoid running the `.includes` twice

rart (changes requested):

rart (approved):

PR #3279, opened 2023-03-03
[5972] Blob Store: Handle case when blob item is missing but present site git/db

jvega190:
https://github.com/craftercms/craftercms/issues/5972

### DIFF
--- a/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
+++ b/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
@@ -36,8 +36,7 @@ import { DateChangeData } from '../DateTimePicker/DateTimePicker';
 import moment from 'moment-timezone';
 import { updatePublishDialog } from '../../state/actions/dialogs';
 import { approve, publish, requestPublish } from '../../services/workflow';
-import { forkJoin } from 'rxjs';
-import { fetchDetailedItem } from '../../services/content';
+import { fetchDetailedItems } from '../../services/content';
 import { DetailedItem } from '../../models';
 import { fetchDetailedItemComplete } from '../../state/actions/content';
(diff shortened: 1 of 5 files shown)


ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -227,17 +205,39 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
     state.publishingTarget,
     state.submissionComment,
-    submissionCommentRequired
+    submissionCommentRequired,
+    state.error
   ]);
 
+  useEffect(() => {
+    setIsFetchingItems(true);
+    forkJoin(items.map((item) => fetchDetailedItem(siteId, item.path))).subscribe({
```
rart: Why not use `fetchDetailedItems`?

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -227,17 +205,39 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
+        setIsFetchingItems(false);
+      },
+      error(error) {
+        setState({
+          error: error.response?.response ?? error
+        });
+        setIsFetchingItems(false);
+      }
+    });
+  }, [items, siteId, setState, dispatch]);
```
rart, line 242: Make sure these dependencies don't cause infinite fetching. Particularly the `items` array.

jvega190: they don't cause infinite fetching, the items array is not the detailedItems fetched in the effect

rart (approved):

PR #3296, opened 2023-04-12
[5974] Update Users and Groups form to adapt to new validation framework regex properties

jvega190:
https://github.com/craftercms/craftercms/issues/5974

### DIFF
--- a/ui/app/src/components/GroupManagement/utils.ts
+++ b/ui/app/src/components/GroupManagement/utils.ts
@@ -27,6 +27,6 @@ export const validateRequiredField = (value: string, isDirty: boolean) => {
 };
 
 export const isInvalidGroupName = (groupName: string): boolean => {
-  const groupNameRegex = /^[a-zA-Z0-9_\\.\\-]{3,512}$/g;
+  const groupNameRegex = /^[a-zA-Z][\w.-]*$/g;
   return Boolean(groupName) && !groupNameRegex.test(groupName);
 };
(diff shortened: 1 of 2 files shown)


ui/app/src/components/GroupManagement/utils.ts
```diff
@@ -27,6 +27,6 @@ export const validateRequiredField = (value: string, isDirty: boolean) => {
 };
 
 export const isInvalidGroupName = (groupName: string): boolean => {
-  const groupNameRegex = /^[a-zA-Z0-9_\\.\\-]{3,512}$/g;
+  const groupNameRegex = /^[a-zA-Z][\w.-]*$/g;
```
jvega190, line 30: Removed the num of chars contraints since that is already being validated with another const

rart: Which expression validates the numbers? Where in the code?

jvega190: https://github.com/craftercms/studio-ui/blob/develop/ui/app/src/components/GroupManagement/utils.ts#L17 those are the consts. The minLength for groups is validated in [here](https://github.com/craftercms/studio-ui/blob/develop/ui/app/src/components/EditGroupDialog/EditGroupDialogContainer.tsx#L77), and the max is set directly on the input in [here](https://github.com/craftercms/studio-ui/blob/develop/ui/app/src/components/EditGroupDialog/EditGroupDialogUI.tsx#L165) for the input to avoid entering more than the allowed number of chars

rart (approved):

PR #3299, opened 2023-04-17
[6015] Websocket throws 302 error when session timeout

jvega190:
https://github.com/craftercms/craftercms/issues/6015


ui/app/src/shared-worker.ts
```diff
@@ -107,6 +107,7 @@ function unauthenticated(excludeClient?: MessagePort) {
   log(`Auth has expired.`);
   clearTimeout(timeout);
   status = 'expired';
+  socketClient?.deactivate();
```
jvega190, line 110: What was missing was to deactivate the client when the session expired. That caused the client to retry the connection causing the 302 error. When re-connecting, the client already gets enabled again.

rart (approved):

PR #3300, opened 2023-04-17
[6008] Site policy does not apply when uploading image from image upload control

jvega190:
https://github.com/craftercms/craftercms/issues/6008

### DIFF
--- a/ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
+++ b/ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
@@ -61,7 +61,7 @@ const messages = defineMessages({
   },
   policyError: {
     id: 'fileUpload.policyError',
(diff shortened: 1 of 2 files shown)


ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
```diff
@@ -61,7 +61,7 @@ const messages = defineMessages({
   },
   policyError: {
     id: 'fileUpload.policyError',
-    defaultMessage: 'The upload file name goes against project policies.'
+    defaultMessage: 'File "{fileName}" doesn\'t comply with project policies and can\'t be uploaded'
```
jvega190, line 64: Updated this message to match the one in XB, since the policy is not only for fileName

rart (approved):

PR #3301, opened 2023-04-18
[5837] Page-URL field is calculated as 2 required fields

jvega190:
https://github.com/craftercms/craftercms/issues/5837

### DIFF
--- a/static-assets/components/cstudio-forms/controls/file-name.js
+++ b/static-assets/components/cstudio-forms/controls/file-name.js
@@ -116,10 +116,10 @@ YAHOO.extend(CStudioForms.Controls.FileName, CStudioForms.CStudioFormField, {
   },
 
   getRequirementCount: function () {
-    // 2 Requirement:
+    // 2 Requirements for this field, both will be validated as one (since it's only one field in the form):
     // 1. The field is required
     // 2. The Path must be valid
-    return 2;
+    return 1;
   },
 
   getCurrentPath: function () {
--- a/static-assets/components/cstudio-forms/controls/file-name.js
+++ b/static-assets/components/cstudio-forms/controls/file-name.js
@@ -134,7 +134,7 @@ YAHOO.extend(CStudioForms.Controls.FileName, CStudioForms.CStudioFormField, {
   _onChange: function (evt, obj) {
     var oldValue = obj.value;
     obj.value = obj.inputEl.value;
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/controls/file-name.js
```diff
@@ -116,10 +116,10 @@ YAHOO.extend(CStudioForms.Controls.FileName, CStudioForms.CStudioFormField, {
   },
 
   getRequirementCount: function () {
```
jvega190, line 118: There are two possible validations (requirements): required and path available. Since they can't collide it can be treated as just one requirement (already done in other controls)

static-assets/components/cstudio-forms/controls/file-name.js
```diff
@@ -134,7 +134,7 @@ YAHOO.extend(CStudioForms.Controls.FileName, CStudioForms.CStudioFormField, {
   _onChange: function (evt, obj) {
     var oldValue = obj.value;
     obj.value = obj.inputEl.value;
-    if (obj.value != '' && oldValue != obj.value)
+    if (oldValue != obj.value)
```
jvega190, line 137: The `not empty` validation was removed because the validation cleans the error when the field is empty. With the validation there was an scenario where the error was not being cleared properly.

rart: So what happens now when it is empty?

static-assets/components/cstudio-forms/controls/file-name.js
rart: Is this done in other controls at this level?

Ideally I'd expect this to return 2 and then somewhere _higher_ in the code, it would be handled as one for the specific purpose. We're not going to redo this everywhere for FE1, specially if done like this elsewhere.

static-assets/components/cstudio-forms/controls/file-name.js
jvega190: When the value is empty, it doesn't verify if the path exists, it only clears any previous `path exists` error

static-assets/components/cstudio-forms/controls/file-name.js
jvega190: Yeah at the same level, higher in the code it doesn't handle it as one, just adds the count to the total display at the top of the form section

rart (approved):

PR #3302, opened 2023-04-19
[6005] Pop-up buttons with multiple options are cut off when at the top of the content form

jvega190:
https://github.com/craftercms/craftercms/issues/6005


rart (commented):
Is there a flicker because of this, @jvega190 ?

jvega190:
@rart not any flicker noticeable (tested in chrome and firefox for linux)

rart (approved):

PR #3303, opened 2023-04-20
[6014] Delete dialog - When users checks the Dependent Items the Child Items should update the child items to remove

jvega190:
https://github.com/craftercms/craftercms/issues/6014

### DIFF
--- a/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
+++ b/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
@@ -64,7 +64,7 @@ export function DeleteDialogContainer(props: DeleteDialogContainerProps) {
   const itemsByPath = useItemsByPath();
   useFetchSandboxItems(dependentItems ?? []);
   const disabledDependentItems = createPresenceTable(
-    dependentItems?.filter((item) => !itemsByPath[item].availableActionsMap.delete) ?? [],
+    dependentItems?.filter((path) => !itemsByPath[path]?.availableActionsMap.delete) ?? [],
     true
   );
 
--- a/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
+++ b/ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
@@ -32,6 +32,8 @@ import { createPresenceTable } from '../../utils/array';
 import { DetailedItem } from '../../models/Item';
 import { isBlank } from '../../utils/string';
 import { ApiResponse } from '../../models';
+import useItemsByPath from '../../hooks/useItemsByPath';
+import useFetchSandboxItems from '../../hooks/useFetchSandboxItems';
 
 function createCheckedList(selectedItems: LookupTable<boolean>, excludedPaths?: string[]) {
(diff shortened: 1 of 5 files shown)


ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
```diff
@@ -125,16 +130,18 @@ export function DeleteDialogContainer(props: DeleteDialogContainerProps) {
   };
 
   const onSelectAllDependantClicked = () => {
-    const setChecked = Boolean(dependentItems.find((path) => !selectedItems[path]));
+    const selectableDependent = dependentItems.filter((item) => !disabledDependentItems.includes(item));
+    const setChecked = Boolean(selectableDependent.find((path) => !selectedItems[path]));
```
rart: Perhaps .some is more appropriate here

ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
```diff
@@ -59,6 +61,9 @@ export function DeleteDialogContainer(props: DeleteDialogContainerProps) {
   const [submitDisabled, setSubmitDisabled] = useState(true);
   const [confirmChecked, setConfirmChecked] = useState(false);
   const authoringBase = useSelection((state) => state.env.authoringBase);
+  const itemsByPath = useItemsByPath();
+  useFetchSandboxItems(dependentItems ?? []);
```
rart, line 65: How are other items fetched? Could we use same block to load these?

ui/app/src/components/DependencySelection/SelectionList.tsx
```diff
@@ -105,6 +106,7 @@ export function SelectionList(props: SelectionListProps) {
     onSelectAllClicked,
     selectedItems,
     disabled = false,
+    disabledPaths = [],
```
rart: Might perform better for this to be a lookup?

rart (changes requested):

ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
jvega190: The items are a prop of the DeleteDialog, so they're coming from outside the dialog (the items selected for deletion)

ui/app/src/components/DeleteDialog/DeleteDialogContainer.tsx
```diff
@@ -59,6 +61,12 @@ export function DeleteDialogContainer(props: DeleteDialogContainerProps) {
...
+  const disabledDependentItems = createPresenceTable(
+    dependentItems?.filter((item) => !itemsByPath[item].availableActionsMap.delete) ?? [],
```
rart: Specially with the new logic, is there a moment where `itemsByPath` does't have all the paths loaded? You should throttle for testing
```suggestion
    dependentItems?.filter((path) => !itemsByPath[path]?.availableActionsMap.delete) ?? [],
```

rart (changes requested):

rart (approved):

PR #3306, opened 2023-04-25
Stability updates

jvega190:
- Fix - pressing enter key on token form of create site dialog doesn’t progress the dialog
- Fix - open content type, without having modified anything, click on an image-picker control, “Close” button becomes “Cancel”, pressing cancel shows “Content Type has been modified, are you sure you want to discard your changes?” despite not changing anything. 
- Fix - In XB when an RTE is initialised, it creates a clone element that doesn’t keep the original’s classes which breaks styles (e.g. above bp, try to edit the marquees). research maintaining styles and scroll position which in a really long RTE content, it scrolls you up to the top of that rte.

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-types.js
+++ b/static-assets/components/cstudio-admin/mods/content-types.js
@@ -2696,7 +2696,7 @@
 
           var updatePropertyFn = function (name, value) {
             for (var l = item.properties.length - 1; l >= 0; l--) {
-              if (item.properties[l].name === name) {
+              if (item.properties[l].name === name && item.properties[l].value !== value) {
                 onSetDirty(true);
                 item.properties[l].value =
                   typeof value == 'object' && !Array.isArray(value) ? JSON.stringify(value) : value;
--- a/ui/app/src/components/CreateSiteDialog/GitForm.tsx
+++ b/ui/app/src/components/CreateSiteDialog/GitForm.tsx
@@ -176,7 +176,12 @@ function GitForm(props: GitFormProps) {
       </Grid>
       <Grid item xs={12}>
         <div className={classes.formControl}>
-          <GitAuthForm inputs={inputs} setInputs={setInputs} handleInputChange={handleInputChange} />
+          <GitAuthForm
+            inputs={inputs}
+            setInputs={setInputs}
+            handleInputChange={handleInputChange}
+            onKeyPress={onKeyPress}
+          />
         </div>
       </Grid>
       <Grid item xs={12} sx={{ mb: 2 }}>
--- a/ui/guest/src/controls/rte.ts
+++ b/ui/guest/src/controls/rte.ts
@@ -42,23 +42,34 @@ export function initTinyMCE(
   const { field, model } = iceRegistry.getReferentialEntries(record.iceIds[0]);
   const type = field?.type;
   const inlineElsRegex =
-    /B|BIG|I|SMALL|TT|ABBR|ACRINYM|CITE|CODE|DFN|EM|KBD|STRONG|SAMP|VAR|A|BDO|BR|IMG|MAP|OBJECT|Q|SCRIPT|SPAN|SUB|SUP|BUTTON|INPUT|LABEL|SELECT|TEXTAREA/;
+    /^B$|^BIG$|^I$|^SMALL$|^TT$|^ABBR$|^ACRINYM$|^CITE$|^CODE$|^DFN$|^EM$|^KBD$|^STRONG$|^SAMP$|^VAR$|^A$|^BDO$|^BR$|^IMG$|^MAP$|^OBJECT$|^Q$|^SCRIPT$|^SPAN$|^SUB$|^SUP$|^BUTTON$|^INPUT$|^LABEL$|^SELECT$|^TEXTAREA$/;
   let rteEl = record.element;
   const isRecordElInline = record.element.tagName.match(inlineElsRegex);
 
   // If record element is of type inline (doesn't matter the display prop), replace it with a block element (div).
   if (isRecordElInline) {
+    const recordEl = record.element;
     const blockEl = document.createElement('div');
-    blockEl.innerHTML = record.element.innerHTML;
+    blockEl.innerHTML = recordEl.innerHTML;
+
+    // Copy original element inline styles (copying all styles causes issues with XB styles because they end up all
+    // being inline styles).
+    const inlineStyles = recordEl.style;
+    blockEl.style.cssText = Array.from(inlineStyles).reduce((str, property) => {
+      return `${str}${property}:${inlineStyles.getPropertyValue(property)};`;
+    }, '');
+
+    // Copy original element className
+    blockEl.className = recordEl.className;
     blockEl.style.display = 'inline-block';
 
-    blockEl.style.minHeight = record.element.offsetHeight + 'px';
+    blockEl.style.minHeight = recordEl.offsetHeight + 'px';
     blockEl.style.minWidth = '10px';
     rteEl = blockEl;
 
     // Hide original element
-    record.element.style.display = 'none';
-    record.element.parentNode.insertBefore(rteEl, record.element);
+    recordEl.style.display = 'none';
+    recordEl.parentNode.insertBefore(rteEl, recordEl);
   }
 
   const openEditForm = () => {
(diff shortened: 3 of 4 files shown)


static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -2696,7 +2696,7 @@
 
           var updatePropertyFn = function (name, value) {
             for (var l = item.properties.length - 1; l >= 0; l--) {
-              if (item.properties[l].name === name) {
+              if (item.properties[l].name === name && item.properties[l].value !== value) {
```
rart, line 2699: This seems to fix the image picker only. There seems to be a problem in the "changes pending dialog": why pressing accept doesn't close?

ui/app/src/components/CreateSiteDialog/GitForm.tsx
```diff
@@ -176,7 +176,12 @@ function GitForm(props: GitFormProps) {
       </Grid>
       <Grid item xs={12}>
         <div className={classes.formControl}>
-          <GitAuthForm inputs={inputs} setInputs={setInputs} handleInputChange={handleInputChange} />
+          <GitAuthForm
+            inputs={inputs}
+            setInputs={setInputs}
+            handleInputChange={handleInputChange}
+            onKeyPress={onKeyPress}
```
rart, line 183: Check the rest of the dialog for same issue

ui/guest/src/controls/rte.ts
```diff
@@ -42,23 +42,34 @@ export function initTinyMCE(
   const { field, model } = iceRegistry.getReferentialEntries(record.iceIds[0]);
   const type = field?.type;
   const inlineElsRegex =
-    /B|BIG|I|SMALL|TT|ABBR|ACRINYM|CITE|CODE|DFN|EM|KBD|STRONG|SAMP|VAR|A|BDO|BR|IMG|MAP|OBJECT|Q|SCRIPT|SPAN|SUB|SUP|BUTTON|INPUT|LABEL|SELECT|TEXTAREA/;
+    /^B$|^BIG$|^I$|^SMALL$|^TT$|^ABBR$|^ACRINYM$|^CITE$|^CODE$|^DFN$|^EM$|^KBD$|^STRONG$|^SAMP$|^VAR$|^A$|^BDO$|^BR$|^IMG$|^MAP$|^OBJECT$|^Q$|^SCRIPT$|^SPAN$|^SUB$|^SUP$|^BUTTON$|^INPUT$|^LABEL$|^SELECT$|^TEXTAREA$/;
   let rteEl = record.element;
   const isRecordElInline = record.element.tagName.match(inlineElsRegex);
 
   // If record element is of type inline (doesn't matter the display prop), replace it with a block element (div).
```
rart, line 49: What was the reason for this? Pls expand comment.

What was the reason to create a different elem in the first place?

ui/guest/src/controls/rte.ts
```diff
@@ -42,23 +42,34 @@ export function initTinyMCE(
...
   if (isRecordElInline) {
+    const recordEl = record.element;
     const blockEl = document.createElement('div');
-    blockEl.innerHTML = record.element.innerHTML;
+    blockEl.innerHTML = recordEl.innerHTML;
+
+    // Copy original element inline styles (copying all styles causes issues with XB styles because they end up all
+    // being inline styles).
```
rart: What are "all styles"? Isn't that what you're doing? What's the problem with "all ending up as inline"?

ui/guest/src/controls/rte.ts
```diff
@@ -42,23 +42,34 @@ export function initTinyMCE(
   const { field, model } = iceRegistry.getReferentialEntries(record.iceIds[0]);
   const type = field?.type;
   const inlineElsRegex =
-    /B|BIG|I|SMALL|TT|ABBR|ACRINYM|CITE|CODE|DFN|EM|KBD|STRONG|SAMP|VAR|A|BDO|BR|IMG|MAP|OBJECT|Q|SCRIPT|SPAN|SUB|SUP|BUTTON|INPUT|LABEL|SELECT|TEXTAREA/;
+    /^B$|^BIG$|^I$|^SMALL$|^TT$|^ABBR$|^ACRINYM$|^CITE$|^CODE$|^DFN$|^EM$|^KBD$|^STRONG$|^SAMP$|^VAR$|^A$|^BDO$|^BR$|^IMG$|^MAP$|^OBJECT$|^Q$|^SCRIPT$|^SPAN$|^SUB$|^SUP$|^BUTTON$|^INPUT$|^LABEL$|^SELECT$|^TEXTAREA$/;
```
rart: Was the regexp not matching?

Maybe we shouldn't be looking at elements but looking at their style property since one can change de display of it via styles. It says below the display prop doesn't matter, though, why?

rart (changes requested):

static-assets/components/cstudio-admin/mods/content-types.js
jvega190: It wouldn't fix only the image picker. What is happening is that in this case (the image picker), when clicking on it and rendering the properties explorer, two of the fields are width and height, and they use a control of type `range` (content-type editor control).
The range control, when rendering, validates and 'fixes' the value format (my guess is that the value format at some point was outdated, and this was made to ensure that the format was up to date. It is a pretty old change, from 2014).
So, what was happening was that the content-type model was updated with this 'fix', so it was updating the state to 'dirty', and causing the 'close' button to become 'cancel'.
Now, the detail is that the format of the range value was already the same that was being set. So the model update was unnecessary. That's the reason of the `item.properties[l].value !== value` validation. 

Still, I added an extra validation on the range content-type control, to avoid calling the `updatePropertyFn` if the value didn't change. But keeping the change in `content-type.js` will avoid updating the model with the same value in any other case.
(I did test the scenario will multiple controls, only got the image-picker to cause the issue)

ui/app/src/components/CreateSiteDialog/GitForm.tsx
jvega190: The only field that doesn't have it now is the description field, because it is a multi line text field so handling the enter would cause unwanted behavior

ui/guest/src/controls/rte.ts
jvega190: We used to change the display of the element, but there was an bug reported [here](https://github.com/craftercms/craftercms/issues/5212), and we ended up cloning the elements because it didn't matter the display style of the elements, the inline elements (the ones in the regex) have the issue, and it doesn't happen to block type elements.

The regex was incorrectly matching, that's the reason I added the `^` and `$` to exactly match those words.

ui/guest/src/controls/rte.ts
jvega190: https://github.com/craftercms/studio-ui/pull/3306#discussion_r1178234876

ui/guest/src/controls/rte.ts
jvega190: By getting the `.style` prop from an element it only retrieves the inline styles set to the element (the ones in the style prop of the element). If I want to retrieve **all** styles for an element (inline styles, and styles applied from css files, etc) I'd use
`window.getComputedStyle(element)`. The issue is that if I retrieve all of the styles, and set them to the cloned element under `.style`, all of the styles would become inline styles and have higher precedence than for example applied styles in XB, causing issues.

For that reason I copied only the inline styles, and also copied the classes of the element (a couple of lines below).

ui/guest/src/controls/rte.ts
rart: What was it matching incorrectly?

Not sure about the performance but seems more readable to do: `/^(B|BIG|I|...)$/`. Please test.

ui/guest/src/controls/rte.ts
rart: Add link to that ticket on the comment pls. Feels like something to revisit.

ui/guest/src/controls/rte.ts
rart: Ok. That 👆🏻 should be the code comment, just clean it up slightly for conciseness, but it's ok to write multiline comments that better explain things.

rart (changes requested):

rart:
Still a draft? What's missing?

rart (approved):

PR #3313, opened 2023-05-10
[6044] There is no icon guide in the new dashboard

jvega190:
https://github.com/craftercms/craftercms/issues/6044


ui/app/src/components/SiteDashboard/SiteDashboard.tsx
```diff
@@ -143,6 +144,17 @@ export function Dashboard(props: DashboardProps) {
...
+        <Grid
+          item
+          md={12}
+          sx={{
+            display: desktopScreen ? 'flex' : 'none',
+            flexWrap: 'wrap'
+          }}
+        >
+          <IconGuideDashlet />
+        </Grid>
```
rart: Wrap both of these in an `if` based on `desktopScreen` so they only render once on their respective screen size.

rart (changes requested):

rart (approved):

PR #3321, opened 2023-05-17
[6068] When an user updates the page-url the sidebar should refresh his data

jvega190:
https://github.com/craftercms/craftercms/issues/6068

### DIFF
--- a/ui/app/src/state/epics/pathNavigator.ts
+++ b/ui/app/src/state/epics/pathNavigator.ts
@@ -23,7 +23,7 @@ import {
   fetchItemsByPath,
   fetchItemWithChildrenByPath
 } from '../../services/content';
-import { getIndividualPaths, getParentPath, getRootPath, withIndex } from '../../utils/path';
+import { getIndividualPaths, getParentPath, getRootPath, withIndex, withoutIndex } from '../../utils/path';
 import { forkJoin, Observable } from 'rxjs';
 import {
   pathNavigatorBackgroundRefresh,
@@ -399,7 +399,10 @@ export default [
             actions.push(pathNavigatorRefresh({ id: navigator.id }));
           } else if (!navigator.isRootPathMissing && navigator.currentPath.startsWith(sourcePath)) {
             actions.push(pathNavigatorSetCurrentPath({ id: navigator.id, path: navigator.rootPath }));
-          } else if (navigator.currentPath === parentOfTargetPath || navigator.currentPath === parentOfSourcePath) {
+          } else if (
+            withoutIndex(navigator.currentPath) === parentOfTargetPath ||
+            withoutIndex(navigator.currentPath) === parentOfSourcePath
+          ) {
             actions.push(pathNavigatorBackgroundRefresh({ id: navigator.id }));
           }
         });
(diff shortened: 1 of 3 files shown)


ui/app/src/state/epics/pathNavigator.ts
```diff
@@ -399,7 +399,10 @@ export default [
             actions.push(pathNavigatorRefresh({ id: navigator.id }));
           } else if (!navigator.isRootPathMissing && navigator.currentPath.startsWith(sourcePath)) {
             actions.push(pathNavigatorSetCurrentPath({ id: navigator.id, path: navigator.rootPath }));
-          } else if (navigator.currentPath === parentOfTargetPath || navigator.currentPath === parentOfSourcePath) {
+          } else if (
+            withoutIndex(navigator.currentPath) === parentOfTargetPath ||
+            withoutIndex(navigator.currentPath) === parentOfSourcePath
```
rart, line 404: What's the logic?

Would you need to equalise the `parentOf*` to also strip index for this comparison. Make sure you're not looking at the issue through a straw.

ui/app/src/state/epics/pathNavigatorTree.ts
```diff
@@ -338,14 +338,18 @@ export default [
             [parentPathOfTargetPath, parentPathOfSourcePath].forEach((path) => {
               if (
                 // If in totalByPath is an item that has been loaded and must update...
-                path in tree.totalByPath
+                // path in totalByPath may be a page, and its path has index.xml, so it needs to be validated too.
+                path in tree.totalByPath ||
+                withIndex(path) in tree.totalByPath
```
rart, line 343: Does it make sense to assure to comparison is made both ways?
```suggestion
                withoutIndex(path) in tree.totalByPath ||
                withIndex(path) in tree.totalByPath
```

ui/app/src/state/epics/pathNavigatorTree.ts
```diff
@@ -338,14 +338,18 @@ export default [
...
+                // Get correct path to fetch (may include index.xml)
+                const fetchPath = path in tree.totalByPath ? path : withIndex(path);
                 // If its children are loaded, then re-fetch to get the new
-                tree.childrenByParentPath[path] &&
+                tree.childrenByParentPath[fetchPath] &&
                   actions.push(
                     pathNavigatorTreeFetchPathChildren({
                       id: id,
-                      path: path,
+                      path: fetchPath,
```
rart, line 352: Please explain changes

rart (commented):
Please explain what the issue is/was and how these changes address.

ui/app/src/state/epics/pathNavigator.ts
jvega190: `parentOf*` is retrieved by using `getParentPath` that already uses `withoutIndex`. A refresh of the pathNavigator is done when the currentPath is the same as the parent of the target or source. The issue was that the current path was matching one of the parents, but not entiretly due to the /index.xml being present in the currentPath

ui/app/src/state/epics/pathNavigatorTree.ts
jvega190: `path` is one of the parents (parentPathofTargetPath or parentPathOfSourcePath), and they're retrieved using `getParentPath`. This is the same scenario as the one above, where `getParentPath` already removes the index.xml from the path

rart (approved):

PR #3325, opened 2023-05-19
[6028] XB - Inherited fields issues

jvega190:
https://github.com/craftercms/craftercms/issues/6028

### DIFF
--- a/ui/guest/src/react/GuestProxy.tsx
+++ b/ui/guest/src/react/GuestProxy.tsx
@@ -433,15 +433,16 @@ export function GuestProxy() {
         }
         case updateFieldValueOperation.type:
           let { modelId, fieldId, index, value } = op.payload;
-          modelId = getModelIdFromInheritedField(modelId, fieldId);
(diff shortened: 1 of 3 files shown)


ui/guest/src/react/GuestProxy.tsx
```diff
@@ -433,15 +433,16 @@ export function GuestProxy() {
         }
         case updateFieldValueOperation.type:
           let { modelId, fieldId, index, value } = op.payload;
-          modelId = getModelIdFromInheritedField(modelId, fieldId);
+          const modelIdToEdit = getModelIdFromInheritedField(modelId, fieldId);
```
rart, line 436: What if it's not inherited?

rart (approved):

ui/guest/src/react/GuestProxy.tsx
jvega190: getModelIdFromInheritedField returns modelId if not inherited

rart (approved):

PR #3326, opened 2023-05-19
[6070] Update XB compatibility checker

jvega190:
https://github.com/craftercms/craftercms/issues/6070

### DIFF
--- a/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
+++ b/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
@@ -328,7 +328,6 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
   const [dataSourceActionsListState, setDataSourceActionsListState] = useSpreadState<DataSourcesActionsListProps>(
     dataSourceActionsListInitialState
   );
-  const env = useEnv();
   const conditionallyToggleEditMode = (nextHighlightMode?: HighlightMode) => {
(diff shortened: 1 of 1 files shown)


ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -563,10 +563,12 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
...
         case guestCheckIn.type:
-          const { version: guestVersion } = payload;
+          const { version: guestFullVersion } = payload;
           const studioVersion = env.version.slice(0, 5);
-
-          if (type === guestCheckIn.type && guestVersion?.slice(0, 5) !== studioVersion) {
+          const guestVersion = guestFullVersion.slice(0, 5);
+          const isCompatibleVersion =
+            guestVersion === studioVersion || (studioVersion === '4.1.0' && guestVersion === '4.0.3');
+          if (type === guestCheckIn.type && !isCompatibleVersion) {
```
rart: How about we simply do: 
```suggestion
          if (type === guestCheckIn.type && ['4.1.0', '4.0.3'].includes(payload.version.slice(0, 5));
```
Don't even need studio version as things stand do we?

rart (changes requested):

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
jvega190: We don't need it for now

rart (approved):

PR #3317, opened 2023-05-12
[6054] Clear dashlet options after performing actions in new Dashboard

jvega190:
https://github.com/craftercms/craftercms/issues/6054

PR #3328, opened 2023-05-23
[6058] Facilitate Item Actions in New Dashboards

jvega190:
https://github.com/craftercms/craftercms/issues/6058


jvega190:
There may be conflicts with https://github.com/craftercms/studio-ui/pull/3317, this will be re-visited after 3317 is merged

rart (approved):

PR #3343, opened 2023-06-01
[6084] Publish dialog accessed from context menu in a dashlet does not display the selected date/time request

jvega190:
https://github.com/craftercms/craftercms/issues/6084

### DIFF
--- a/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
+++ b/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
@@ -84,7 +84,7 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
     };
 
     if (detailedItems) {
-      let itemsChecked = detailedItems.filter((item) => selectedItems[item.path]);
+      const itemsChecked = detailedItems.flatMap((item) => (selectedItems[item.path] ? [item] : []));
 
       if (itemsChecked.length === 0) {
         state.publishingTarget = '';
--- a/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
+++ b/ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
@@ -83,68 +83,70 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
       publishingTarget: null
     };
 
-    let itemsChecked = items.filter((item) => selectedItems[item.path]);
+    if (detailedItems) {
+      let itemsChecked = detailedItems.filter((item) => selectedItems[item.path]);
 
-    if (itemsChecked.length === 0) {
-      state.publishingTarget = '';
-      return state;
-    }
-
-    // region Discover mixed targets and/or schedules and sets the publishingTarget based off the items
-    let target: string;
-    let schedule: string;
-    itemsChecked.some((item, index) => {
-      const computedTarget = getComputedPublishingTarget(itemsChecked[0]);
-      const computedSchedule = getDateScheduled(itemsChecked[0]);
-      if (index === 0) {
-        target = computedTarget;
-        schedule = computedSchedule;
-      } else {
-        if (target !== computedTarget) {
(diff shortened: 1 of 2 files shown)


ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -83,68 +83,70 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
       publishingTarget: null
     };
 
-    let itemsChecked = items.filter((item) => selectedItems[item.path]);
+    if (detailedItems) {
```
jvega190, line 86: This section is a memo that used the items (Dialog prop) to set the scheduled value (and some other values for the component) that is used for the now/custom setting. We already fetch the detailedItems in this component, so now the calculations are done using them, instead of using the items prop (that may be incomplete Sandbox items).

rart: Make sure all the logic still applies correctly when there are mixed items (e.g with and without schedule)

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -83,68 +83,70 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
+      let itemsChecked = detailedItems.filter((item) => selectedItems[item.path]);
```
rart: Wouldn't iterating selectedItems be more efficient? ie. less iterations, same result?
```suggestion
      let itemsChecked = selectedItems.map((item) => detailedItems[item.path]);
```
What's the relation of existence of items in one or the other?

rart (changes requested):

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
jvega190: selectedItems is a `LookupTable<boolean>` of the selected items in the dialog. Detailed items is an array of the items in the dialog. I could make a change so we have a Lookup of detailed items instead of an array (it would need some changes because the array is expected), the do what you suggested:
`Object.keys(selectedItems).map((path) => detailedItems[path]);`

jvega190: Also a filtering of the selected items lookup would be needed, since the value may be true/false:
`Object.keys(selectedItems).filter((path) => selectedItems[path]).map((path) => detailedItems[path]);`

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -83,68 +84,72 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
       publishingTarget: null
     };
 
-    let itemsChecked = items.filter((item) => selectedItems[item.path]);
+    if (detailedItems) {
+      const itemsChecked = Object.keys(selectedItems)
+        .filter((path) => selectedItems[path])
+        .map((itemPath) => detailedItems[itemPath]);
```
rart:
```suggestion
      const itemsChecked = Object.keys(selectedItems)
        .flatMap((itemPath) => selectedItems[itemPath] ? [detailedItems[itemPath]] : []);
```

rart (changes requested):

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
jvega190: Per conversation, rolled back the change to lookup table of detailed items and a map to just use a flatMap instead of the filter of selectedItems

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
jvega190: Tested and it does work with mixed items

rart (approved):

PR #3293, opened 2023-03-31
[502E] Deps update

jvega190:
https://github.com/craftersoftware/craftercms/issues/502

PR #3356, opened 2023-06-13
[6079] Remove all jQuery usage from guest & host

jvega190:
https://github.com/craftercms/craftercms/issues/6079


ui/guest/src/react/GuestProxy.tsx
```diff
@@ -196,23 +199,22 @@ export function GuestProxy() {
...
-      .on('mouseleave', '[data-craftercms-model-id]', handler)
-      .on('dragstart', '[data-craftercms-model-id]', handler)
-      .on('dragover', '[data-craftercms-model-id]', handler)
-      .on('dragleave', '[data-craftercms-model-id]', handler)
-      .on('drop', '[data-craftercms-model-id]', handler)
-      .on('dragend', '[data-craftercms-model-id]', handler)
-      .on('click', '[data-craftercms-model-id]', handler)
-      .on('dblclick', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'mouseover', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'mouseout', '[data-craftercms-model-id]', handler);
```
jvega190: The javascript approach uses event bubbling, so we need to use `mouseout` instead of `mouseleave` since mouseleave does not bubble. `mouseout` is fired when the pointer leaves the element *or* leaves of the element's descendants, and taking that into consideration helps achieving event delegation

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -180,13 +179,17 @@ export function GuestProxy() {
...
+    const insertElement = (
+      element: Element | HTMLElement,
+      daddy: Element | HTMLElement,
+      targetIndex: string | number
+    ): void => {
       const index = typeof targetIndex === 'string' ? parseInt(popPiece(targetIndex)) : targetIndex;
-      const $siblings = $daddy.find('> *');
-      if ($siblings.length === index) {
-        $daddy.append($element);
+      const siblings = daddy.querySelectorAll('*');
```
rart: Not an equivalent selection (missing >). Triple check for similar instances of bad selector translation. 

Here we're looking for direct descendants, so .children might be sufficient and faster. Check.

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -196,23 +199,22 @@ export function GuestProxy() {
         document.querySelectorAll<HTMLElement>('[data-craftercms-model-id]').forEach(registerElement);
       });
 
-    const handler: JQuery.EventHandlerBase<any, any> = (e: Event): void => {
-      let record = ElementRegistry.fromElement(e.currentTarget as Element);
+    const handler: JQuery.EventHandlerBase<any, any> = (e: Event, target): void => {
```
rart: Fix types. This should no longer be a JQuery.Event.
This type should be giving some sort of typescript error. Otherwise, there may still be some jquery typings or something on the env.

Make sure you're running yarn build:modern to detect typescript errors

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -196,23 +199,22 @@ export function GuestProxy() {
...
-      .on('dblclick', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'mouseover', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'mouseout', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'dragstart', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'dragover', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'dragleave', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'drop', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'dragend', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'click', '[data-craftercms-model-id]', handler);
+    delegateEventListener(document, 'dblclick', '[data-craftercms-model-id]', handler);
```
rart: All these delegates require the most intensive testing. Need to test over and over in all bps and from different angles.

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -232,17 +234,17 @@ export function GuestProxy() {
...
           const phyRecord = ElementRegistry.fromICEId(iceId);
-          const $el = $(phyRecord.element);
-          const $targetSibling = $el.parent().children().eq(targetIndexParsed);
+          const el = phyRecord.element;
+          const targetSibling = el.parentElement.children[targetIndexParsed];
 
           // Move...
           if (currentIndexParsed < targetIndexParsed) {
-            $el.insertAfter($targetSibling);
+            el.parentElement.insertBefore(el, targetSibling.nextSibling);
```
rart, line 247: Make sure this doesn't result in out of bounds or null when using .nextSibling on target element. Could the target element be the last?

rart, line 247: Make sure this doesn't result in out of bounds or null when using .nextSibling on target element. Could the target element be the last child?

ui/guest/src/react/GuestProxy.tsx
```diff
@@ -288,29 +290,31 @@ export function GuestProxy() {
...
-              .each((i, elem) => {
-                ElementRegistry.deregister(ElementRegistry.fromElement(elem).id);
-                $(elem).attr('data-craftercms-model-id', record.modelId);
-                $(elem).attr('data-craftercms-field-id', record.fieldId);
-                $(elem).attr('data-craftercms-index', generateIndex(newIndex, i));
-
-                registerElement(elem);
-              });
+
+            record.element.querySelectorAll('[data-craftercms-index]').forEach((elem, i) => {
```
rart: Missing >

rart (changes requested):
This entire PR requires very systematic testing of **every feature**.

ui/guest/src/react/GuestProxy.tsx
jvega190: the reason why there's a @types/jquery is because it is a dep of @types/tinymce@npm. Types fixed

ui/guest/src/react/GuestProxy.tsx
jvega190: For some reason when the element target was the last position it didn't fail (sorting was performed properly). But I added a different case for when the target is the last position, now it uses `appendChild`

ui/guest/src/utils/dom.ts
```diff
@@ -400,11 +393,68 @@ export function updateDropZoneValidations(
...
+  let iterate = true;
+  const propagateEvents = ['dragover'];
+
+  while (iterate) {
+    if (currentNode.matches?.(selector)) {
+      nodes.push(currentNode);
+      currentNode = currentNode.parentNode;
+      if (!propagateEvents.includes(event.type)) {
+        iterate = false;
+      }
```
rart:
```suggestion
  while (iterate) {
    if (currentNode.matches?.(selector)) {
      nodes.push(currentNode);
      currentNode = currentNode.parentNode;
      // Explain why dragover is different...
      iterate = event.type === 'dragover';
```

ui/guest/src/utils/dom.ts
```diff
@@ -400,11 +393,68 @@ export function updateDropZoneValidations(
...
+    } else {
+      iterate = false;
+    }
+  }
+  return nodes;
+}
+
+export function delegateEventListener(element, eName, selector, fn) {
+  element.addEventListener(eName, function (event) {
+    const nodes = searchSelector(event, selector, event.currentTarget);
```
rart: Not a great name. Also not used for anything else. Let's just merge the code here.

ui/guest/src/utils/dom.ts
```diff
@@ -400,11 +393,68 @@ export function updateDropZoneValidations(
...
+    if (nodes.length) {
+      // Execute the callback with the context set to the found element
+      // jQuery goes way further, it even has its own event object
```
rart: What's the point of this?

ui/guest/src/utils/dom.ts
```diff
@@ -400,11 +393,68 @@ export function updateDropZoneValidations(
...
+    } else if (currentNode !== stopAt && currentNode !== document.body) {
+      currentNode = currentNode.parentNode;
+    } else {
+      iterate = false;
+    }
+  }
+  return nodes;
+}
+
+export function delegateEventListener(element, eName, selector, fn) {
```
rart:
```suggestion
export function addDelegateEventListener(element, eName, selector, fn) {
```

ui/guest/src/react/GuestProxy.tsx
rart: Can we identify how event abstraction libraries (jQuery, React, etc) do it? I worry how this might affect things.

ui/guest/src/react/hooks.ts
```diff
@@ -61,7 +61,7 @@ interface UseModelProps extends Omit<ICEProps, 'modelId'> {
 
 const handlerMap = {
   mouseover: 'onMouseOver',
-  mouseleave: 'onMouseLeave',
+  mouseout: 'onMouseout',
```
rart: Maybe we can keep the react layer with mouseleave and handle both (mouseout & mouseleave) on epics and reducers.

ui/guest/src/store/reducers/root.ts
```diff
@@ -176,9 +178,14 @@ const reducer = createReducer(initialState, {
...
-    const { record } = action.payload;
+    let { record } = action.payload;
     // onMouseOver pre-populates the draggable record
-    const iceId = state.draggable?.[record.id];
+    let iceId = state.draggable?.[record.id];
+    if (nou(iceId) && !isSimple(record.fieldId[0])) {
+      const movableElement = record.element.closest('[draggable="true"]');
+      record = ElementRegistry.fromElement(movableElement) ?? record;
+      iceId = state.draggable?.[record.id];
+    }
```
rart: What's this handling? Add comment.

rart (changes requested):

ui/guest/src/react/GuestProxy.tsx
jvega190: Same as https://github.com/craftercms/studio-ui/pull/3356#discussion_r1240436667

sumerjabri:
Targeting 4.2.0

jvega190:
Holding this PR until https://github.com/craftercms/studio-ui/pull/3293 is merged

ui/guest/src/jquery/README.md
rart, line 1: Could we run these process as part of the build and not commit the build?

ui/guest/src/store/epics/root.ts
```diff
@@ -152,7 +151,7 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
             e.dataTransfer.setData('text/plain', `${record.id}`);
             e.dataTransfer.setDragImage(document.querySelector('.craftercms-dragged-element'), 20, 20);
           }
-          $('html').addClass(dragAndDropActiveClass);
+          document.querySelector('html').classList.add(dragAndDropActiveClass);
```
rart: Perhaps easier/more performant to do `document.documentElement`. Change all that apply. Similarly, if you're querying for the body somewhere, use `document.body`.

rart (changes requested):

ui/guest/src/jquery/README.md
rart: Maybe this whole thing should rather be on the npm-content package

jvega190: The reason why the build is committed is because it's a custom build with only the core modules of jquery. The README file explains the command used for the build

rart: That doesn't answer my question. The npm content has custom stuff and runs scripts. Is there a reason we can't move and consume from there? It'd be more or less like the `uppy` package — which is built and consumed by the app.

jvega190: You're right, I'll move it to npm-content

ui/guest/src/store/epics/root.ts
```diff
@@ -124,7 +123,7 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
       ofType('mouseover', 'mouseleave'),
       withLatestFrom(state$),
       filter((args) => args[1].status === EditingStatus.LISTENING),
-      tap(([action]) => action.payload.event.stopPropagation()),
+      tap(([action]) => action.payload.event.stopPropagation?.()),
```
rart: Why would an event not have the stopProp?

rart (commented):
- To avoid requiring to publish as an npm package: 
  - Create and configure the scripts to generate and put all that machinery on `npm-content`. Add the build steps to the pom file.
  - The build should copy the unminified file only to `guest/src/jquery.js` (with the typings). These files should be ignored by git but should be copied to the npm build directory — by the npm build scripts, not the jquery build script — so that we publish it with the package.

ui/guest/src/store/epics/root.ts
jvega190: It probably was a wrong type at some moment, removing the optional chaining

jvega190:
> * To avoid requiring to publish as an npm package:
>   
>   * Create and configure the scripts to generate and put all that machinery on `npm-content`. Add the build steps to the pom file.
>   * The build should copy the unminified file only to `guest/src/jquery.js` (with the typings). These files should be ignored by git but should be copied to the npm build directory — by the npm build scripts, not the jquery build script — so that we publish it with the package.

Build steps already in pom file since it's the same npm-content npm build script

ui/guest/package.json
```diff
@@ -119,7 +118,6 @@
     "uuid": "^9.0.1"
   },
   "devDependencies": {
-    "@types/jquery": "^3.5.25",
```
rart, line 122: Don't we still need the typings for typescript's sake?

ui/npm-content/build.sh
```diff
@@ -58,3 +58,41 @@ rm -rf "$libsDirectory/tinymce"
 cp -r ../../node_modules/tinymce "$libsDirectory/tinymce"
 
 echo "TinyMCE build complete"
+
+# Monaco
+yarn build:monaco
+echo "Monaco editor build complete"
+
+# jquery
```
rart, line 66: Please test the `build-if-changed` is working and configured correctly.

yarn.lock
```diff
@@ -1604,6 +1604,7 @@ __metadata:
     "@craftercms/classes": "npm:4.1.2"
     "@craftercms/content": "npm:4.1.2"
     "@craftercms/ice": "npm:4.1.2-1"
+    "@craftercms/npm-content": "workspace:^"
```
rart: Looks like you need to re-run `yarn`; this seems outdated.

rart (changes requested):
Confirm:
- [ ] This builds with gradlew from scratch; in a clean copy out of your dev env
- [ ] The `build:npm` script copies the jquery stuff on the `build_tsc` (where we publish the package to npm from)
- [ ] The bps that use XB via npm work well

jvega190:
@rart adding in a new comment since I can't edit the check list in your comment

- [x]  This builds with gradlew from scratch; in a clean copy out of your dev env
- [x] The build:npm script copies the jquery stuff on the build_tsc (where we publish the package to npm from)
- [x] The bps that use XB via npm work well

rart (approved):

jvega190:
@rart added the config for `build-if-changed` in npm-content. Tested in local dev and in studio installation and all good

rart (approved):

PR #3358, opened 2023-06-16
[6121] UI error when trying to create a project with only create_site permission enabled

jvega190:
https://github.com/craftercms/craftercms/issues/6121


ui/app/src/components/CreateSiteDialog/CreateSiteDialogContainer.tsx
```diff
@@ -459,6 +462,10 @@ export function CreateSiteDialogContainer(props: CreateSiteDialogContainerProps)
     setDialog({ inProgress: false });
   }
 
+  useMount(() => {
+    hasGlobalPermission('list_plugins').subscribe(setHasListPluginPermission);
+  });
```
rart: Can we move this check to the effect that fetches the marketplace bps? That way we also remove the `hasListPluginPermission` dependency from it.

rart (changes requested):

rart (approved):

PR #3360, opened 2023-06-21
[6135] Recoloring Icons in the Left Nav Tree does not work in new versions

jvega190:
https://github.com/craftercms/craftercms/issues/6135


rart (changes requested):
- Does the PathNavTree need any update?
- Rollback SystemIcon and do the updates on its usage on the PathNav so we don't unintentionally affect **everything** using SystemIcon

ui/app/src/components/SystemIcon/SystemIcon.tsx
```diff
@@ -48,7 +48,7 @@ export function SystemIcon(props: SystemIconProps) {
       </Suspense>
     ) : (
       <Tooltip title={`Icon ${icon.id} not found. Check config.`}>
-        <ErrorRounded sx={combinedSx} />
+        <ErrorRounded sx={combinedSx} className={iconClassName} />
```
jvega190, line 51: Added className in here to apply classes in non-existing icon scenario

jvega190:
@rart no changes needed in PathNavTree since it uses the same `PathNavigatorHeader`

rart (approved):

PR #3363, opened 2023-06-27
[6140] Browse Files Dialog not working properly with pages

jvega190:
https://github.com/craftercms/craftercms/issues/6140

### DIFF
--- a/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
+++ b/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
@@ -73,7 +73,7 @@ export function BrowseFilesDialogContainer(props: BrowseFilesDialogContainerProp
 
   const fetchItems = useCallback(
     () =>
-      search(site, { ...searchParameters, path: `${currentPath}/([^/]+|[^/]+/index.xml)` }).subscribe((response) => {
+      search(site, { ...searchParameters, path: `${currentPath}/[^/]+(/index.xml)?` }).subscribe((response) => {
         setTotal(response.total);
         setItems(response.items);
         setSortKeys(response.facets.map((facet) => facet.name));
--- a/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
+++ b/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
@@ -73,7 +73,7 @@ export function BrowseFilesDialogContainer(props: BrowseFilesDialogContainerProp
 
   const fetchItems = useCallback(
     () =>
-      search(site, { ...searchParameters, path: `${currentPath}/[^/]+` }).subscribe((response) => {
+      search(site, { ...searchParameters, path: `${currentPath}/([^/]+|[^/]+/index.xml)` }).subscribe((response) => {
         setTotal(response.total);
         setItems(response.items);
         setSortKeys(response.facets.map((facet) => facet.name));
--- a/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
+++ b/ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
@@ -73,7 +73,7 @@ export function BrowseFilesDialogContainer(props: BrowseFilesDialogContainerProp
 
   const fetchItems = useCallback(
     () =>
-      search(site, { ...searchParameters, path: `${currentPath}/[^/]+(/index.xml)?` }).subscribe((response) => {
+      search(site, { ...searchParameters, path: `${currentPath}/[^/]+(/index\\.xml)?` }).subscribe((response) => {
         setTotal(response.total);
         setItems(response.items);
         setSortKeys(response.facets.map((facet) => facet.name));
--- a/ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
+++ b/ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
@@ -24,8 +24,6 @@ import useActiveUser from '../../hooks/useActiveUser';
 import { useDispatch } from 'react-redux';
 import { pathNavigatorTreeExpandPath, pathNavigatorTreeFetchPathChildren } from '../../state/actions/pathNavigatorTree';
 import { getIndividualPaths, withIndex } from '../../utils/path';
-import { checkPathExistence } from '../../services/content';
-import { map } from 'rxjs/operators';
 import { forkJoin, of } from 'rxjs';
 import { batchActions } from '../../state/actions/misc';
 import useSelection from '../../hooks/useSelection';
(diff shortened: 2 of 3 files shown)


ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
```diff
@@ -73,13 +73,7 @@ export function FolderBrowserTreeView(props: FolderBrowserTreeViewProps) {
                     path: withIndexXml in refs.current.tree.childrenByParentPath ? withIndexXml : p
                   })
                 )
-              : checkPathExistence(siteId, withIndexXml).pipe(
```
jvega190, line 76: Need to discuss this a bit

ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
```diff
@@ -73,13 +71,7 @@ export function FolderBrowserTreeView(props: FolderBrowserTreeViewProps) {
...
-                  map((exists) =>
-                    exists
-                      ? pathNavigatorTreeFetchPathChildren({ id, path: withIndexXml, expand: true })
-                      : pathNavigatorTreeFetchPathChildren({ id, path: p, expand: true })
-                  )
-                );
+              : of(pathNavigatorTreeFetchPathChildren({ id, path: p, expand: true }));
```
jvega190, line 74: The FolderBrowserTreeView uses paths without index, so when fetching/expanding using withIndex it was causing an issue when expanding the root path of the dialog.

rart: You took away the path existence checking, though. What's that about?

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
```diff
@@ -332,7 +333,7 @@ export function PathNavigatorTreeItem(props: PathNavigatorTreeItemProps) {
                   minHeight: '23.5px'
                 }
               }}
-              item={itemsByPath[path]}
+              item={itemsByPath[path] ?? itemsByPath[withIndex(path)]}
```
rart: use the little util we have that checks both ways (called something like lookupItemByPath)

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
```diff
@@ -198,7 +199,7 @@ export function PathNavigatorTreeItem(props: PathNavigatorTreeItemProps) {
   const [showFilter, setShowFilter] = useState(Boolean(keywordByPath[path]));
   const [keyword, setKeyword] = useState(keywordByPath[path] ?? '');
   const { formatMessage } = useIntl();
-  const children = childrenByParentPath[path] ?? [];
+  const children = childrenByParentPath[path] ?? childrenByParentPath[withIndex(path)] ?? [];
```
rart: use little util we have that checks both ways (called something like lookupItemByPath)

ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
```diff
@@ -73,7 +73,7 @@ export function BrowseFilesDialogContainer(props: BrowseFilesDialogContainerProp
 
   const fetchItems = useCallback(
     () =>
-      search(site, { ...searchParameters, path: `${currentPath}/[^/]+` }).subscribe((response) => {
+      search(site, { ...searchParameters, path: `${currentPath}/([^/]+|[^/]+/index.xml)` }).subscribe((response) => {
```
rart: Regexp looks strange. Not too sure how this regex works/what it tells the service to do; what exactly is the `path` argument to the search service? The `index.xml` part seems almost redundant.
```suggestion
      search(site, { ...searchParameters, path: `${currentPath}/[^/]+(index\.xml)?` }).subscribe((response) => {
```

rart (changes requested):

ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
jvega190: The path existence check was to see if sending withIndex or not, other than that the calls are the same, so that's why I removed it

ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
jvega190: Updated the regex. The idea is to retrieve everything under `currentPath` with no `/` (to avoid getting files from subfolders), **but** consider pages, so files with name `index.xml` under a folder should match too. Used your suggestion and only added an `/` before `index.xml` (IDE suggested to remove escaping chars like the one before the dot in index.xml and the regex worked)

ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
rart: So how does it work now?

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
rart: Why is this resolved? Wasn't changed.

ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
rart: `.` means anything so not sure if it's working because of that or because escaping is actually not needed.

jvega190: Yeah it was actually considering index{anyChar}xml, updated for it to be `${currentPath}/[^/]+(/index\\.xml)?`. That way it does return the results the way be want (first level children and `{folder}/index.xml`)

ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
jvega190: It used to be:
```
checkPathExistence => if exists fetchPathChildren using withIndex, if not fetchPathChildren with path (without withIndex)
```
now it would fetchPathChildren withouth `withIndex` in both cases, so checkPathExistence is unnecessary

ui/app/src/components/BrowseFilesDialog/BrowseFilesDialogContainer.tsx
```diff
@@ -73,7 +73,7 @@ export function BrowseFilesDialogContainer(props: BrowseFilesDialogContainerProp
 
   const fetchItems = useCallback(
     () =>
-      search(site, { ...searchParameters, path: `${currentPath}/[^/]+` }).subscribe((response) => {
+      search(site, { ...searchParameters, path: `${currentPath}/[^/]+(/index\\.xml)?` }).subscribe((response) => {
```
rart, line 76: This still doesn't make sense to me. Ping me to discuss.

rart (changes requested):

ui/app/src/components/FolderBrowserTreeView/FolderBrowserTreeView.tsx
rart: But what if it's a folder? Like `articles` in editorial?

rart (approved):

PR #3364, opened 2023-06-27
Hide SearchAhead when typing/clicking enter in PreviewAddressBar

jvega190:


ui/app/src/components/PreviewAddressBar/PreviewAddressBar.tsx
```diff
@@ -134,6 +134,7 @@ export function PreviewAddressBar(props: AddressBarProps) {
   };
 
   const onUrlChange = (url: string) => {
+    setFocus(false);
```
rart: Does this take the actual focus away from the field? What if people keep typing again after having pressed enter?

jvega190: Per conversation the behavior will be updated in the PagesSearchAhead component instead of the PreviewAddressBar

rart (approved):

PR #3366, opened 2023-07-03
[6102] XB assumes file based items are components

jvega190:
https://github.com/craftercms/craftercms/issues/6102

### DIFF
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -487,9 +487,6 @@ function parseElementByContentType(
         items.forEach((item) => {
           let path = getInnerHtml(item.querySelector(':scope > include'));
           const component = item.querySelector(':scope > component');
-          if (!path && !component) {
-            path = getInnerHtml(item.querySelector(':scope > key'));
-          }
           const itemKey = getInnerHtml(item.querySelector(':scope > key'));
           if (!path && !component) {
             path = itemKey;
@@ -618,7 +615,7 @@ export function createModelHierarchyDescriptorMap(
             source[field.id]
               // Just as controllers are not included in HierarchyDescriptor, files inside a node-selector are not included either.
               // (files and controllers are stored as a key/value object)
-              .filter((component) => typeof component === 'string')
+              .filter((componentId) => typeof componentId === 'string')
               .forEach((componentId, index) => {
                 lookup[currentModelId].children.push(componentId);
                 if (lookup[componentId]) {
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -491,6 +491,9 @@ function parseElementByContentType(
             path = getInnerHtml(item.querySelector(':scope > key'));
           }
           const itemKey = getInnerHtml(item.querySelector(':scope > key'));
+          if (!path && !component) {
+            path = itemKey;
+          }
           // Embedded components have no .xml in their key
           const isFile = Boolean(itemKey) && !itemKey.endsWith('.xml') && Boolean(!item.getAttribute('inline'));
           if (isFile) {
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -491,13 +491,22 @@ function parseElementByContentType(
             // TODO: Groovy Controller Issue;
             path = getInnerHtml(item.querySelector(':scope > key'));
           }
-          const instance = parseContentXML(
-            component ? wrapElementInAuxDocument(component) : null,
-            path,
-            contentTypesLookup,
-            instanceLookup
-          );
-          array.push(instance);
+          const itemKey = getInnerHtml(item.querySelector(':scope > key'));
+          const isFile = !itemKey.endsWith('.xml');
+          if (isFile) {
+            array.push({
+              key: itemKey,
+              value: getInnerHtml(item.querySelector(':scope > value'))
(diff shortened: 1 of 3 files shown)


ui/app/src/utils/content.ts
```diff
@@ -466,13 +466,23 @@ function parseElementByContentType(
...
             path = getInnerHtml(item.querySelector(':scope > key'));
           }
-          const instance = parseContentXML(
-            component ? wrapElementInAuxDocument(component) : null,
-            path,
-            contentTypesLookup,
-            instanceLookup
-          );
-          array.push(instance);
+          const itemKey = getInnerHtml(item.querySelector(':scope > key'));
```
rart:
```ts
const itemKey = getInnerHtml(item.querySelector(':scope > key'));
if (!path && !component) {
  path = itemKey
}
```
(You can take out that TODO; it doesn't really say much and this sort of addresses)

ui/app/src/utils/content.ts
```diff
@@ -466,13 +466,23 @@ function parseElementByContentType(
...
+          // Embedded components have no .xml in their key
+          const isFile = Boolean(itemKey) && !itemKey.endsWith('.xml') && Boolean(!item.getAttribute('inline'));
```
rart, line 495: @russdanner @sumerjabri FYI this is the kind of juggling we have to do because of the mixed datasource (item selector used for files, assets, code, AND content)

rart (changes requested):

ui/app/src/utils/content.ts
```diff
@@ -488,16 +488,28 @@ function parseElementByContentType(
           let path = getInnerHtml(item.querySelector(':scope > include'));
           const component = item.querySelector(':scope > component');
           if (!path && !component) {
-            // TODO: Groovy Controller Issue;
             path = getInnerHtml(item.querySelector(':scope > key'));
           }
```
rart: This if is a duplicate of below? Remove this one.

ui/app/src/utils/content.ts
```diff
@@ -603,34 +615,38 @@ export function createModelHierarchyDescriptorMap(
...
-                      `Model ${component} was found in multiple parents (${lookup[component].parentId} and ${model.craftercms.id}). ` +
-                        `Same model twice on a single page may have unexpected behaviours for in-context editing.`,
-                      // @ts-ignore
-                      typeof component !== 'string' && component
-                    ].filter(Boolean)
-                  );
+            source[field.id]
+              // Just as controllers are not included in HierarchyDescriptor, files inside a node-selector are not included either.
+              // (files and controllers are stored as a key/value object)
+              .filter((component) => typeof component === 'string')
```
rart:
```suggestion
              .filter((componentId) => typeof component === 'string')
```

ui/guest/src/iceRegistry.ts
```diff
@@ -175,11 +175,14 @@ export function getMediaDropTargets(type: string): ICERecord[] {
 export function getRecordDropTargets(id: number): ICERecord[] {
   const record = getById(id);
   const { index, field, fieldId, model } = getReferentialEntries(record);
+  // With components, the model lookup contains ids of each of the components, in cases like rep-groups and files, the model
+  // contains the actual model of the file/rep-group.
```
```suggestion
  // With components, the model lookup contains ids of each of the components, in cases like repeat groups and files, the model lookup
  // contains the actual model of the file/repeat group.
```

ui/guest/src/iceRegistry.ts
```diff
@@ -202,7 +205,7 @@ export function getRecordDropTargets(id: number): ICERecord[] {
       // move an item deeper inside itself).
       return rec.modelId === id || allChildren.includes(rec.modelId);
     });
-  } else if (field.type === 'repeat') {
+  } else if (field.type === 'repeat' || (field.type === 'node-selector' && !isModelId)) {
     return getRepeatGroupItemDropTargets(record);
```
rart, line 209: This seems confusing. How can you look the repeat group drop target for a node selector item? Doesn't make sense to me.

rart (changes requested):

ui/guest/src/iceRegistry.ts
jvega190: As explained in the comment above, when having a node-selector of files, the model lookup contains the actual model of the files (not the ids, as with components), with a `key/value` just as in repeating groups. That's the reason why in that case it gets treated just as a repeating group.

ui/app/src/utils/content.ts
```diff
@@ -487,17 +487,26 @@ function parseElementByContentType(
...
-            path = getInnerHtml(item.querySelector(':scope > key'));
+            path = itemKey;
+          }
+          // Embedded components have no .xml in their key
+          const isFile = Boolean(itemKey) && !itemKey.endsWith('.xml') && Boolean(!item.getAttribute('inline'));
+          if (isFile) {
+            array.push({
+              key: itemKey,
+              value: getInnerHtml(item.querySelector(':scope > value'))
+            });
```
rart, line 500: Check/reminder: I believe some things have more than just key & value. Might not be the case for files per se.

rart (approved):

sumerjabri:
@jvega190 please fix conflicts

jvega190:
@sumerjabri done

rart (approved):

PR #3370, opened 2023-07-12
[6152] Item treated as locked for me when sorting using XB

jvega190:
https://github.com/craftercms/craftercms/issues/6152

### DIFF
--- a/ui/app/src/state/reducers/content.ts
+++ b/ui/app/src/state/reducers/content.ts
@@ -70,7 +70,10 @@ const initialState: ContentState = {
   itemsBeingFetchedByPath: {}
 };
 
-const updateItemLockState = (state: ContentState, { path, user, locked }) => {
+const updateItemLockState = (
+  state: ContentState,
+  { targetPath: path, user, locked }: Partial<SocketEventBase & { locked: boolean }>
+) => {
   if (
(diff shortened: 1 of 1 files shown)


rart (approved):

ui/app/src/state/reducers/content.ts
```diff
@@ -84,7 +84,7 @@ const updateItemLockState = (state: ContentState, { path, username, locked }) =>
       ...state.itemsByPath,
       [path]: {
         ...state.itemsByPath[path],
-        lockOwner: locked ? username : null,
+        lockOwner: locked ? user : null,
```
rart: What is this property now? If it was a string and now an object, sounds like that could causes issues with just these changes alone. 

Also, check typings. We should be getting some form of compilation error.

rart (changes requested):

ui/app/src/state/reducers/content.ts
jvega190: This prop changed from string to a 'Person' object as part of some backend changes, it is already updated in its usages (from `item.lockOwner` to `item.lockOwner?.username`). I'll check to add a proper typing there

rart (approved):

PR #3372, opened 2023-07-13
[5953] Compatibility check against a known list of compatible versions

rart:
craftercms/craftercms#5953

### DIFF
--- a/ui/app/src/components/PreviewConcierge/compatibilityList.ts
+++ b/ui/app/src/components/PreviewConcierge/compatibilityList.ts
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2007-2023 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


ui/app/src/components/PreviewConcierge/compatibilityList.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const list = ['4.1.1', '4.1.0', '4.0.3'];
```
sumerjabri, line 17: What about 4.0.4/5/6/7 and 4.1.2?

sumerjabri (changes requested):

ui/app/src/components/PreviewConcierge/compatibilityList.ts
rart: We never published a package with those versions to npm.

This list needs updating with each release and it will grow until the compatibility breaks and the list is restarted from the first compatible release — which will be the latest release version at that point.

sumerjabri (approved):

PR #3373, opened 2023-07-13
[6155] Remove sites prop from user in state

jvega190:
https://github.com/craftercms/craftercms/issues/6155

### DIFF
--- a/ui/app/src/state/actions/system.ts
+++ b/ui/app/src/state/actions/system.ts
@@ -20,7 +20,6 @@ import { OptionsObject } from 'notistack';
 import GlobalState from '../../models/GlobalState';
 import { ObtainAuthTokenResponse } from '../../services/auth';
 import User from '../../models/User';
-import { Site } from '../../models/Site';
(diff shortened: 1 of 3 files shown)


ui/app/src/state/actions/system.ts
```diff
@@ -110,7 +109,6 @@ export const contentTypeDeleted = /*#__PURE__*/ createAction<StandardAction>('CO
 export const storeInitialized = /*#__PURE__*/ createAction<{
   auth: ObtainAuthTokenResponse;
   user: User;
-  sites: Array<Site>;
```
rart: This does go into the payload of the storeInitialized, if I'm not mistaken. It just doesn't go into the users state but it goes into sites state. Where do we fetch sites otherwise?

rart (changes requested):

rart (changes requested):
This PR got mixed up with others.

jvega190:
@rart fixed PR

rart (approved):

PR #3374, opened 2023-07-13
[5953][4.0.x] Compatibility check against a known list of compatible versions

jvega190:
https://github.com/craftercms/craftercms/issues/5953

### DIFF
--- a/ui/app/src/components/PreviewConcierge/compatibilityList.ts
+++ b/ui/app/src/components/PreviewConcierge/compatibilityList.ts
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2007-2023 Crafter Software Corporation. All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as published by
+ * the Free Software Foundation.
+ *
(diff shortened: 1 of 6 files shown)


ui/app/src/components/PreviewConcierge/compatibilityList.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const list = ['4.0.3', '4.0.4', '4.0.5', '4.0.6', '4.0.7'];
```
rart, line 17: Only current version and versions published to npm are pertinent. Also, reverse order so the least iterations happen on the latest version.
```suggestion
const list = ['4.0.7', '4.0.3'];
```

sumerjabri (approved):

PR #3377, opened 2023-07-14
[5830] Add a sort option to the sidebar path navigator component

jvega190:
https://github.com/craftercms/craftercms/issues/5830

### DIFF
--- a/ui/app/src/components/PathNavigator/PathNavigator.tsx
+++ b/ui/app/src/components/PathNavigator/PathNavigator.tsx
@@ -34,8 +34,7 @@ import {
   pathNavigatorRefresh,
   pathNavigatorSetCollapsed,
   pathNavigatorSetKeyword,
-  pathNavigatorSetLocaleCode,
-  pathNavigatorUpdate
+  pathNavigatorSetLocaleCode
 } from '../../state/actions/pathNavigator';
 import { showEditDialog, showItemMegaMenu, showPreviewDialog } from '../../state/actions/dialogs';
 import {
@@ -67,7 +66,6 @@ import { getStoredPathNavigator } from '../../utils/state';
 import { useActiveSite } from '../../hooks/useActiveSite';
 import { useActiveUser } from '../../hooks/useActiveUser';
 import { GetChildrenOptions } from '../../models';
-import { batchActions } from '../../state/actions/misc';
 
 interface Menu {
   path?: string;
--- a/ui/app/src/components/PathNavigator/PathNavigator.tsx
+++ b/ui/app/src/components/PathNavigator/PathNavigator.tsx
@@ -34,7 +34,8 @@ import {
   pathNavigatorRefresh,
   pathNavigatorSetCollapsed,
   pathNavigatorSetKeyword,
-  pathNavigatorSetLocaleCode
+  pathNavigatorSetLocaleCode,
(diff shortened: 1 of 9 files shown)


ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -117,6 +122,8 @@ export interface PathNavigatorStateProps {
   isFetching: boolean;
   error: any;
   isRootPathMissing: boolean;
+  sortStrategy?: GetChildrenOptions['sortStrategy'];
+  order?: GetChildrenOptions['order'];
```
rart:
```suggestion
  sortStrategy: GetChildrenOptions['sortStrategy'];
  order: GetChildrenOptions['order'];
```

ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -195,9 +206,15 @@ export function PathNavigator(props: PathNavigatorProps) {
...
-    uuid
+    uuid,
+    sortStrategy,
+    order
   ]);
 
+  useEffect(() => {
+    dispatch(batchActions([pathNavigatorUpdate({ id, sortStrategy, order }), pathNavigatorRefresh({ id })]));
+  }, [dispatch, id, sortStrategy, order]);
+
```
rart: This doesn't apply to other config-supplied props. Let's leave out for now.
```suggestion
```

ui/app/src/state/epics/pathNavigator.ts
```diff
@@ -284,7 +299,14 @@ export default [
...
             );
           } else {
-            return fetchItemWithChildrenByPath(site, path, { excludes, limit, offset, keyword }).pipe(
+            return fetchItemWithChildrenByPath(site, path, {
+              excludes,
+              limit,
+              offset,
+              keyword,
+              sortStrategy: state.pathNavigator[id].sortStrategy,
+              order: state.pathNavigator[id].order
```
rart, line 310: Let's check why the other come from payload and these don't. It might be ok, but check.

ui/app/src/state/epics/pathNavigatorTree.ts
```diff
@@ -286,16 +286,18 @@ export default [
...
               tree.isRootPathMissing &&
               (targetPath === rootPath || withIndex(targetPath) === rootPath)
             ) {
               actions.push(pathNavigatorTreeRefresh({ id }));
             } else if (
-              // If an entry for the path exists, assume it's an update to an existing item
-              targetPath in tree.totalByPath ||
-              withIndex(targetPath) in tree.totalByPath
+              // If an entry for the path exists, assume it's an update to an existing item.
+              // If sorting options are set, the parent paths needs to be updated sort the sort order to be correct.
```
rart: typo

rart (changes requested):

ui/app/src/state/epics/pathNavigator.ts
jvega190: I added the other props into the payload for it to be consistent. But in the update config scenario it doesn't make the tree update, since `pathNavigatorFetchParentItems` is called by the `pathNavigatorInit`, that is only called when there is no state set (not on updates).

rart (approved):

PR #3385, opened 2023-07-20
[570E]

jvega190:
https://github.com/craftersoftware/craftercms/issues/570


static-assets/components/cstudio-common/common-api.js
```diff
@@ -295,7 +295,7 @@ var nodeOpen = false,
         '/static-assets/themes/cstudioTheme/css/contextNav.css',
         '/static-assets/yui/container/assets/container.css',
         '/static-assets/jquery/jquery-time/jquery.timeentry.css',
-        '/static-assets/libs/jquery-ui/jquery-ui.min.css'
+        '/static-assets/libs/jquery-ui-1.13.2/jquery-ui.min.css'
```
rart: Let's check on when these are all imported and if they are all needed at that particular time.
That jQuery.timeentry.css looks suspiciously unused.

static-assets/libs/bootstrap-5.3.0/js/bootstrap.js
rart, line 1: Did you manage to do a custom build for this?

rart (changes requested):
- I see a new bs sheet added but don't see one removed. How did that play out?
- Could we take an approach similar to the one we do on ace and avoid checking in the library code/dist on our repo?

static-assets/libs/bootstrap-5.3.0/js/bootstrap.js
jvega190: No I didn't. The option from their page seems to be removed and I didn't see an option when building their source

static-assets/components/cstudio-common/common-api.js
jvega190: Removed timeentry and treeview from the list (not used). For the others I found matching styles/elements in legacy site-config (content-types view)

ui/bootstrap/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@craftercms/bootstrap-adapter",
+  "version": "1.0.0",
+  "main": "index.js",
+  "author": "Jose V.",
+  "license": "LGPL-3.0-only",
+  "private": true,
+  "scripts": {
+    "build": "chmod +x build.sh; ./build.sh"
```
rart: This can probably be done via simple gradlew copy. Preferably we would avoid sh scripts Let's see if we can rather remove the ace build.sh and consolidate (see comment below)

static-assets/libs/jquery-ui-1.13.2/package.json
rart, line 1: This is also a candidate for copying from `node_modules` directly (like bootstrap). Maybe we can put all these together/consolidate everything that's mostly a copy and it's not a real package.

rart (changes requested):

package.json
```diff
@@ -8,7 +8,9 @@
     "ui/scss",
     "ui/monaco",
     "ui/ace",
-    "ui/uppy"
+    "ui/uppy",
+    "ui/bootstrap",
+    "ui/npm-content"
```
rart: Didn't we add bootstrap into the npm content?

ui/npm-content/build.sh
```diff
@@ -0,0 +1,47 @@
...
+rm -rf "$libsDirectory/ace"
+rsync -ar --delete ../../node_modules/ace-builds/src-min-noconflict/ "$libsDirectory/ace/"
+cat src/ace-append.js >> "$libsDirectory/ace/ace.js"
+cp src/mode-yaml/* "$libsDirectory/ace"
+echo "Ace build complete"
+
+# js-yaml build
+rm -rf "$libsDirectory/js-yaml"
+mkdir "$libsDirectory/js-yaml"
+rsync -ar --delete ../../node_modules/js-yaml/dist/js-yaml.min.js "$libsDirectory/js-yaml/js-yaml-4.0.0.min.js"
```
rart: Who pulls this resource? Is the consumer using the new `-4.0.0` string as part of the name. Anyway we decided to leave the cache control on the server so we can dismiss the version piece of the string.

rart (changes requested):

ui/npm-content/build.sh
jvega190: Actually the thing is that it was already named like that (with the `-4.0.0` string) and I didn't update it according to the actual version. I'll remove the version string from the file and update usages.

package.json
jvega190: Yeah it is added, removing entry from here

rart (approved):

PR #3386, opened 2023-08-01
[6169] Incorrect path in content/item_by_path API 2 request on search

jvega190:
https://github.com/craftercms/craftercms/issues/6169

### DIFF
--- a/ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
+++ b/ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
@@ -14,7 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import AsyncVideoPlayer from '../AsyncVideoPlayer/AsyncVideoPlayer';
 import LoadingState, { ConditionalLoadingState } from '../LoadingState/LoadingState';
 import IFrame from '../IFrame/IFrame';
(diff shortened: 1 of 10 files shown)


ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
```diff
@@ -52,11 +53,12 @@ export function PreviewDialogContainer(props: PreviewDialogContainerProps) {
           <AsyncVideoPlayer playerOptions={{ src: url, autoplay: true, ...(mimeType ? { type: mimeType } : {}) }} />
         );
       case 'page':
+        const pagePreviewUrl = `${guestBase}${getPreviewURLFromPath(url)}?crafterCMSGuestDisabled=true`;
         return (
           <>
             {isLoading && <LoadingState />}
             <IFrame
-              url={url}
+              url={pagePreviewUrl}
```
rart: Not sure about this. How does this fix things?

ui/guest/src/index.tsx
```diff
@@ -89,7 +89,7 @@ export function initExperienceBuilder(props: ExperienceBuilderProps) {
   root.render(
     // @ts-ignore - typing system is not playing nice with the {path} | {model} options of GuestProps
     <ExperienceBuilder isAuthoring={crafterCMSGuestDisabled !== 'true'} {...props}>
-      <GuestProxy />
+      {crafterCMSGuestDisabled !== 'true' && <GuestProxy />}
```
rart: Let's discuss.

rart (changes requested):

ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
jvega190: What is now `pagePreviewUrl` was sent by the search hook as `url`, but it is used in this component for multiple things, and one of the usages is to retrieve a detailedItem, so that is the reason why there was an incorrect call (wrong path). Now the item path is the one being sent, and the previewUrl is defined in the PreviewDialog component for the iframe to render the proper page

ui/app/src/components/Search/utils.ts
```diff
@@ -364,7 +364,7 @@ export const useSearchState = ({ searchParameters, onSelect }: useSearchStatePro
           showPreviewDialog({
             type: 'page',
             title,
-            url: `${guestBase}${getPreviewURLFromPath(item.path)}?crafterCMSGuestDisabled=true`
+            url: `${guestBase}${getPreviewURLFromPath(url)}?crafterCMSGuestDisabled=true`
```
rart: Notice on line 350 path is getting renamed to url. Let's fix that and review if there's any other misuse here.

ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
```diff
@@ -26,23 +26,37 @@ import DialogFooter from '../DialogFooter';
...
   const [isLoading, setIsLoading] = useState(true);
 
+  useEffect(() => {
+    if (type === 'editor' && path) {
+      if (items[path]) {
+        setItem(items[path]);
+      } else {
+        fetchSandboxItem(siteId, path).subscribe((item) => setItem(item));
+      }
+    }
```
rart: Can we dispatch the fetch, retrieve from state and remove this additional local state?

rart (changes requested):

ui/app/src/components/PreviewDialog/PreviewDialogContainer.tsx
```diff
@@ -26,23 +26,32 @@ import DialogFooter from '../DialogFooter';
...
   const dispatch = useDispatch();
   const guestBase = useSelection<string>((state) => state.env.guestBase);
-
   const [isLoading, setIsLoading] = useState(true);
 
+  useEffect(() => {
+    if (type === 'editor' && path && !items[path]) {
+      dispatch(fetchSandboxItem({ path }));
+    }
+  }, [siteId, items, path, type, dispatch]);
```
rart: Does this cause multiple fetch calls to be made while the thing is being fetched? You may need to also use `itemsBeingFetchedByPath` to avoid issuing multiple calls. Use network throttle test the case well.

rart (changes requested):

rart (approved):

PR #3387, opened 2023-08-04
[6161] Node selector UI not usable if options are too numerous

jvega190:
https://github.com/craftercms/craftercms/issues/6161

### DIFF
--- a/ui/scss/src/forms-engine.scss
+++ b/ui/scss/src/forms-engine.scss
@@ -212,7 +212,8 @@ h6,
 }
 
 // Styles to fix style conflicts between material-ui and bootstrap
-legend[class^="mui-"], hr.MuiDivider-root {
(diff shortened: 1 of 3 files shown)


ui/scss/src/forms-engine.scss
```diff
@@ -219,3 +219,8 @@ legend[class^="mui-"], hr.MuiDivider-root {
 .MuiTooltip-popper {
   z-index: 1500 !important;
 }
+
+.cstudio-dropdown .dropdown-menu {
+  max-height: 500px;
```
rart: use 80-90vh

rart (changes requested):

rart (approved):
Will this somehow collide with the bootstrap upgrade? @jvega190

PR #3389, opened 2023-08-07
[6165] XB buttons are not role aware

jvega190:
https://github.com/craftercms/craftercms/issues/6165

### DIFF
--- a/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
+++ b/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
@@ -198,7 +198,16 @@ const startCommunicationDetectionTimeout = (timeoutRef, setShowSnackbar, timeout
 
 // region const issueDescriptorRequest = () => {...}
 const issueDescriptorRequest = (props) => {
-  const { site, path, contentTypes, requestedSourceMapPaths, flatten = true, dispatch, completeAction } = props;
+  const {
+    site,
+    path,
+    contentTypes,
+    requestedSourceMapPaths,
+    flatten = true,
+    dispatch,
+    completeAction,
+    permissions
+  } = props;
   const hostToGuest$ = getHostToGuestBus();
   const guestToHost$ = getGuestToHostBus();
 
@@ -280,7 +289,8 @@ const issueDescriptorRequest = (props) => {
           modelLookup: normalizedModels,
           hierarchyMap,
           modelIdByPath: modelIdByPath,
-          sandboxItems
+          sandboxItems,
+          permissions
         }
       });
     });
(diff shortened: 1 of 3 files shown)


ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -609,7 +620,8 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
             contentTypes,
             requestedSourceMapPaths,
             dispatch,
-            completeAction: fetchPrimaryGuestModelComplete
+            completeAction: fetchPrimaryGuestModelComplete,
+            permissions
```
rart, line 624: This needs to be added to the upToDateRefs. Right now its stale. You should be getting compilation warnings.

ui/guest/src/react/ZoneMenu.tsx
```diff
@@ -146,15 +156,17 @@ export function ZoneMenu(props: ZoneMenuProps) {
       const maxValidation = validations?.maxCount?.value;
       const minValidation = validations?.minCount?.value;
       const trashableValidation = minValidation ? minValidation < numOfItemsInContainerCollection : true;
-      const duplicateValidation = maxValidation ? maxValidation > numOfItemsInContainerCollection : true;
+      const duplicateValidation =
+        permissions.includes('content_create') &&
```
rart, line 160: This is to delete. Isn't there a delete permission?

rart (changes requested):

ui/guest/src/react/ZoneMenu.tsx
jvega190: Per conversation this section is for duplicateValidation, not for trashableValidation

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
```diff
@@ -298,6 +308,7 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
   const store = useStore<GlobalState>();
   const { id: siteId, uuid } = useActiveSite() ?? {};
   const user = useActiveUser();
+  const permissions = user?.permissionsBySite[siteId];
```
rart: Actually, this is part of the user object which is already in the updated refs. There's no point on having this declared outside of the effect, is there?

rart (changes requested):

ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
jvega190: no point, changing it to use user from refs to get permissions

rart (approved):

PR #3390, opened 2023-08-09
[573E]

jvega190:
https://github.com/craftersoftware/craftercms/issues/573

### DIFF
--- a/ui/app/src/components/LoginView/LoginView.tsx
+++ b/ui/app/src/components/LoginView/LoginView.tsx
@@ -48,7 +48,7 @@ import { useMount } from '../../hooks/useMount';
 import { useDebouncedInput } from '../../hooks/useDebouncedInput';
 import { PasswordStrengthDisplayPopper } from '../PasswordStrengthDisplayPopper';
 import { USER_USERNAME_MAX_LENGTH } from '../UserManagement/utils';
-import useIntCountdown from '../../hooks/useIntCountdown';
+import useTimer from '../../hooks/useTimer';
 import { nnou } from '../../utils/object';
 import moment from 'moment-timezone';
 
@@ -134,12 +134,7 @@ const translations = defineMessages({
     defaultMessage: 'Token validation failed.'
   },
   lockedAccountTryAgain: {
-    id: 'loginView.lockedAccountTryAgain',
-    defaultMessage: 'Try again {time}'
-  },
-  lockedAccountTryAgainSeconds: {
-    id: 'loginView.lockedAccountTryAgainSeconds',
-    defaultMessage: 'in {seconds} seconds'
+    defaultMessage: 'Try again {fullTime, select, true {{time}} other {in {time} seconds}}'
   }
 });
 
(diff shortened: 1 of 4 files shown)


ui/app/src/hooks/useIntCountdown.ts
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    if (countdown) {
+      const timer =
+        countdown > 0 &&
+        setInterval(() => {
+          setCountdown(countdown - 1);
+        }, 1000);
+      return () => clearInterval(timer);
+    }
+  }, [countdown]);
```
rart: As it stands, this works with a timeout rather than an interval. It continually clears the interval every second. 
```suggestion
  useEffect(() => {
    if (seconds > 0) {
      let num = seconds;
      const timer = setInterval(() => {
        setCountdown(--num);
        if (num === 0) {
          clearInterval(timer);
          onCompleteRef.current?.();
        }
      }, 1000);
      return () => clearInterval(timer);
    }
  }, [seconds]);
```

ui/app/src/hooks/useIntCountdown.ts
```diff
@@ -0,0 +1,36 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { useEffect, useState } from 'react';
+
+export const useIntCountdown = (number: number) => {
+  const [countdown, setCountdown] = useState(number);
```
rart:
```suggestion
export const useTimer = (seconds: number, onComplete: (): void) => {
  const [countdown, setCountdown] = useState(seconds);
  const onCompleteRef = useUpdatedRefs(onComplete);
```

ui/app/src/components/LoginView/LoginView.tsx
```diff
@@ -125,6 +132,14 @@ const translations = defineMessages({
...
     defaultMessage: 'Token validation failed.'
+  },
+  lockedAccountTryAgain: {
+    id: 'loginView.lockedAccountTryAgain',
+    defaultMessage: 'Try again {time}'
+  },
+  lockedAccountTryAgainSeconds: {
+    id: 'loginView.lockedAccountTryAgainSeconds',
+    defaultMessage: 'in {seconds} seconds'
   }
```
rart: We should be able to remove ids on these two, no?
Also, I think these can be combined into a single translation message with the conditional logic thingie.
Preferably, use the FormattedMessage API.

ui/app/src/components/LoginView/LoginView.tsx
```diff
@@ -189,11 +204,14 @@ function LoginView(props: SubViewProps) {
...
     xsrfToken,
-    language
+    language,
+    lockedErrorMessage,
+    lockedTimeSeconds
   } = props;
   const [username, setUsername] = useState(() => localStorage.getItem('username') ?? '');
   const [password, setPassword] = useState('');
   const [error, setError] = useState('');
+  const lockedCountdown = useIntCountdown(lockedTimeSeconds);
```
rart: What's the value of this prop when the user isn't locked out?

rart (changes requested):

ui/app/src/components/LoginView/LoginView.tsx
jvega190: the value is 'undefined'

ui/app/src/components/LoginView/LoginView.tsx
jvega190: updated/combined translation. The error message is already using `formatMessage` (string) for other messages, should I refactor it?

rart: nah

rart (approved):

PR #3391, opened 2023-08-10
[6173] Row and Column settings for Text Area field has no effect

jvega190:
https://github.com/craftercms/craftercms/issues/6173


rart (changes requested):
Please remove cols all around

rart (approved):

PR #3395, opened 2023-08-18
[6176] Studio Search fails when searching with Special Characters

jvega190:
https://github.com/craftercms/craftercms/issues/6176

### DIFF
--- a/ui/app/src/components/Search/utils.ts
+++ b/ui/app/src/components/Search/utils.ts
@@ -279,7 +279,7 @@ export const useSearchState = ({ searchParameters, onSelect }: useSearchStatePro
   const refreshSearch = useCallback(() => {
     setError(null);
     setIsFetchingResults(true);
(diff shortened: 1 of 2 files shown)


ui/app/src/components/Search/utils.ts
```diff
@@ -279,7 +279,7 @@ export const useSearchState = ({ searchParameters, onSelect }: useSearchStatePro
   const refreshSearch = useCallback(() => {
     setError(null);
     setIsFetchingResults(true);
-    search(site, searchParameters).subscribe({
+    search(site, { ...searchParameters, keywords: encodeURIComponent(searchParameters.keywords) }).subscribe({
```
rart: Encode everything at the service. The site id should be encoded too, use the QS util.

rart (changes requested):

rart (approved):

PR #3396, opened 2023-08-21
[6153] Cut & Paste content does not carry over the content's history

jvega190:
https://github.com/craftercms/craftercms/issues/6153

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -3691,7 +3691,7 @@ var nodeOpen = false,
 
         this.getVersionHistory(site, contentTO, {
           success: function (response) {
-            callback.success(response.items[0].versionNumber);
+            callback.success(response[0].versionNumber);
           }
         });
       },
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -3691,7 +3691,7 @@ var nodeOpen = false,
 
         this.getVersionHistory(site, contentTO, {
           success: function (response) {
-            callback.success(response.versions[0].versionNumber);
+            callback.success(response.items[0].versionNumber);
           }
         });
       },
--- a/ui/app/src/CHANGELOG.md
+++ b/ui/app/src/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.2
+* Update return type of configuration/fetchHistory and content/fetchItemHistory to `ItemHistoryEntry[]`.
+* Update content/fetchItemHistory service to use new `/studio/api/2/content/item_history` API.
+
 ## 4.1.1
 
 * SandboxItem and DetailedItem changes:
@@ -5,5 +9,3 @@
   * Add `submitter` prop of type `Person`.
   * Add `dateSubmitted` prop of type string.
 * Update PendingApprovalDashlet and ScheduledDashlet to use new `submitter` and `dateSubmitted` props.
-* Update return type of configuration/fetchHistory and content/fetchItemHistory to `ItemHistoryEntry[]`.
-* Update content/fetchItemHistory service to use new `/studio/api/2/content/item_history` API.
--- a/ui/app/src/CHANGELOG.md
+++ b/ui/app/src/CHANGELOG.md
@@ -5,3 +5,5 @@
   * Add `submitter` prop of type `Person`.
   * Add `dateSubmitted` prop of type string.
 * Update PendingApprovalDashlet and ScheduledDashlet to use new `submitter` and `dateSubmitted` props.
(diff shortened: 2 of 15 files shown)


static-assets/components/cstudio-common/common-api.js
```diff
@@ -3691,7 +3691,7 @@ var nodeOpen = false,
 
         this.getVersionHistory(site, contentTO, {
           success: function (response) {
-            callback.success(response.versions[0].versionNumber);
+            callback.success(response.items[0].versionNumber);
```
rart: Didn't we make the service return directly the versions array?

ui/app/src/models/Version.ts
```diff
@@ -14,21 +14,20 @@
...
+import { DetailedItem } from './Item';
 import ApiResponse from './ApiResponse';
 import ContentInstance from './ContentInstance';
+import Person from './Person';
 
-export interface LegacyVersion {
-  lastModifiedDate: string;
-  lastModifier: string;
-  versionNumber: string;
+export interface ItemVersion {
```
rart:
```suggestion
export interface ItemHistoryEntry {
```

rart (changes requested):
Please update CHANGELOG with important & breaking changes (for example the change in return of the service functions)

ui/app/src/CHANGELOG.md
```diff
@@ -5,3 +5,5 @@
   * Add `submitter` prop of type `Person`.
   * Add `dateSubmitted` prop of type string.
 * Update PendingApprovalDashlet and ScheduledDashlet to use new `submitter` and `dateSubmitted` props.
+* Update return type of configuration/fetchHistory and content/fetchItemHistory to `ItemHistoryEntry[]`.
+* Update content/fetchItemHistory service to use new `/studio/api/2/content/item_history` API.
```
rart: This goes over 4.1.2, dude...

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -168,7 +162,7 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
         ])
       );
     } else if (isItemPreviewable) {
-      fetchContentByCommitId(site, item.path, version.versionNumber).subscribe((content) => {
+      fetchContentByCommitId(site, ensureSingleSlash(`/${versionPath}`), version.versionNumber).subscribe((content) => {
```
rart: Why would a path not have it's starting slash?
Why is this needed only on this code path?

rart (changes requested):

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
jvega190: paths are coming without starting slash in the item history API, but only `fetchContentByCommitId` seems to be failing when the path doesn't have its starting slash

rart: @sumerjabri @jmendeza is it expected that the paths come without their leading slash [or is it a bug]?

jmendeza: @rart I would say it is a bug, since everywhere we used the path with the leading slash.

jmendeza: Fixed here https://github.com/craftercms/studio/pull/3314

rart: @jvega190 please update and adjust

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -51,14 +50,16 @@ import {
...
 import SingleItemSelector from '../SingleItemSelector';
-import { SuspenseWithEmptyState } from '../Suspencified/Suspencified';
 import VersionList from '../VersionList';
 import DialogFooter from '../DialogFooter/DialogFooter';
 import { HistoryDialogPagination } from './HistoryDialogPagination';
 import { historyStyles } from './HistoryDialog';
 import useSelection from '../../hooks/useSelection';
 import useFetchSandboxItems from '../../hooks/useFetchSandboxItems';
 import { UNDEFINED } from '../../utils/constants';
+import { ensureSingleSlash } from '../../utils/string';
```
rart: Is this still in use? Don't see it used on the diff insertions and seems like a new dependency. You're probably getting warnings on the build?

rart (changes requested):

rart (approved):

PR #3399, opened 2023-08-23
[6182] Configuration for targeting with date-time control does not work

jvega190:
https://github.com/craftercms/craftercms/issues/6182

### DIFF
--- a/ui/app/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/ui/app/src/components/DateTimePicker/DateTimePicker.tsx
@@ -30,12 +30,14 @@ import {
   create8601String,
   get8601Pieces,
   getTimezones,
+  getUserTimeZone,
   TimezoneDescriptor
 } from '../../utils/datetime';
 import FormControl from '@mui/material/FormControl';
-import { nnou } from '../../utils/object';
+import { nnou, nou } from '../../utils/object';
 import { useSpreadState } from '../../hooks/useSpreadState';
 import { UNDEFINED } from '../../utils/constants';
+import useLocale from '../../hooks/useLocale';
 
 export interface DateChangeData {
   date: Date;
@@ -50,6 +52,8 @@ export interface DateTimePickerProps {
   disabled?: boolean;
   classes?: any;
   controls?: Array<'date' | 'time' | 'timeZone'>; // options: ['date', 'time', 'timezone'], ['date', 'time'], ['date']
+  showTimeSelector?: boolean;
+  showTimeZoneSelector?: boolean;
   disablePast?: boolean;
   localeCode: string;
   dateTimeFormatOptions: Intl.DateTimeFormatOptions;
@@ -88,6 +92,7 @@ const useStyles = makeStyles()(() => ({
 }));
 
 function DateTimePicker(props: DateTimePickerProps) {
+  const locale = useLocale();
   const {
     id,
     onChange,
@@ -96,10 +101,11 @@ function DateTimePicker(props: DateTimePickerProps) {
     onTimeZoneChange,
     onError,
     value,
-    timeZone,
+    timeZone = locale.dateTimeFormatOptions.timeZone ?? getUserTimeZone(),
     disablePast,
     disabled = false,
-    controls = ['date', 'time', 'timeZone'],
+    showTimeSelector = true,
+    showTimeZoneSelector = true,
     localeCode = 'en-US',
     dateTimeFormatOptions
   } = props;
@@ -109,9 +115,12 @@ function DateTimePicker(props: DateTimePickerProps) {
   // displayed when the time picker is opened. `internalDate` is a transposed date/time
   // used to sync the timepicker with what's displayed to the user.
   const internalDate = useMemo(() => {
-    const date = value ? new Date(value) : new Date();
-    const localOffset = moment().format().substr(-6);
-    const dateWithoutOffset = moment(date).tz(timeZone).format().substr(0, 19);
+    if (nou(value)) {
+      return null;
+    }
+    const date = value ? new Date(value) : null;
+    const localOffset = moment().format().slice(-6);
+    const dateWithoutOffset = moment(date).tz(timeZone).format().substring(0, 19);
     return new Date(`${dateWithoutOffset}${localOffset}`);
   }, [value, timeZone]);
   const timeZones = getTimezones();
@@ -133,12 +142,8 @@ function DateTimePicker(props: DateTimePickerProps) {
     // Date/time change handler
     (newMoment: Moment) => {
       let newDate = newMoment.toDate();
-      if (newDate === null) {
-        setPickerState({ dateValid: false });
-        onError?.();
-      }
       let changes: DateChangeData;
-      const internalDatePieces = get8601Pieces(internalDate);
+      const internalDatePieces = get8601Pieces(internalDate ?? newDate);
       const pickerDatePieces = get8601Pieces(newDate);
       switch (name) {
         case 'date': {
@@ -156,7 +161,6 @@ function DateTimePicker(props: DateTimePickerProps) {
           break;
         }
       }
-      setPickerState({ dateValid: true });
       onChange?.(changes);
     };
 
@@ -164,7 +168,7 @@ function DateTimePicker(props: DateTimePickerProps) {
   // date to the new timezone but that you can select, date, time and timezone
   // individually without one changing the other fields.
   const handleTimezoneChange = (event, value: TimezoneDescriptor) => {
-    const pieces = get8601Pieces(internalDate);
+    const pieces = get8601Pieces(internalDate ?? new Date());
     // Keep date/time we had and apply new offset
     const dateString = create8601String(pieces[0], pieces[1], value.offset);
     const changes: DateChangeData = { date: new Date(dateString), dateString, timeZoneName: value.name };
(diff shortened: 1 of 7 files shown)


ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -30,10 +30,11 @@ import CheckboxGroup from '../FormEngineControls/CheckboxGroup';
 import DateTime from '../FormEngineControls/DateTime';
 import LookupTable from '../../models/LookupTable';
 import { Alert } from '@mui/material';
+import { create8601String, get8601Pieces } from '../../utils/datetime';
 
 interface AudiencesPanelUIProps {
   model: ContentInstance;
-  fields: LookupTable<ContentTypeField>;
+  fields: LookupTable<ContentTypeField & { timeZone?: string }>;
```
rart: Why treat time zone different?

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -67,14 +68,26 @@ const messages = defineMessages({
...
-  });
 
+    if (fields[fieldId].type === 'date-time') {
+      // If there is no default value set for date-time, set it to the current date/time
+      if (!propValue) {
+        props[fieldId] = getDateTimeString(new Date());
+      }
+      props[`${fieldId}_tz`] = fields[fieldId].timeZone;
+    }
+  });
```
rart: This would make this field essentially non nullable, right? Need to check with Russ if that's a problem. It sounds like it could.

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -123,8 +136,15 @@ export function AudiencesPanelUI(props: AudiencesPanelUIProps) {
...
                 disabled: modelApplying
               };
-              if (controlProps.field.type === 'date-time' && model[`${fieldId}_tz`]) {
-                controlProps['timezone'] = model[`${fieldId}_tz`];
+              if (controlProps.field.type === 'date-time') {
+                if (!controlProps.value) {
+                  controlProps.value = fields[fieldId].defaultValue ?? getDateTimeString(new Date());
+                }
+                if (model[`${fieldId}_tz`]) {
+                  controlProps['timeZone'] = model[`${fieldId}_tz`];
```
rart:
```suggestion
                  controlProps.timeZone = model[`${fieldId}_tz`];
```

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -123,8 +136,15 @@ export function AudiencesPanelUI(props: AudiencesPanelUIProps) {
...
+                } else {
+                  controlProps['timeZone'] = fields[fieldId].timeZone;
```
rart:
```suggestion
                  controlProps.timeZone = fields[fieldId].timeZone;
```

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -123,8 +136,15 @@ export function AudiencesPanelUI(props: AudiencesPanelUIProps) {
...
+                }
```
rart: Need to make sure making things non-nullable isn't a problem. I think it could.

rart (changes requested):

rart:
@jvega190 remember/FYI this has pending changes

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
jvega190: AudiencesPanel fields happen to match the CotentTypeField type, but it doesn't have a `timeZone`. Should I create a separate type for this since the match is not exact (e.g. AudiencesPanel fields do not have validations)?

rart: I meant timeZone could be treated as a field.
But, is this still applicable based on the last discussions? We're not tracking the time zone value separately from the 8610 string. Can we remove this stuff?

ui/app/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -109,9 +115,12 @@ function DateTimePicker(props: DateTimePickerProps) {
   // displayed when the time picker is opened. `internalDate` is a transposed date/time
   // used to sync the timepicker with what's displayed to the user.
   const internalDate = useMemo(() => {
-    const date = value ? new Date(value) : new Date();
-    const localOffset = moment().format().substr(-6);
-    const dateWithoutOffset = moment(date).tz(timeZone).format().substr(0, 19);
+    if (nou(value)) {
+      return null;
+    }
+    const date = value ? new Date(value) : null;
```
rart: This line seems like a redundant check based on the previous lines conditional statement

ui/app/src/components/PreviewAudiencesPanel/PreviewAudiencesPanel.tsx
```diff
@@ -54,6 +55,7 @@ export function PreviewAudiencesPanel(props: PreviewAudiencesPanelProps) {
   const { fields } = props;
   const panelState = useSelection<GlobalState['preview']['audiencesPanel']>((state) => state.preview.audiencesPanel);
   const hasNoFields = !fields || Object.values(fields).length === 0;
+  const [timeZonesLookup, setTimeZonesLookup] = useSpreadState({});
```
rart: This being local state would get lost going in and out of the panel [as opposed to the rest of the state which is in redux or comes via props].

Where are the fields coming from?

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -67,14 +69,20 @@ const messages = defineMessages({
...
   Object.keys(fields).forEach((fieldId: string) => {
     const propValue = fields[fieldId].defaultValue;
     props[fieldId] = propValue;
-  });
+    props[fieldId] = propValue;
 
+    if (fields[fieldId].type === 'date-time') {
+      setTimeZoneLookups({
+        [fieldId]: fields[fieldId].timeZone
+      });
```
rart: Avoid calling state setters in loops.

rart (changes requested):
Overall the time zone keeping seems unclean. We need to think through make it cleaner. 

Maybe the control itself keeps that state and the timezone part is not controlled directly?

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
jvega190: the thing is that timeZone is not a field itself, but a part of a field (when the control is date-time). We can't remove since the one in here is the one coming from the ui.xml where you can set the timeZone you want as the default. Will explain a bit better on a separate comment.

jvega190:
There are 2 `timeZone` considerations.

1. timeZone property that is part of the `AudiencesPanel` fields prop. This one is a property of a field of the AudiencesPanel component, that maps to the timeZone configuration we have in ui.xml for AudiencesPanel. This ends up being the timeZone showing in the control if set, if not set, then the time zone shown will the the configured in the site, or the browser's tz.
An example of a date-time field is this:
```
currentTime: {
  defaultValue: "2023-08-25T21:46:22+02:00",
  id: "currentTime",
  name: "Date",
  timeZone: "Europe/Berlin",
  type: "date-time"
} 
```
See how the timeZone is a part of the 'currentTime' field, not a field itself

2. timeZonesLookup. This one is for the AudiencesPanel to keep a local state of the selected timezone in the control while you're in the panel. If you go out of the panel and go in again now it'll display the configured or default one. If we update the control to keep its own state it'll be lost after applying changes since AudiencesPanel does a re-render when applying  them.

rart (commented):
Please unwrap from the useless fragment and `{` block starting on lines 120 and 121

rart:
Ping me to discuss next steps

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
rart: What about now? Is this still applicable now that we've removed the timezone prop?

jvega190: doesn't apply anymore, already removed in https://github.com/craftercms/studio-ui/pull/3399/commits/273671a2d0590d12ae6a353964732c71b499ae17

ui/app/src/components/PreviewAudiencesPanel/AudiencesPanelUI.tsx
```diff
@@ -111,38 +107,36 @@ export function AudiencesPanelUI(props: AudiencesPanelUIProps) {
...
+          const controlProps = {
+            field: fields[fieldId],
+            value: model[fieldId] ?? undefined,
+            onChange: onFieldChange(fieldId, type),
+            disabled: modelApplying
+          };
+          if (controlProps.field.type === 'date-time') {
+            if (!controlProps.value) {
+              controlProps.value = fields[fieldId].defaultValue;
+            }
```
rart:
```suggestion
          if (controlProps.field.type === 'date-time' && !controlProps.value) {
            controlProps.value = fields[fieldId].defaultValue;
```

rart (changes requested):

rart (approved):

PR #3404, opened 2023-08-28
[6175] Remove the maximum width on the sidebar

jvega190:
https://github.com/craftercms/craftercms/issues/6175


ui/app/src/hooks/usePreviewUrlControl.tsx
```diff
@@ -126,6 +127,19 @@ export function usePreviewUrlControl(history) {
...
+  });
+  useEffect(() => {
+    const handleResize = () => {
+      dispatch(setWindowSize({ size: window.innerWidth }));
+    };
+    window.addEventListener('resize', handleResize);
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, [dispatch]);
```
rart: 1. Why here? Don't think this is the right place for this as this file is got to do with URL monitoring. Make this into a hook and call it in PreviewConcierge. 
2. You need to debounce the dispatch.
```ts 
fromEvent(window, 'resize').pipe(debounce(200)).subscribe(()=>dispatch(setWindowSize()))
```

ui/app/src/state/reducers/preview.ts
```diff
@@ -191,9 +192,17 @@ const fetchGuestModelsCompleteHandler = (state, { type, payload }) => {
 
 const reducer = createReducer<GlobalState['preview']>(initialState, {
   [openToolsPanel.type]: (state) => {
+    const { windowSize, icePanelWidth } = state;
+    let toolsPanelWidth;
+    if (icePanelWidth - windowSize >= 320) {
```
rart: Extract 320 as a const for reuse. You'll need it in a few places.

ui/app/src/components/ICEToolsPanel/ICEToolsPanel.tsx
```diff
@@ -63,7 +63,14 @@ export function ICEToolsPanel() {
   }, [uiConfig.xml, dispatch, icePanel, site, username, uuid]);
 
   return (
-    <ResizeableDrawer open={isOpen} belowToolbar anchor="right" width={width} onWidthChange={onWidthChange}>
+    <ResizeableDrawer
+      open={isOpen}
+      belowToolbar
+      anchor="right"
+      width={width}
+      maxWidth={windowSize}
```
rart, line 66: Max width size would be `windowSize-minPreviewWidth-otherToolbarSize`?
What's this for anyway? It's controlled in reducers and it wasn't there before.

ui/app/src/components/ToolsPanel/ToolsPanel.tsx
```diff
@@ -104,6 +103,7 @@ export function ToolsPanel() {
       belowToolbar
       open={showToolsPanel}
       width={toolsPanelWidth}
+      maxWidth={windowSize}
```
rart, line 101: As above

ui/app/src/state/reducers/preview.ts
rart, line 1: Haven't checked the maths but there's a bit of code repetition. Factor when you got it all clear.

rart (changes requested):

ui/app/src/hooks/useWindowSizeControl.ts
```diff
@@ -0,0 +1,51 @@
...
+  useMount(() => {
+    dispatch(setWindowSize({ size: window.innerWidth }));
+  });
+  // region Sidebars localStorage update
+  useEffect(() => {
+    setStoredICEToolsPanelWidth(siteId, user.username, icePanelWidth);
+  }, [icePanelWidth, siteId, user?.username]);
+  useEffect(() => {
+    setStoredPreviewToolsPanelWidth(siteId, user.username, toolsPanelWidth);
+  }, [toolsPanelWidth, siteId, user?.username]);
```
rart: I'm not sure user.username will ever be null (ie doesn't need optional chaining). 
Did you see that case? If that was the case, the inner logic needs to consider this.

ui/app/src/hooks/useWindowSizeControl.ts
rart, line 1: I'm thinking...
```ts
// epics/preview
  (action$, state$) =>
    action$.pipe(
      ofType(storeInitialized),
      switchMap(() =>
        fromEvent(window, 'resize').pipe(
          debounceTime(200),
          map((e) => setWindowSize({ width: e.target.innerWidth, height: e.target.innerHeight })),
          // Not quite sure where this needs to be exactly, so test 👇🏻. Might need to be put elsewhere
          // or use a different strategy to get this set on initialization.
          startWith(setWindowSize({ width: window.innerWidth, height: window.innerHeight }))
        )
      )
    ),
  (action$, state$) =>
    action$.pipe(
      ofType(setWindowSize),
      tap(([, state]) => {
        setStoredICEToolsPanelWidth(...)
        setStoredPreviewToolsPanelWidth(...)
      }),
      ignoreElements()
    ),
```
...and we get rid of this whole effect, which isn't very general purpose anyway.

ui/app/src/state/reducers/preview.ts
rart, line 1: ReduxToolkit sends you a wrapped state object which you can modify directly — no need for spreading. It does tend to look cleaner. Let's migrate to using that style.

rart (changes requested):

ui/app/src/hooks/useWindowSizeControl.ts
jvega190: the optional chaining is not needed, removing

ui/app/src/state/reducers/preview.ts
rart: This isn't done. Did you miss a commit?

rart (changes requested):
Comment above

ui/app/src/state/reducers/preview.ts
jvega190: Yeah there was a missing commit. Though now that I think, I only updated the handlers that I modified in this ticket. Should I update all the reducer with the new style?

rart (approved):

PR #3412, opened 2023-09-11
[6207] Create Content-Type form field validator showing label and ID already exists for empty values

jvega190:
https://github.com/craftercms/craftercms/issues/6207


ui/scss/src/forms-engine.scss
```diff
@@ -114,6 +114,9 @@ div.xdsoft_datetimepicker {
 @import '../../static-assets/libs/bootstrap-sass-official/assets/stylesheets/bootstrap/dropdowns';
 @import '../../static-assets/libs/bootstrap-sass-official/assets/stylesheets/bootstrap/panels';
 
+// Utility classes
+@import '../../static-assets/libs/bootstrap-sass-official/assets/stylesheets/bootstrap/utilities';
```
jvega190, line 118: Bootstrap utilities was removed from this file, containing styles like for `.hidden` class, which was causing the labels to be displayed

rart (approved):

PR #3450, opened 2023-10-19
[3941] Make content diff better

jvega190:
https://github.com/craftercms/craftercms/issues/3941


ui/app/src/components/CompareVersionsDialog/RepeatGroupItems.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+  useEffect(() => {
+    const contentALength = (contentA ?? []).length;
+    const contentBLength = (contentB ?? []).length;
+    const maxLength = contentALength > contentBLength ? contentALength : contentBLength;
+    const diffArray = [];
+
+    for (let x = 0; x < maxLength; x++) {
+      const itemA = contentA?.[x] ? JSON.stringify(contentA[x]) : null;
+      const itemB = contentB?.[x] ? JSON.stringify(contentB[x]) : null;
```
rart, line 62: JSON.strigify doesn't guarrantee order of properties and overall is likely not a reliable comparison. We already have the XML, it'd make more sense to compare that native string instead, which will also take away the stringification cost.

rart:
Superseded by https://github.com/craftercms/studio-ui/pull/4165

PR #3459, opened 2023-11-09
[6249] Move monaco package to npm-content

jvega190:
https://github.com/craftercms/craftercms/issues/6249


ui/npm-content/build.sh
```diff
@@ -58,3 +58,7 @@ rm -rf "$libsDirectory/tinymce"
 cp -r ../../node_modules/tinymce "$libsDirectory/tinymce"
 
 echo "TinyMCE build complete"
+
+# monaco build
+webpack --config webpack.config.js
```
rart: Will this line work in all cases? How? Webpack is typically not a global bash command. Should probably be ran through a package.json script via `yarn`.

Check if you have webpack installed globally.

ui/npm-content/webpack.config.js
rart, line 1: Move into Monaco folder

rart (changes requested):

ui/npm-content/build.sh
jvega190: Moved webpack command to yarn script

rart (approved):

PR #3461, opened 2023-11-09
[6270] Add hotkey support to the configuration editor

jvega190:
https://github.com/craftercms/craftercms/issues/6270


jvega190:
@rart I wonder if we should make an util for a saveCommand for this. Still not sure since ace addCommand is pretty general (multiple possible purposes)

rart (changes requested):
Lets add this shortcut to the file editor too.

PR #3464, opened 2023-11-10
[617E] Limit the size of configuration files in UI editor

jvega190:
https://github.com/craftersoftware/craftercms/issues/617

### DIFF
--- a/ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
+++ b/ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
@@ -33,6 +33,7 @@ import { useHistory } from 'react-router';
 import ConfirmDialog from '../ConfirmDialog/ConfirmDialog';
 import { useMount } from '../../hooks/useMount';
 import Paper from '@mui/material/Paper';
+import { getContentSizeInBytes, MAX_CONFIG_SIZE } from '../../utils/content';
 
 const translations = defineMessages({
   configSaved: {
@@ -106,6 +107,8 @@ export function GlobalConfigManagement() {
   };
 
   const onSaveClick = () => {
+    const value = aceEditorRef.current.getValue();
+    const contentSize = getContentSizeInBytes(value);
     const errors = aceEditorRef.current
       .getSession()
       .getAnnotations()
@@ -122,8 +125,18 @@ export function GlobalConfigManagement() {
           }
         })
       );
+    } else if (contentSize > MAX_CONFIG_SIZE) {
+      dispatch(
+        showSystemNotification({
+          message: formatMessage({
+            defaultMessage: 'Maximum configuration size exceeded, please reduce it in order to properly save.'
+          }),
+          options: {
+            variant: 'error'
+          }
+        })
+      );
     } else {
-      const value = aceEditorRef.current.getValue();
       writeConfiguration('studio_root', '/configuration/studio-config-override.yaml', 'studio', value).subscribe(
         () => {
           setLastSavedContent(value);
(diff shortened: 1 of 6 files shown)


ui/app/src/utils/content.ts
```diff
@@ -77,6 +77,8 @@ import { showCodeEditorDialog, showEditDialog } from '../state/actions/dialogs';
 import { Dispatch } from 'react';
 import { AnyAction } from 'redux';
 
+export const MAX_CONFIG_SIZE = 512000; // Max configuration files allowed (in bytes)
```
rart: Move to constants.

rart (changes requested):
What was the delta of the calculation in UI vs the OS?

rart:
What's the answer to the question above?

jvega190:
@rart exact max size in bytes for configuration files used in backend is 512*1024, differing from the OS calculation (524288 studio vs 512000 OS). Tested with the exact max size in bytes set by backend and when using an extra char/byte the backend request fails.

ui/app/src/utils/constants.ts
```diff
@@ -99,3 +99,5 @@ export const XSRF_TOKEN_COOKIE_NAME = 'XSRF-TOKEN';
 export const plugins = new Map<string, PluginDescriptorWithSource>();
 
 export const components = new Map<string, WidgetRecord>();
+
+export const MAX_CONFIG_SIZE = 512 * 1024; // Max configuration files allowed (in bytes)
```
rart:
```suggestion
export const MAX_CONFIG_FILE_SIZE = 524288; // Max configuration file size allowed (in bytes) (512 * 1024 = 524288)
```

ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
```diff
@@ -122,8 +126,18 @@ export function GlobalConfigManagement() {
...
+      dispatch(
+        showSystemNotification({
+          message: formatMessage({
+            defaultMessage: 'Maximum configuration size exceeded, please reduce it in order to properly save.'
+          }),
+          options: {
+            variant: 'error'
+          }
+        })
+      );
```
rart: With the [new approach](https://github.com/craftersoftware/craftercms/issues/617#issuecomment-1813948305) in mind, perhaps we simply disable the Save button and we don't need this.

ui/app/src/components/GlobalConfigManagement/GlobalConfigManagement.tsx
```diff
@@ -106,6 +108,8 @@ export function GlobalConfigManagement() {
   };
 
   const onSaveClick = () => {
+    const value = aceEditorRef.current.getValue();
+    const contentSize = getContentSizeInBytes(value);
```
rart: Let's just match the backend and check for string length.

ui/app/src/utils/content.ts
```diff
@@ -1067,3 +1067,7 @@ export function generateComponentBasePath(contentType: string) {
 export function generateComponentPath(modelId: string, contentType: string) {
   return `${generateComponentBasePath(contentType)}/${modelId}.xml`;
 }
+
+export function getContentSizeInBytes(content: string) {
+  return new TextEncoder().encode(content).length;
+}
```
rart: Guess we don't need this anymore.

rart (changes requested):
See [comment on the ticket](https://github.com/craftersoftware/craftercms/issues/617#issuecomment-1813948305).

ui/app/src/utils/content.ts
rart: Not resolved.

ui/app/src/utils/constants.ts
rart: Not resolved.

ui/app/src/utils/content.ts
jvega190: Committed

ui/app/src/utils/constants.ts
jvega190: Just committed

jvega190:
@rart tested with the max config size and there is no noticeable performance downgrade

rart (approved):

PR #3465, opened 2023-11-13
[4502] Item retrieval APIs (item_by*, children_by*) return systemType 'unknown' for config files

jvega190:
- Update types and icons for config and content type files 

https://github.com/craftercms/craftercms/issues/4502


rart (changes requested):
What about the tooltip for the hover of the icon?
image

Be thorough.

jvega190:
Added the tooltip translation, I did check the tooltip (that was saying `application/xml`) but didn't notice that it was the fallback for the lack of translation

rart (approved):

PR #3467, opened 2023-11-14
[6288][4.1.x] File-name control is removing the first char when renaming pages.

jvega190:
https://github.com/craftercms/craftercms/issues/6288


jvega190:
Closing, wrong target branch

PR #3471, opened 2023-11-14
[stability] Disable backdrop click and escape keydown in ConfirmDialog

jvega190:
disable backdrop click and escape keydown in ConfirmDialog


ui/app/src/components/ConfirmDialog/ConfirmDialog.tsx
```diff
@@ -27,6 +27,8 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
   const { formatMessage } = useIntl();
   return (
     <AlertDialog
+      disableBackdropClick
```
jvega190, line 30: @rart I think that the confirm dialog should never be dismissed unless clicking one of the actions, what do you think?

rart (approved):

rart:
@jvega190 please send to the relevant support branches.

PR #3478, opened 2023-11-15
[6288][4.2.x] Firefox issue not triggering rename message 

jvega190:
https://github.com/craftercms/craftercms/issues/6288


rart (changes requested):
Changing the call to the getTopLegacyWindow to directly call the contentWindow could be leaving some cases of nested forms uncovered.

I need to understand this better. Prepare to explain the issue in detail and ping me to discuss.

rart (approved):

PR #3479, opened 2023-11-15
[6285][4.0.x] Remove language names from translation message files

jvega190:
https://github.com/craftercms/craftercms/issues/6285


ui/app/src/components/PathNavigator/PathNavigator.tsx
```diff
@@ -331,7 +331,7 @@ export function PathNavigator(props: PathNavigatorProps) {
     if (type === 'language') {
       const locales = siteLocales.localeCodes?.map((code) => ({
         id: `locale.${code}`,
-        label: formatMessage(languages[code.toLowerCase()])
+        label: languages[code.toLowerCase()]
       }));
       setWidgetMenu({
```
rart: Is there currently a way of using this? I'm thinking we should clean out this code and avoid the languages to even be bundled if they're not in use.

ui/legacy/src/components/cstudio-forms/controls/locale-selector.js
```diff
@@ -20,11 +20,7 @@
   }
 
   function formatLanguageMessage(id) {
-    if (CrafterCMSNext.i18n.messages.languages[id]) {
-      return CrafterCMSNext.i18n.intl.formatMessage(CrafterCMSNext.i18n.messages.languages[id]);
-    } else {
-      return id;
-    }
+    return CrafterCMSNext.i18n.messages.languages[id] ?? id;
```
rart: This shouldn't even work. You need to remove the languages from the json translation files too. 
Provided it is no longer in the default dictionary (English), this mostly looks untested. How are the languages even getting to the i18n object?

ui/app/src/env/i18n-legacy.ts
```diff
@@ -1118,274 +1118,274 @@ export const controlsCommonMessages = defineMessages({
...
-  uz: { id: 'languages.uz', defaultMessage: 'Uzbek' },
-  ve: { id: 'languages.ve', defaultMessage: 'Venda' },
-  vo: { id: 'languages.vo', defaultMessage: 'Volapük' },
-  wa: { id: 'languages.wa', defaultMessage: 'Walloon' },
-  wo: { id: 'languages.wo', defaultMessage: 'Wolof' },
-  yo: { id: 'languages.yo', defaultMessage: 'Yoruba' },
-  za: { id: 'languages.za', defaultMessage: 'Zhuang; Chuang' },
-  zh: { id: 'languages.zh', defaultMessage: 'Chinese' }
-});
+export const languages = {
```
rart: Where is this used? Make a full inventory so we can decide where it needs to be moved to.

rart (changes requested):

rart (changes requested):
Also, `i18n-legacy` seems to have a bunch of dead code; let's clean that out too. The clean up might be different for each branch.

ui/app/src/components/PathNavigator/PathNavigator.tsx
jvega190: there's no way of using this currently, so yeah, I can remove this languages usage

ui/app/src/env/i18n-legacy.ts
jvega190: languages are being used in 2 places right now, PathNavigator (it may end up removed) and locale-selector control. I was thinking on moving languages from i18n-legacy to utils/languages.ts, and `locale-selector` to use it as `craftercms.utils.languages`

rart: I don't want it bundled if it's just for FE1. Let's put it directly in the FE1 control/ds if it's just used in a single place.

jvega190:
@rart moved this PRs to draft so I can finish testing the removal of dead code in i18-legacy

rart (approved):
Cleaning out the translation files (i.e es.json, de.json, ko.json) was missed but I can do that from our translation manager software with more ease now that the thing grew to remove other unused translations too.

PR #3487, opened 2023-11-22
Fix graphiql/explorer-plugin not updating on actions

jvega190:


jvega190:
This only applies on 4.2.x (issue not present in 4.1.x or older versions)

rart (approved):

PR #3489, opened 2023-11-27
[504E] Lock Engine in Preview mode with a token

jvega190:
https://github.com/craftersoftware/craftercms/issues/504

### DIFF
--- a/ui/app/src/services/security.ts
+++ b/ui/app/src/services/security.ts
@@ -14,8 +14,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { get, postJSON } from '../utils/ajax';
-import { pluck } from 'rxjs/operators';
+import { get, post, postJSON } from '../utils/ajax';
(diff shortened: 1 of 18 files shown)


ui/app/src/services/security.ts
```diff
@@ -31,3 +31,7 @@ export function getUserPermissions(site: string, path: string): Observable<strin
     pluck('response', 'permissions')
   );
 }
+
+export function previewSwitch(): Observable<true> {
+  return post(`/studio/api/2/security/preview/switch`).pipe(map(() => true));
```
rart, line 36:
```suggestion
  return post(`/studio/api/2/security/preview/switch`).pipe(mapTo(true));
```

rart (changes requested):

ui/app/src/services/security.ts
jvega190: per conversation, we won't update `mapTo`, as its usage is deprecated

rart (approved):

PR #3490, opened 2023-11-29
[6247] Cleanup unused APIs in Studio API 1

jvega190:
https://github.com/craftercms/craftercms/issues/6247


rart (changes requested):
List all the breaking changes (file/method removals, etc) in changelog.md

static-assets/components/cstudio-common/common-api.js
```diff
@@ -3486,16 +3412,14 @@ var nodeOpen = false,
...
-          .subscribe(
-            function (response) {
-              callback.success({ permissions: response });
-            },
-            function (response) {
-              callback.failure(response);
-            }
-          );
+        craftercms.services.users.fetchMyPermissions(site).subscribe({
+          next: (response) => {
```
rart, line 3416: You mentioned this response format is different from the previous, but you are not matching the previous response here. Are the consumers matching the new signature?

jvega190: The format didn't change, the difference is that the api v1 service returned both permissions for current user and global permissions, and v2 service returns only the permissions for current user (we have a separate API for global permissions). Checking the usage of this service, is is only for user permissions, so there is no other change needed

rart (approved):

PR #3491, opened 2023-11-29
[6305] Base repository path with macros not working when creating components using components datasource

jvega190:
https://github.com/craftercms/craftercms/issues/6305

### DIFF
--- a/ui/legacy/src/components/cstudio-forms/data-sources/components.js
+++ b/ui/legacy/src/components/cstudio-forms/data-sources/components.js
@@ -236,7 +236,7 @@
         true,
         type === 'embedded',
         (item) => item.type === 'component',
-        this.baseRepoPath ?? this.defaultBaseRepoPath
+        this._processPathsForMacros(this.baseRepoPath) ?? this.defaultBaseRepoPath
       );
     },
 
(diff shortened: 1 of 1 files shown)


ui/legacy/src/components/cstudio-forms/data-sources/components.js
```diff
@@ -236,7 +236,7 @@
         true,
         type === 'embedded',
         (item) => item.type === 'component',
-        this.baseRepoPath ?? this.defaultBaseRepoPath
+        this._processPathsForMacros(this.baseRepoPath) ?? this.defaultBaseRepoPath
```
rart: Would the process macros ever return nullish? If baseRepoPath is nullish, what will the function return?

ui/legacy/src/components/cstudio-forms/data-sources/components.js
```diff
@@ -357,7 +357,9 @@
 
     _openContentTypeForm(contentType, type, control) {
       const self = this;
-      const path = this.baseRepoPath ?? craftercms.utils.content.generateComponentBasePath(contentType);
+      const path =
+        this._processPathsForMacros(this.baseRepoPath) ??
```
rart: As above.

ui/legacy/src/components/cstudio-forms/data-sources/components.js
jvega190: it doesn't return nullish, added a validation prior the process call

rart (approved):

PR #3522, opened 2023-12-12
[6299][4.2.x] Long keyword filter values and special characters cause search to fail

jvega190:
https://github.com/craftercms/craftercms/issues/6299


ui/app/src/components/Search/URLDrivenSearch.tsx
```diff
@@ -209,7 +209,7 @@ export function URLDrivenSearch(props: URLDrivenSearchProps) {
       sortOrder={queryParams['sortOrder'] as string}
       currentView={currentView}
       embedded={embedded}
-      keyword={keyword}
+      keyword={Array.isArray(keyword) ? keyword.join(' ') : keyword}
```
jvega190, line 212: URLDrivenSearch set the keyword by getting the urlParam, so it may be an array of strings. Moved the logic here so the SearchUI will always get a string as the keyword

rart (approved):

PR #3548, opened 2023-12-20
[653E][4.2.x]

jvega190:
https://github.com/craftersoftware/craftercms/issues/653


templates/web/common/page-fragments/head.ftl
```diff
@@ -52,7 +52,7 @@
   <script src="/studio/static-assets/yui/resize/resize-min.js"></script>
 
   <script src="/studio/static-assets/components/cstudio-common/common-api.js"></script>
-  <script src="/studio/static-assets/libs/jquery/dist/jquery-3.4.1.min.js"></script><#--File names have version info & not changed across builds. No need for cache buster.-->
```
jvega190, line 55: @rart naming for caching is not a problem anymore, right?

ui/npm-content/build.sh
```diff
@@ -65,7 +65,7 @@ echo "Building Monaco editor..."
 yarn build:monaco
 echo "Monaco editor build complete"
 
-# jquery
+# jquery - custom build for guest
```
rart:
```suggestion
# jQuery - Custom build for XB
```

rart: Since we now have another purpose for this and we have it in node modules, does the npm package have everything needed to execute the build? i.e. do we still need to to git pull it separately?

rart (changes requested):

ui/npm-content/build.sh
jvega190: No it doesn't have everything, `Gruntfile.js` is not available, and when I tried to copy it and build, it started failing due to some other missing files.

ui/npm-content/build.sh
```diff
@@ -65,7 +65,7 @@ echo "Building Monaco editor..."
 yarn build:monaco
 echo "Monaco editor build complete"
 
-# jquery
+# jquery - custom build for XB
```
rart:
```suggestion
# jQuery - Custom build for XB
```

rart (approved):

PR #3580, opened 2024-01-06
Preview UI stylistic updates

rart:
https://github.com/craftercms/craftercms/issues/6415


sumerjabri (changes requested):
Let's get a ticket in to clearly request changes/enhancements and tie the two together.

sumerjabri (approved):

PR #3583, opened 2024-01-09
[6412][4.1.x] Publishing dialog fails when current date values are supplied in non-offset format

jvega190:
https://github.com/craftercms/craftercms/issues/6412


jvega190:
`Z` stands for the zero UTC offset. So when getting the GMT timezone the formatted string will contain the `Z` at the end instead of the offset. So, when retrieving the offset we now validate the scenario and if the date/time value contains a `Z`, the offset used will be `+00:00`, otherwise, offset calculation will be the same as before.

rart (approved):

PR #3639, opened 2024-01-24
[6430][4.0.x] Static Assets flattens nested folders on bulk upload

jvega190:
- Ensure that there's an ending slash at the end of uppy base path

https://github.com/craftercms/craftercms/issues/6430

### DIFF
--- a/ui/app/src/components/UppyDashboard/UppyDashboard.tsx
+++ b/ui/app/src/components/UppyDashboard/UppyDashboard.tsx
@@ -30,6 +30,7 @@ import { Subject } from 'rxjs';
 import { debounceTime } from 'rxjs/operators';
 import { UppyFile } from '@uppy/utils';
 import { CSSObject } from 'tss-react';
+import { ensureSingleSlash } from '../../utils/string';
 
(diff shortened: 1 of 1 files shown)


ui/app/src/components/UppyDashboard/UppyDashboard.tsx
```diff
@@ -442,7 +443,7 @@ export function UppyDashboard(props: UppyDashboardProps) {
       title,
       id: 'craftercms:Dashboard',
       site,
-      path,
+      path: ensureSingleSlash(`${path}/`),
```
rart, line 446: What's the need for this?

jvega190: `path` is the base path that's used in the dashboard. It may come without a slash at the end, so when appending the rest of the path (in the case of folder structures) it was causing the full path to be incorrect.
For example
- path = `/static-assets/images`
- upload folder structure like `folder1/folder2/image.jpg`
was ending up in: `/static-assets/imagesfolder1/folder2/image.png`

rart (approved):

PR #3648, opened 2024-01-26
[4.1.x][6390] Improve XB Communication status and error experience

jvega190:
- Add timeout in socket snack, do not show socket snack in iframed windows

https://github.com/craftercms/craftercms/issues/6390


rart (changes requested):
This won't do. There's no cancellation, timeout executed even when not needed, etc. Review what was done for PreviewAddressBar with the xbDetectionTimeoutMs.

rart (approved):

PR #3652, opened 2024-01-26
[6453] Compute bucket xml prop as before, handle xml not having the url prop, add video preview

jvega190:
https://github.com/craftercms/craftercms/issues/6453

### DIFF
--- a/static-assets/themes/cstudioTheme/css/forms-default.css
+++ b/static-assets/themes/cstudioTheme/css/forms-default.css
@@ -1012,15 +1012,27 @@ div.yui-calcontainer.cstudio-form-field-calendar {
   padding: 0;
   border: 1px solid rgba(0, 0, 0, 0.5);
   background: rgba(0, 0, 0, 0.6);
-  box-shadow: 0 2px 6px rgba(0, 0, 0, 0.5), inset 0 1px rgba(255, 255, 255, 0.3), inset 0 10px rgba(255, 255, 255, 0.2),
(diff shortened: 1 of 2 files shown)


static-assets/themes/cstudioTheme/css/forms-default.css
```diff
@@ -1012,15 +1012,27 @@ div.yui-calcontainer.cstudio-form-field-calendar {
   padding: 0;
   border: 1px solid rgba(0, 0, 0, 0.5);
   background: rgba(0, 0, 0, 0.6);
-  box-shadow: 0 2px 6px rgba(0, 0, 0, 0.5), inset 0 1px rgba(255, 255, 255, 0.3), inset 0 10px rgba(255, 255, 255, 0.2),
-    inset 0 10px 20px rgba(255, 255, 255, 0.25), inset 0 -15px 30px rgba(0, 0, 0, 0.3);
-  -o-box-shadow: 0 2px 6px rgba(0, 0, 0, 0.5), inset 0 1px rgba(255, 255, 255, 0.3),
-    inset 0 10px rgba(255, 255, 255, 0.2), inset 0 10px 20px rgba(255, 255, 255, 0.25),
+  box-shadow:
```
jvega190, line 1015: @rart I got some formatted lines like with by prettier auto format, is that ok or should I revert?

rart (changes requested):
Compare with the very latest version on 4.1.x. There's at least one difference.

rart (changes requested):
Also recheck styles match

rart (approved):

PR #3655, opened 2024-01-30
[4.1.x][6491] Deleting all children of an expanded folder and then uploading an item, will set the folder into loading state

jvega190:
https://github.com/craftercms/craftercms/issues/6491


ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -162,7 +162,7 @@ export function deleteItemFromState(tree: PathNavigatorTreeStateProps, targetPat
   tree.expanded = tree.expanded.filter(
     // If the parent is left without children, remove from expanded too.
     totalByPath[parentPath] === 0
-      ? (expandedPath) => expandedPath !== targetPath || expandedPath !== parentPath
+      ? (expandedPath) => expandedPath !== targetPath && expandedPath !== parentPath
```
jvega190, line 165: Condition was already there, but it needed to be an and instead of an or, since for any of the cases it needs to be filtered/removed from state

rart (approved):

PR #3675, opened 2024-02-02
[4.1.x/6499] Fix legacy ICE not working

rart:
craftercms/craftercms#6499

### DIFF
--- a/ui/app/src/components/Host/Host.tsx
+++ b/ui/app/src/components/Host/Host.tsx
@@ -62,7 +62,7 @@ export function Host() {
 
     const h2g = getHostToGuestBus()
       .asObservable()
-      .pipe(map((action) => ({ ...action, craftercms: true, source: 'host' })))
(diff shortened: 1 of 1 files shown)


sumerjabri (approved):

ui/app/src/components/Host/Host.tsx
```diff
@@ -62,7 +62,7 @@ export function Host() {
 
     const h2g = getHostToGuestBus()
       .asObservable()
-      .pipe(map((action) => ({ ...action, craftercms: true, source: 'host' })))
+      .pipe(map((action) => ({ ...action, meta: { craftercms: true, source: 'host' } })))
```
rart, line 65: The cause: 
- https://github.com/craftercms/studio-ui/blob/33545768b3e4189b0f27c94aea72baa60e80ec15/static-assets/scripts/communicator.js#L252
- During recent changes, the meta object got lost and its props were added to the root payload object

PR #3679, opened 2024-02-05
[695E] Update auth service incorrect property

jvega190:
https://github.com/craftersoftware/craftercms/issues/695


ui/app/src/services/users.ts
```diff
@@ -217,7 +217,7 @@ export function hasPermissions(site: string, ...permissions: string[]): Observab
 }
 
 export function hasPermission(site: string, permission: string): Observable<boolean> {
-  return hasPermissions(site, permission).pipe(map((response) => response?.permission));
```
jvega190, line 220: this service and `hasGlobalPermission` are not in use, should we remove then or leave them (as they are possible usages in the future)?

rart: Leave

rart (approved):

PR #3689, opened 2024-02-06
[4.1.x][6514] RTE: click the list button doesn't produce the list-unlist result in the Form Engine.

jvega190:
https://github.com/craftercms/craftercms/issues/6514


jvega190:
@rart the list issue was caused because the content was initialized/set in 'raw' format, so when processing the list, there were some text nodes (`\n`) causing a problem. By setting the format in 'html' (as in XB's RTE), the problem doesn't happen anymore

rart (approved):
Pls test this in many different blueprints

PR #3694, opened 2024-02-08
[4.1.x][6518] Level Descriptor being counted as children, also ...X more items button not working when publishing

jvega190:
https://github.com/craftercms/craftercms/issues/6518

### DIFF
--- a/ui/app/src/state/epics/pathNavigatorTree.ts
+++ b/ui/app/src/state/epics/pathNavigatorTree.ts
@@ -210,9 +210,9 @@ export default [
           optionsByPath = {
             ...optionsByPath,
             ...createPresenceTable(expanded, (value) => ({
+              ...createGetChildrenOptions(chunk, pluckProps(payload, true, 'limit', 'excludes')),
               ...(keywordByPath[value] ? { keyword: keywordByPath[value] } : {}),
-              ...(offsetByPath[value] ? { limit: limit + offsetByPath[value] } : {}),
-              ...createGetChildrenOptions(chunk, pluckProps(payload, true, 'limit', 'excludes'))
+              ...(offsetByPath[value] ? { limit: limit + offsetByPath[value] } : {})
             }))
           };
         });
--- a/ui/app/src/state/reducers/pathNavigatorTree.ts
+++ b/ui/app/src/state/reducers/pathNavigatorTree.ts
@@ -212,9 +212,16 @@ const reducer = createReducer<LookupTable<PathNavigatorTreeStateProps>>(
         ...state[id].offsetByPath,
         [path]: 0
       };
+      // When the item contains a levelDescriptor, it is rendered as another child. When children length equals or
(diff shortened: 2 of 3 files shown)


ui/app/src/state/epics/pathNavigatorTree.ts
```diff
@@ -210,9 +210,9 @@ export default [
           optionsByPath = {
             ...optionsByPath,
             ...createPresenceTable(expanded, (value) => ({
+              ...createGetChildrenOptions(chunk, pluckProps(payload, true, 'limit', 'excludes')),
```
jvega190, line 213: Those are the global options for children, they need to go before the specific options per path

ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -212,9 +212,16 @@ const reducer = createReducer<LookupTable<PathNavigatorTreeStateProps>>(
         ...state[id].offsetByPath,
         [path]: 0
       };
+      // When the item contains a levelDescriptor, it is rendered as another child. When children length equals or
+      // surpasses the limit, the amount of children rendered is the limit + 1, to account for the levelDescriptor.
+      // So, when collapsing and doing a splice of the childrenByParentPath, we need to consider the levelDescriptor.
+      const containsDescriptor = state[id].childrenByParentPath[path].some((path) =>
+        path.includes('crafter-level-descriptor.level.xml')
```
rart: I don't believe the level descriptor is necessarily called this way. @sumerjabri can it be a different name?

Maybe we need to account for it from the moment the fetch payload arrives.

@sumerjabri if the back won't account for the lev desc being on its own, should we just put it back with the children? This logic is not ideal.

rart (changes requested):

ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -52,14 +52,14 @@ const updatePath = (state, payload) => {
...
   chunk.childrenByParentPath[parentPath] = [];
-  children.forEach((item) => {
-    chunk.childrenByParentPath[parentPath].push(item.path);
-    chunk.totalByPath[item.path] = item.childrenCount;
-  });
   if (children.levelDescriptor) {
     chunk.childrenByParentPath[parentPath].push(children.levelDescriptor.path);
     chunk.totalByPath[children.levelDescriptor.path] = 0;
   }
+  children.forEach((item) => {
```
jvega190, line 59: This was causing the children to be rendered before the levelDescriptor in some scenarios.

ui/app/src/state/reducers/pathNavigatorTree.ts
jvega190: Per decision, collapse logic to reset state was removed to avoid repeating children calls

rart (approved):

PR #3701, opened 2024-02-15
[4.1.x][6542] Drag and drop content item into the trash bin returns error for publisher or author

jvega190:
https://github.com/craftercms/craftercms/issues/6542

### DIFF
--- a/ui/app/src/services/content.ts
+++ b/ui/app/src/services/content.ts
@@ -660,36 +660,40 @@ export function deleteItem(
   indexToDelete: number | string,
   path: string
 ): Observable<any> {
-  return performMutation(
-    site,
-    path,
-    (element) => {
-      let index = indexToDelete;
-      let fieldNode = element.querySelector(`:scope > ${fieldId}`);
+  return lock(site, path).pipe(
+    switchMap(() => {
+      return performMutation(
+        site,
+        path,
+        (element) => {
+          let index = indexToDelete;
+          let fieldNode = element.querySelector(`:scope > ${fieldId}`);
 
-      if (typeof indexToDelete === 'string') {
-        index = parseInt(popPiece(indexToDelete));
-        // A fieldId can be in the form of `a.b`, which translates to `a > item > b` on the XML.
-        // In terms of index, since all it should ever arrive here is collection items,
-        // this assumes the index path points to the item itself, not the collection.
-        // By calling removeLastPiece(indexToDelete), we should get the collection node here.
-        fieldNode = extractNode(element, fieldId, removeLastPiece(`${indexToDelete}`));
-      }
+          if (typeof indexToDelete === 'string') {
+            index = parseInt(popPiece(indexToDelete));
+            // A fieldId can be in the form of `a.b`, which translates to `a > item > b` on the XML.
+            // In terms of index, since all it should ever arrive here is collection items,
+            // this assumes the index path points to the item itself, not the collection.
+            // By calling removeLastPiece(indexToDelete), we should get the collection node here.
+            fieldNode = extractNode(element, fieldId, removeLastPiece(`${indexToDelete}`));
+          }
 
-      const $fieldNode = $(fieldNode);
+          const $fieldNode = $(fieldNode);
 
-      $fieldNode
-        .children()
-        .eq(index as number)
-        .remove();
+          $fieldNode
+            .children()
+            .eq(index as number)
+            .remove();
 
-      if ($fieldNode.children().length === 0) {
-        // If the node isn't completely blank, the xml formatter won't do it's job in converting to a self-closing tag.
-        // Also, later on, when retrieved, some *legacy* functions would impaired as the deserializing into JSON had unexpected content
-        $fieldNode.html('');
-      }
-    },
-    modelId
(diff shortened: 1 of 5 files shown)


ui/app/src/services/content.ts
```diff
@@ -660,36 +660,40 @@ export function deleteItem(
...
-
-      if (typeof indexToDelete === 'string') {
-        index = parseInt(popPiece(indexToDelete));
-        // A fieldId can be in the form of `a.b`, which translates to `a > item > b` on the XML.
-        // In terms of index, since all it should ever arrive here is collection items,
-        // this assumes the index path points to the item itself, not the collection.
-        // By calling removeLastPiece(indexToDelete), we should get the collection node here.
-        fieldNode = extractNode(element, fieldId, removeLastPiece(`${indexToDelete}`));
-      }
+  return lock(site, path).pipe(
```
jvega190: A lock is needed before writing, the issue was not happening because in admin because admin can unlock and it doesn't matter the lock owner. For author/publisher/etc, since it was not locking first, the lock owner wasn't itself then it was not being able to unlock.

rart: Where is the unlock call happening?

jvega190: performMutation does a `writeContent` with an `unlock` parameter set to true. So at the end of the writing process it unlocks the content

rart: Please look at previous releases and identify if this was a change in the API behaviour and report back.

jvega190: It seems like it hasn't changed. Is the same behavior since 4.0.0

rart: This probably breaks the zone menu delete. Did you test that? Best case scenario, there's a redundant lock call.

rart (changes requested):

ui/app/src/services/content.ts
rart: This might have been a good pattern if all worked this way. For now, we should stick to the way it works all over which is the consumer issues the lock apart, before issuing the op. Perhaps we should use the beforeWrite$.

jvega190: I wasn't breaking the ZoneMenu delete, though there were 2 lock calls. Updated for it to use beforeWrite$

ui/guest/src/store/epics/root.ts
```diff
@@ -669,15 +669,34 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
...
+        const { username, activeSite } = state;
+        const models = getCachedModels();
+        let parentModelId = getParentModelId(modelId, models, modelHierarchyMap);
+        let path = models[modelId].craftercms.path ?? models[parentModelId].craftercms.path;
+
+        if (isInheritedField(modelId, fieldId)) {
+          const modelIdToLock = getModelIdFromInheritedField(modelId, fieldId);
+          const parentModelId = getParentModelId(modelIdToLock, models, modelHierarchyMap);
+          path = models[parentModelId ?? modelIdToLock].craftercms.path;
+        }
```
rart: Let's try to refactor this stuff so its not duplicated

ui/guest/src/store/epics/root.ts
```diff
@@ -669,15 +669,34 @@ const epic = combineEpics<GuestStandardAction, GuestStandardAction, GuestState>(
...
+        beforeWrite$({
```
rart: This returns an observable, shouldn't we be switchMapping to the beforeWrite$ here?

rart (changes requested):

ui/guest/src/store/epics/root.ts
jvega190: The beforeWrite$ is inside a `tap`, and the deleteItem (previously called with no return) is inside the subscribe part of the beforeWrite$.

rart: I can see the code. I'm asking we should rather be doing what I mentioned above.

jvega190: Got it, yeah I switchMapped beforeWrite$

jvega190: One thing that changes is that the computedDragEnd gets fired after beforeWrite$ actions, but it does get triggered properly

rart (approved):

ui/app/src/utils/content.ts
```diff
@@ -1092,3 +1093,22 @@ export function generateComponentBasePath(contentType: string) {
...
+ * If the field is inherited, swaps the modelId and parentModelId with
+ * the inheritance parent's. */
+export function getInheritanceParentIdsForField(
+  fieldId: string,
+  modelLookup: LookupTable<ContentInstance>,
+  modelId: string,
+  parentModelId: string,
+  modelIdByPath: LookupTable<string>,
+  hierarchyMap: ModelHierarchyMap
+): { modelId: string; parentModelId: string } {
```
rart, line 1107: @jvega190 I made some updates. You didn't add types and the function name was way too vague. Also the doc comment is always appreciated. Other minor changes on the other files; review changes and port to develop.

PR #3705, opened 2024-02-16
[4.1.x][6533] Node selector control should only provide the user with an Edit option for content items that are editable by Studio

jvega190:
https://github.com/craftercms/craftercms/issues/6533

### DIFF
--- a/static-assets/components/cstudio-forms/controls/node-selector.js
+++ b/static-assets/components/cstudio-forms/controls/node-selector.js
@@ -365,7 +365,7 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField, 
         itemEl.classList.add('disabled');
       }
 
-      const isComponent = item.key.includes('/site') || item.inline;
+      const isComponent = item.key.startsWith('/site') || item.inline;
       const editBtnLabel = this.readonly ? 'View' : 'Edit';
       const editBtnIconClass = this.readonly ? 'fa-eye' : 'fa-pencil';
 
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-forms/controls/node-selector.js
```diff
@@ -355,17 +355,19 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField,
       $(itemEl).append(`<span class="name">${item.value}</span>`);
       if (item.include) {
         $(itemEl).append(`<span class="path">${item.include}</span>`);
-      } else {
+      } else if (item.inline) {
```
rart: Certain that this comes as a boolean?

static-assets/components/cstudio-forms/data-sources/file-desktop-upload.js
```diff
@@ -101,48 +101,18 @@ YAHOO.extend(CStudioForms.Datasources.FileDesktopUpload, CStudioForms.CStudioFor
...
-
-    var site = CStudioAuthoringContext.site;
-    var path = this._self.repoPath;
-    var isUploadOverwrite = true;
-
-    for (var i = 0; i < this.properties.length; i++) {
-      if (this.properties[i].name === 'repoPath') {
-        path = this.properties[i].value;
-
-        path = this.processPathsForMacros(path);
```
rart, line 115: What about all this removed code, including parsing the path macros, etc?

rart (changes requested):

static-assets/components/cstudio-forms/controls/node-selector.js
jvega190: It's not, but only the embedded components have the property. Though I'll better validate the string in case the prop evolves

static-assets/components/cstudio-forms/data-sources/file-desktop-upload.js
jvega190: The file upload behavior when editing was to upload/override the item (removed code). From the ticket, the behavior should be displaying the editor (if editable). That's what the browse-file datasource does.

rart (approved):

PR #3707, opened 2024-02-16
[4.1.x][6541] Checkbox with default value in repeat group will affect items in the same group when "Add Another"

jvega190:
https://github.com/craftercms/craftercms/issues/6541

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -2598,7 +2598,12 @@ var CStudioForms =
               }
 
               const defaultValue = moduleConfig.config.field.defaultValue;
-              if (!value && defaultValue && typeof defaultValue === 'string' && !noDefaultLookup[formField.id]) {
+              if (
+                craftercms.utils.object.nou(value) &&
+                defaultValue &&
+                typeof defaultValue === 'string' &&
+                !noDefaultLookup[formField.id]
+              ) {
                 value = moduleConfig.config.field.defaultValue;
               }
 
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -2609,7 +2609,7 @@ var CStudioForms =
                 typeof defaultValue === 'string' &&
                 !noDefaultLookup[formField.id]
               ) {
-                value = moduleConfig.config.field.defaultValue;
+                value = defaultValue;
               }
 
               if (value) {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2598,7 +2598,15 @@ var CStudioForms =
               }
 
               const defaultValue = moduleConfig.config.field.defaultValue;
-              if (!value && defaultValue && typeof defaultValue === 'string' && !noDefaultLookup[formField.id]) {
+              // When value is of type string, an empty string is considered unset (unless it is contained in the
+              // `noDefaultLookup`, where it means that an empty string was the value set).
+              const unsetStoredValue = typeof value === 'string' ? !value : craftercms.utils.object.nou(value);
```
jvega190, line 2603: @rart I was wrong about the `no-default` attribute in boolean values. It's only for string values as we have no distinction between empty and not set (as they end up being an empty string). For boolean, the stored value will never be an empty string. So, the issue was that when rendering the field, if the value was set to false by the control, when re-rendering the repeating group it was going to be considered as 'false' in the condition and ended up setting the defaultValue.

static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2598,7 +2603,12 @@ var CStudioForms =
               }
 
               const defaultValue = moduleConfig.config.field.defaultValue;
-              if (!value && defaultValue && typeof defaultValue === 'string' && !noDefaultLookup[formField.id]) {
+              if (
```
jvega190, line 2606: @rart I was wrong about the 'no-default' attribute, it is only used in type string values as we can't distinguish between empty and unset values. Though there was this issue, not only on booleans, where if you had a default value in a rep-group item and you empty the value, then add another item, it was re-setting it to the default value.

static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2598,7 +2603,12 @@ var CStudioForms =
...
+                craftercms.utils.object.nou(value) &&
+                defaultValue &&
+                typeof defaultValue === 'string' &&
+                !noDefaultLookup[formField.id]
+              ) {
                 value = moduleConfig.config.field.defaultValue;
```
rart:
```suggestion
                value = defaultValue;
```

static-assets/components/cstudio-forms/forms-engine.js
rart: Confirm/test this with the rest of controls that they continue to behave correctly.

rart (changes requested):

jvega190:
@rart tested with all controls, inside and outside repeating groups

rart (approved):

PR #3709, opened 2024-02-19
[4.1.x][6550] Editing Date picker fields via XB is broken

jvega190:
https://github.com/craftercms/craftercms/issues/6550

### DIFF
--- a/static-assets/components/cstudio-forms/controls/date-time.js
+++ b/static-assets/components/cstudio-forms/controls/date-time.js
@@ -1152,11 +1152,15 @@ YAHOO.extend(CStudioForms.Controls.DateTime, CStudioForms.CStudioFormField, {
         dateEl,
         'click',
         function (e) {
-          this.show();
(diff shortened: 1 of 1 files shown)


rart (changes requested):
I need this triple checked please. It looks this was address on https://github.com/craftercms/studio-ui/pull/3534 for https://github.com/craftercms/craftercms/issues/6368 and it has come back. Let's make sure this is final.

jvega190:
Checked, the control was rendering properly in the first render, but if you closed/opened again it would re-position incorrectly. Now, it resets the top position on each render to avoid those miscalculations. Also, I'm avoiding opening the control when it's already open.

rart (commented):
Was it tested also in full forms with multiple controls in different positions in relation to the scroll?

jvega190:
Yeah, tested with full forms and positioning several date-pickers in different positions

rart (approved):

PR #3724, opened 2024-02-27
[6584] XB dnd should not trigger any request if item doesn't change its position

jvega190:


ui/guest/src/store/util.ts
rart, line 1: There's too much duplicated code on this file.

ui/guest/src/store/util.ts
```diff
@@ -108,3 +108,45 @@ export const checkIfLockedOrModified = (state: GuestState, record: ElementRecord
...
+export const movedToSamePosition = (dragContext: GuestState['dragContext']): boolean => {
+  let { dragged, targetIndex } = dragContext,
+    record = dragged,
+    draggedElementIndex = record.index;
+
+  if (typeof draggedElementIndex === 'string') {
+    // If the index is a string, it's a nested index with dot notation.
+    // At this point, we only care for the last index piece, which is
+    // the index of this item in the collection that's being manipulated.
+    draggedElementIndex = parseInt(draggedElementIndex.substr(draggedElementIndex.lastIndexOf('.') + 1), 10);
```
rart: Carefully substr to substring

rart (changes requested):

ui/guest/src/store/util.ts
jvega190: The difference is the end parameter (second param). They both return the string starting at position "start"

substr: The number of characters to extract.
substring: The index of the first character to exclude from the returned substring.

Example:
'hello_world'.substr(3, 6) => it will get 6 characters starting from character at pos 3 => ''lo_wor"

'hello_world'.substr(3, 6) => it will exclude everything after position 6 of the initial string ("w") => "lo_"

If no end parameter is provided, it will return the remaining part of the string starting at "start" parameter position.

For this scenario, what we want is everything after the last '.', so by just doing 
```draggedElementIndex.substr(draggedElementIndex.lastIndexOf('.') + 1)```
is enough right?

jvega190: Re-requested review with this update

rart (changes requested):
- See updates on 4.1.x PR.
- Fix conflicts

rart (approved):

PR #3728, opened 2024-02-28
[4.1.x][6514] Set html content with proper format when initializing RTE

jvega190:
https://github.com/craftercms/craftercms/issues/6514

### DIFF
--- a/ui/guest/src/controls/rte.ts
+++ b/ui/guest/src/controls/rte.ts
@@ -239,9 +239,15 @@ export function initTinyMCE(
       }
 
       function replaceLineBreaksIfApplicable(content: string) {
-        // Replace line breaks with <br> for textarea fields
-        // Address line breaks in textarea fields: https://github.com/craftercms/craftercms/issues/6432
-        type === 'textarea' && editor.setContent(content.replaceAll('\n', '<br>'), { format: 'html' });
+        if (type === 'textarea') {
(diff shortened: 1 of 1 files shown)


ui/guest/src/controls/rte.ts
```diff
@@ -239,9 +239,15 @@ export function initTinyMCE(
...
-        // Replace line breaks with <br> for textarea fields
-        // Address line breaks in textarea fields: https://github.com/craftercms/craftercms/issues/6432
-        type === 'textarea' && editor.setContent(content.replaceAll('\n', '<br>'), { format: 'html' });
+        if (type === 'textarea') {
+          // Replace line breaks with <br> for textarea fields
+          // Address line breaks in textarea fields: https://github.com/craftercms/craftercms/issues/6432
+          editor.setContent(content.replaceAll('\n', '<br>'), { format: 'html' });
+        } else if (type === 'html') {
+          // Set content in 'html' format to clean line breaks and avoid editor issues
+          // Address list/unlist issue: https://github.com/craftercms/craftercms/issues/6514
```
rart:
```suggestion
          // Set content in 'html' format for the editor to exec its internal cleanup mechanisms
          // For example, removal of potentially problematic line breaks which we're seeing cause the list plugin to crash (https://github.com/craftercms/craftercms/issues/6514)
```

rart (approved):

PR #3733, opened 2024-02-29
[4.1.x][6584] XB dnd should not trigger any request if item doesn't change its position

jvega190:


jvega190:

rart (changes requested):
Conflicts

rart (approved):

PR #3735, opened 2024-03-01
[4.1.x][6525] The config editor manages site-policy-config.xml as env specific but site-policy-config.xml is meant to be site wide

jvega190:
https://github.com/craftercms/craftercms/issues/6525


jvega190:

rart (approved):

PR #3742, opened 2024-03-06
[4.1.x][6597] Revert to previous doesn't refresh the current item

jvega190:
https://github.com/craftercms/craftercms/issues/6597


jvega190:

rart (approved):

PR #3761, opened 2024-03-14
[4.1.x][6527] Cut/Paste via the Sidebar should warn users or allow them to act on broken dependencies caused by the move operation

jvega190:
https://github.com/craftercms/craftercms/issues/6527


jvega190:
Left as draft as this is 4.1.5 work

rart (approved):

PR #3774, opened 2024-03-19
[5947] Give a developer the ability to create and select subfolders when creating new templates

jvega190:
https://github.com/craftercms/craftercms/issues/5947


jvega190:
Per conversation we'll keep this PR as draft since there's some backend work needed for this to fully work.

PR #3780, opened 2024-03-25
[6645] Valid email address should follow RFC rules

jvega190:
https://github.com/craftercms/craftercms/issues/6645

### DIFF
--- a/ui/app/src/components/UserManagement/utils.ts
+++ b/ui/app/src/components/UserManagement/utils.ts
@@ -30,7 +30,11 @@ export const minLengthMap = {
 };
 
 export const isInvalidEmail = (email: string) => {
-  const emailRegex = /^([\w\d._\-#])+@([\w\d._\-#]+[.][\w\d._\-#]+)+$/g;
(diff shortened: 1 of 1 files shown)


ui/app/src/components/UserManagement/utils.ts
```diff
@@ -30,7 +30,11 @@ export const minLengthMap = {
 };
 
 export const isInvalidEmail = (email: string) => {
-  const emailRegex = /^([\w\d._\-#])+@([\w\d._\-#]+[.][\w\d._\-#]+)+$/g;
+  // From https://emailregex.com/, according to the RFC 5322 standard
+  const emailRegex =
+    // Linter shows a couple of warnings, but the regex is correct
```
rart: What are the linter warnings? Are they unnecessary escaped chars? Remove those then.
Anything else we can either fix or discuss. Anything that's ignored, should be explicitly reported in a comment and the reason to ignore it.

rart (changes requested):

rart (approved):

PR #3798, opened 2024-04-03
[6205] Placeholder image generation util

rart:
craftercms/craftercms#6205

### DIFF
--- a/ui/app/src/utils/content.ts
+++ b/ui/app/src/utils/content.ts
@@ -1126,3 +1126,49 @@ export function getInheritanceParentIdsForField(
   }
   return ids;
 }
+
+export interface GeneratePlaceholderImageDataUrlArgs {
+  width: number;
+  height: number;
+  fillStyle: string;
+  textFillStyle: string;
+  text: string;
+  textPositionX: number;
+  textPositionY: number;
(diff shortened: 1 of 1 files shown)


ui/app/src/utils/content.ts
```diff
@@ -1126,3 +1126,49 @@ export function getInheritanceParentIdsForField(
...
+  textAlign: CanvasTextAlign;
+}
+
+export function generatePlaceholderImageDataUrl(attributes?: Partial<GeneratePlaceholderImageDataUrlArgs>): string {
+  let attrs: GeneratePlaceholderImageDataUrlArgs = Object.assign(
+    {
+      width: 300,
+      height: 150,
+      fillStyle: '#f0f0f0',
+      text: 'Sample Image',
```
sumerjabri, line 1148: Is this parametrized or hard coded to that string?

sumerjabri (changes requested):

ui/app/src/utils/content.ts
rart: All these are defaults. Whatever comes in the `attributes` argument prevails.

sumerjabri (approved):

PR #3805, opened 2024-04-08
[6652] Components Data Source Enhancements

jvega190:
https://github.com/craftercms/craftercms/issues/6652


rart (changes requested):
There's a lot more to port for this ticket.

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -227,11 +227,15 @@ export function PreviewSearchPanel() {
...
-      });
+      hostToGuest$.next(
+        componentInstanceDragStarted({
+          instance,
+          contentType: contentTypesLookup[instance.craftercms.contentTypeId]
+        })
+      );
     } else {
+      // TODO: This action's payload doesn't match the payload definition on the action typing.
+      //   Fix/match typing and usages, both from dispatchers and handlers of this action.
```
rart: Fix

rart (changes requested):

ui/guest/src/react/ExperienceBuilder.tsx
```diff
@@ -342,7 +342,7 @@ function ExperienceBuilderInternal(props: InternalGuestProps) {
           dispatch(action);
           break;
         case assetDragStarted.type:
-          dispatch(assetDragStarted({ asset: payload }));
+          dispatch(assetDragStarted(payload));
```
rart: Looks like this can simply be moved to the whitelisted actions.

Also, please confirm you've reviewed the Payload typings match the usages not just at the dispatchers but at the *handlers*

rart (changes requested):

ui/guest/src/react/ExperienceBuilder.tsx
jvega190: Yeah, I reviewed all usages of the action, including handlers.
Moving the case to the whitelisted actions.

rart (approved):

PR #3823, opened 2024-04-19
Compatibility list update

rart:


ui/app/src/components/PreviewConcierge/compatibilityList.ts
```diff
@@ -14,6 +14,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const list = ['4.1.2', '4.1.1', '4.1.0', '4.0.3'];
+const list = ['4.2.0'];
```
sumerjabri, line 17: Are we no longer compatible with 4.0 and 4.1?

sumerjabri (changes requested):

sumerjabri:
@rart bump

sumerjabri:
@rart confirms the Authoring JS SDK (XB for non-FreeMarker sites) is no longer compatible with older releases. Most things will work, but some things will break. Users should recompile against the compatible 4.2.0 SDK.

This is specific to developers developing _authoring_ XB using the Authoring JS SDK. This doesn't affect _delivery_ or FreeMarker-based sites.

ui/app/src/components/PreviewConcierge/compatibilityList.ts
sumerjabri: The answer from @rart is yes.

sumerjabri (approved):

PR #3837, opened 2024-04-29
[6704] Dropdown form control field should support values of type _i and _f

jvega190:
https://github.com/craftercms/craftercms/issues/6704


rart (changes requested):
See updates on 4.1.x

rart (approved):

PR #3838, opened 2024-04-29
[4.1.x][6704] Dropdown form control field should support values of type _i and _f

jvega190:
https://github.com/craftercms/craftercms/issues/6704


rart (changes requested):
The dropdown control on the type editor has a popup tooltip regarding the suffixes. As I mentioned on the ticket, that needs to be updated.

jvega190:
> The dropdown control on the type editor has a popup tooltip regarding the suffixes. As I mentioned on the ticket, that needs to be updated.

@rart yes, by updating the `supportedPostFixes` property in the control, the tooltip gets the info to display the available suffixes

rart (approved):

PR #3843, opened 2024-04-30
[6702] Image cropper increases size of an image relative to the original

jvega190:
https://github.com/craftercms/craftercms/issues/6702


static-assets/components/cstudio-dialogs/crop-dialog.js
```diff
@@ -363,7 +363,7 @@ CStudioAuthoring.Dialogs.CropDialog = CStudioAuthoring.Dialogs.CropDialog || {
         self = self;
 
       const canvas = $image.cropper('getCroppedCanvas');
-      var dataUrl = canvas.toDataURL();
+      const dataUrl = canvas.toDataURL(imageData.meta.type);
```
jvega190, line 366: Default type value for `toDataURL` is `image/png`, causing the size of jpg images to be bigger than the original

rart (approved):

PR #3854, opened 2024-05-08
[4.1.x][6707] Validate macros in Content Type form before saving

jvega190:
https://github.com/craftercms/craftercms/issues/6707
https://github.com/craftercms/craftercms/issues/6705


jvega190:
https://github.com/craftercms/craftercms/issues/6705 was worked in the same PR for develop, but the ticket is not labeled for 4.1.x.

rart (approved):

PR #3879, opened 2024-05-14
[6789] Dependency management 5/24 

jvega190:
Downgrade @prettier/plugin-xml to v3.3.0 (Issue introduced by version 3.3.1).

https://github.com/craftercms/craftercms/issues/6789


jvega190:
This downgrade is due to an issue introduced in `@prettier/plugin-xml` in v3.3.1. It adds extra spaces causing the content to break

rart (changes requested):
Use @3.3.0, no `^`. No manual edits to `yarn.lock`.

rart (approved):

rart:
https://github.com/prettier/plugin-xml/issues/784

PR #3909, opened 2024-05-23
[4.1.x][5772] Use Vite & SWC to produce build bundles

jvega190:


sumerjabri:
This is now in 4.2, closing this.

PR #3919, opened 2024-05-28
[4.1.x][6584] XB D&D should not trigger any request if item doesn't change its position

jvega190:


ui/guest/src/store/util.ts
```diff
@@ -141,7 +141,7 @@ export const getMoveComponentInfo = (dragContext: GuestState['dragContext']) =>
     if (draggedElementIndex < targetIndex) {
       // Hence the final target index in reality is
       // the drop marker's index minus 1
-      newTargetIndex = targetIndex--;
+      newTargetIndex = --targetIndex;
```
rart:
```suggestion
      newTargetIndex = targetIndex - 1;
```

rart (changes requested):

jvega190:
@rart done, te-tested with the change to avoid mutating the state

rart (approved):

PR #3932, opened 2024-05-29
[5772]  Update legacy code to work with new mechanics

jvega190:


ui/app/src/CHANGELOG.md
```diff
@@ -40,6 +40,8 @@
 * Migrated the Studio UI build to Vite/SWC
 * Rollup's XB build to use SWC
 * Upgraded target compilation to ES2022, dropping many code transforms for features that are supported by most modern browsers such as nullish coalescing, optional chaining, object spreading and destructuring.
+* Remove legacy `browseCMIS` dialog.
```
rart: In the new PR for the legacy removals, please be more specific of all things removed. For example...
```suggestion
* Remove legacy `browseCMIS` dialog and `openCMISBrowse` function from common-api.
```

static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -55,468 +55,450 @@ var setFormSize = function (height, id) {
...
-    };
-
-    CStudioFormDatasource.prototype = {
-      /**
-       * type of data returned by datasource
-       */
-      getInterface: function () {
-        return '';
-      },
+var CStudioForms = CStudioForms;
```
rart: What's this for?

templates/web/browseS3.ftl
```diff
@@ -161,11 +161,15 @@
       });
     </script>
 
+    <#include "/static-assets/app/pages/legacy.html">
     <script type="text/javascript">
-        $(function() {
+      document.addEventListener("DOMLegacyReady", () => {
+        CrafterCMSNext.system.getStore().subscribe(() => {
```
rart: Let's add a `store$` fn to `craftercms` global that reexports store.ts getStore so we can use that and not use `CrafterCMSNext` in any of these updates. Update all these type of usages. Explicitly, use `window.craftercms`.
```suggestion
      document.addEventListener("DOMLegacyReady", () => {
        window.craftercms.store$().subscribe(() => {
```

Now, that being said, wherever possible, don't even invoke store$() if not necessary. It might be enough for the custom event listener. Check how it worked before, I don't think all these waited for or needed store readyness.

ui/guest/src/models/Rte.ts
```diff
@@ -14,11 +14,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { Settings } from 'tinymce/index';
+import { EditorOptions } from 'tinymce';
 import LookupTable from '@craftercms/studio-ui/models/LookupTable';
 
-export interface TinymceOptions extends Settings {
-  paste_postprocess?(plugin: string, args?: {}): void;
+export interface TinymceOptions extends EditorOptions {
```
rart: These updates seem unrelated to this PR. Pls move.

ui/app/src/main.project-tools.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { unescapeHTML } from './utils/string';
+import { createRoot } from 'react-dom/client';
+import React, { StrictMode } from 'react';
+import CrafterCMSNextBridge from './components/CrafterCMSNextBridge';
+import SiteTools from './pages/SiteTools';
+
+registerComponents();
+publishCrafterGlobal();
+setRequestForgeryToken();
+createCodebaseBridge();
```
rart, line 30: Check that we don't have cases where this is being done twice. For example, on the project tools ftl, there might be an include in the header ftl that ends up importing the `index.legacy.html` that imports all the `main.legacy.tsx` which also does this stuff.

ui/app/src/env/i18n-legacy.ts
```diff
@@ -426,17 +426,6 @@ export const simpleTaxonomyDSMessages = defineMessages({
...
-export const browseCMISMessages = defineMessages({
-  cmis: {
-    id: 'browseCMIS.cmis',
-    defaultMessage: 'CMIS'
-  },
-  cloningCMIS: {
-    id: 'browseCMIS.cloningCMIS',
-    defaultMessage: 'Cloning CMIS Asset'
-  }
-});
```
rart: Move

ui/app/src/env/codebase-bridge.tsx
rart, line 1: What are all these changes for? Ping me to review if simpler.
FYI/Note overall, the use of `any` is longer acceptable. We must pursuit better typings.

ui/app/scripts/package.js
rart, line 1: What's this about? Looks like it needs its own ticket/pr

rart (changes requested):
- [ ] Process dictates we create a separate ticket to track the removed legacy stuff. Please move that to a separate PR. Those screens are broken before this PR so no need to fix them, but we must move that to its own ticket. Other unrelated stuff too pls.
- [ ] Check if we can remove `js-global-context.ftl` (that should be part of this PR).
- [ ] Take note of the updates to carry them over to 4.1.x PR too

static-assets/components/cstudio-forms/forms-engine.js
jvega190: not needed, removing

ui/app/src/env/codebase-bridge.tsx
jvega190: @rart without these changes it does compile, but when opening a form this error shows up `require is not defined` and the form doesn't load.

templates/web/browseS3.ftl
jvega190: Without the store being ready the api calls fail due to lack of token (unauthenticated). Though I'm still not sure how it worked before

ui/app/src/main.project-tools.tsx
jvega190: I didn't add the legacy.html file to head.ftl, so there are not duplicates like the one you mentioned. Check in all imports added for this PR

rart (approved):

PR #3933, opened 2024-05-29
[6705] Update pathMacros regex to handle anything between {} as macros

jvega190:
https://github.com/craftercms/craftercms/issues/6705


jvega190:
@rart just to double check, I updated the regex to get the path macros to handle anything between `{}` as a macro, including curly brackets. So if something like `{{incorrect-macro}}` comes in the path, it'll be completely removed

rart (approved):

PR #3940, opened 2024-05-30
[6033] Enhance the Sidebar navigators

jvega190:
https://github.com/craftercms/craftercms/issues/6033


ui/app/src/components/PathNavigator/PathNavigatorUI.tsx
```diff
@@ -36,6 +36,9 @@ import { ErrorState } from '../ErrorState';
 import { renderErrorState } from '../ErrorState/util';
 import { Pagination } from '../Pagination';
 import Box from '@mui/material/Box';
+import NavigateBeforeIcon from '@mui/icons-material/NavigateBefore';
+import IconButton from '@mui/material/IconButton';
+import Tooltip from '@mui/material/Tooltip';
```
rart: This was added but no other changes were made to this file; seems it is not in use?

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
```diff
@@ -99,6 +101,7 @@ export interface PathNavigatorTreeStateProps {
   keywordByPath: LookupTable<string>;
   totalByPath: LookupTable<number>;
   offsetByPath: LookupTable<number>;
+  currentLimitByPath: LookupTable<number>;
```
rart, line 106: We said the limit was global for all paths. Why a limit lookup table by path? Could we just keep 1 limit? Would paths that have already loaded with a different limit to get correctly loaded with the single limit as it was?

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
```diff
@@ -354,22 +354,26 @@ export function PathNavigatorTreeItem(props: PathNavigatorTreeItemProps) {
...
-                    onClearKeywords();
-                    onFilterButtonClick();
-                  }}
-                >
-                  <SearchRoundedIcon color={showFilter ? 'primary' : 'action'} />
-                </IconButton>
-              </Tooltip>
-            )}
+            {showFilter ||
+              (Boolean(item.childrenCount) && (
```
rart, line 370: Don't think it matters in this particular case, but gotta be careful how prettier groups conditions if you don't group them yourself.
```suggestion
            {(showFilter || Boolean(item.childrenCount) && (
```

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
```diff
@@ -354,22 +354,26 @@ export function PathNavigatorTreeItem(props: PathNavigatorTreeItemProps) {
...
+                <Tooltip title={<FormattedMessage id="words.filter" defaultMessage="Filter" />}>
+                  <IconButton
+                    size="small"
+                    className={classes.iconButton}
+                    sx={{
+                      opacity: over ? '1' : '.2'
```
rart, line 376: Please check light/dark modes

rart (changes requested):

ui/app/src/components/PathNavigatorTree/PathNavigatorTree.tsx
jvega190: Yeah it's done with a global limit for all paths under a Tree. The thing is that when fetching items the offsets gets calculated using the limit. If the limit is set to 5, and theres a path that already has 5 items fetched, if the limit changes and more items are fetched from the path, the new limit will be used for the offset calculation, and a wrong set of items will be retrieved. So the lookup is to 'cache' the limits used for each path and used to correctly calculate the offset (the updated limit is used to retrieve the correct amount of items)

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
jvega190: Prettier still re-formats to the previous one after setting it as in the suggestion

ui/app/src/components/PathNavigatorTree/PathNavigatorTreeItem.tsx
jvega190: Yeah all good, 
image
image

rart (approved):

PR #3944, opened 2024-05-31
[911E] Roles & Permissions Fixes

jvega190:
https://github.com/craftersoftware/craftercms/issues/911

- Fix issue where XB edit mode is blocked for pages that are readonly to the user, but contain items that the usere does have write access to.
- Fix issue where the item selector items to which the user doesn't have edit permissions, display the edit button. Clicking the edit button and trying to save an edit crashes the form leaving it in an unusable.
- Adding a tree navigator with the path of the folder of a page shows an item which clicking it doesn't take the user to preview as expected and clicking to open the item menu shows a skeletonised item menu that never resolves.
- Fix issue where the TreeNavigator UI doesn't handle error in the service response and displays a spinner that doesn't go away.


rart (changes requested):
Conflicts

rart (approved):

PR #3978, opened 2024-06-11
[4.1.x][6920] Opening a form with read-only access to the main item disables editing items to which the user does have write access to

jvega190:
https://github.com/craftercms/craftercms/issues/6920

### DIFF
--- a/static-assets/components/cstudio-forms/controls/node-selector.js
+++ b/static-assets/components/cstudio-forms/controls/node-selector.js
@@ -385,6 +385,7 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField, 
 
           const isComponent = item.key.startsWith('/site') || item.inline;
           const hasWritePermission = !(item.key in itemsByPath) || itemsByPath[item.key].availableActionsMap.edit;
+          // If the user has write permission, the user can edit the item. If the item is an embedded component, check for the readonly property.
           const hasEditAction = hasWritePermission && !(item.inline && this.readonly);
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-forms/controls/node-selector.js
```diff
@@ -396,38 +397,36 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField,
             '<button class="fa fa-trash node-selector-item-icon" title="Delete" aria-label="Delete" role="button"></button>'
           );
           const isEditable = this.allowEdit && (isComponent || craftercms.utils.content.isEditableAsset(item.key));
-          if (isEditable) {
-            if (isComponent || !this.readonly) {
```
rart, line 400: What was the role of the `isComponent` part of the condition? Are you certain it can be safely removed?
If removed, is the const still in use elsewhere or should it be removed altogether?

jvega190: See how `isComponent` is already checked in the `isEditable` validation, and readOnly is already validated in `hasEditAction`. That `isComponent` validation is not valid anymore because basically just because it was a component it was allowing the edit action. Now we consider the permissions

I'll add comments for it to be clearer.
There are 3 variables related with edition:
- allowEdit: the datasource has edit function (`datasource.edit` exists)
- isEditable: the item can be edited by studio (e.g. is a component, or a text file). A non-editable item is for example an image.
- hasEditAction: there's where we check for the edit permission, but if the item is an inline component, then we check for the readonly property.

rart (approved):

PR #3987, opened 2024-06-12
[6921] SingleFileUpload component enhancements

jvega190:
https://github.com/craftercms/craftercms/issues/6921

### DIFF
--- a/ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
+++ b/ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
@@ -35,10 +35,11 @@ import useSiteUIConfig from '../../hooks/useSiteUIConfig';
 import { ensureSingleSlash } from '../../utils/string';
 import { toQueryString } from '../../utils/object';
 import Alert from '@mui/material/Alert';
-import CloseIcon from '@mui/icons-material/Close';
 import IconButton from '@mui/material/IconButton';
 import Tooltip from '@mui/material/Tooltip';
 import { getResponseError } from '../UploadDialog/util';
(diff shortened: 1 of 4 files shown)


ui/app/src/components/UploadDialog/util.ts
```diff
@@ -62,3 +64,21 @@ export interface UploadDialogContainerProps extends UploadDialogProps {
   setPendingChanges?(pending: boolean): void;
   onMinimized?(): void;
 }
+
+export const getResponseError = (responseText: string, formatMessage: IntlShape['formatMessage']) => {
+  try {
+    const parsed = JSON.parse(responseText);
+    if (parsed.response) {
```
jvega190, line 71: @rart let's check this tomorrow, the write-content response can't be parsed and ends up in the catch, so we still don't have a clear error message when it fails to upload

ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
```diff
@@ -236,8 +251,8 @@ export function SingleFileUpload(props: SingleFileUploadProps) {
 
   useEffect(() => {
     const onUploadError = (file, error, response) => {
-      uppy.cancelAll();
```
jvega190, line 239: When in error, we used to cancel all uploads to avoid uppy triggering a 'complete' event (uppy triggers 'complete' even though there are errors). But cancelling the uploads caused the retry upload not to work. So, by removing this line and validating the successful files in the `onCompleteUpload` we avoid sending the completion action when there are errors, and the retry action gets to work correctly.

rart (approved):

PR #3995, opened 2024-06-14
[6954] XB does not show up added repeating group item and need a browser refresh

jvega190:
https://github.com/craftercms/craftercms/issues/6954

### DIFF
--- a/ui/guest/src/contentController.ts
+++ b/ui/guest/src/contentController.ts
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
+  const currentItems = models[modelId][fieldId];
+  const newItems = index ? currentItems.splice(index, 0, instance) : [...(currentItems ?? []), instance];
+  models$.next({
+    ...models,
+    [modelId]: {
+      ...models[modelId],
+      [fieldId]: newItems
+    }
+  });
+
   const action = insertItemOperation({
     modelId,
     fieldId,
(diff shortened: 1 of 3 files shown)


ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
+  const currentItems = models[modelId][fieldId];
+  const newItems = index ? currentItems.splice(index, 0, instance) : [...(currentItems ?? []), instance];
```
rart: do not mutate

rart (changes requested):

ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
+  const currentItems = (models[modelId][fieldId] ?? []).slice();
```
rart:
```suggestion
  const currentItems = [...(models[modelId][fieldId] ?? [])];
```

ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
...
+  const newItems = index ? currentItems.splice(index, 0, instance) : [...currentItems, instance];
+  models$.next({
+    ...models,
+    [modelId]: {
+      ...models[modelId],
+      [fieldId]: newItems
+    }
+  });
```
rart: Create a nested fields/indexes example (e.g. repeat_o.itemSelector_o, 0.1) and see if this works. My guess if we need to accommodate the dotted notation by using the utility functions we have.

See if we have other instances of this issue. We should file a ticket and address.

rart (changes requested):

rart (approved):

PR #3996, opened 2024-06-14
[4.1.x][6954] XB does not show up added repeating group item and need a browser refresh

jvega190:
https://github.com/craftercms/craftercms/issues/6954

### DIFF
--- a/ui/guest/src/contentController.ts
+++ b/ui/guest/src/contentController.ts
@@ -385,7 +385,7 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
-  const currentItems = (models[modelId][fieldId] ?? []).slice();
+  const currentItems = [...(models[modelId][fieldId] ?? [])];
   const newItems = index ? currentItems.splice(index, 0, instance) : [...currentItems, instance];
   models$.next({
     ...models,
--- a/ui/guest/src/contentController.ts
+++ b/ui/guest/src/contentController.ts
@@ -17,6 +17,7 @@
 import { BehaviorSubject, Observable, of, Subject } from 'rxjs';
 import { filter, map, pluck, switchMap, take, tap } from 'rxjs/operators';
 import * as Model from '@craftercms/studio-ui/utils/model';
+import { extractCollectionPiece } from '@craftercms/studio-ui/utils/model';
 import Cookies from 'js-cookie';
 import { fromTopic, post } from './utils/communicator';
 import { v4 as uuid } from 'uuid';
@@ -46,7 +47,6 @@ import { crafterConf } from '@craftercms/classes';
 import { getDefaultValue } from '@craftercms/studio-ui/utils/contentType';
 import { ModelHierarchyDescriptor, ModelHierarchyMap, modelsToLookup } from '@craftercms/studio-ui/utils/content';
 import { SandboxItem, StandardAction } from '@craftercms/studio-ui/models';
-import { mergeArraysAlternatively } from '@craftercms/studio-ui/utils/array';
 
 // if (process.env.NODE_ENV === 'development') {
 // TODO: Notice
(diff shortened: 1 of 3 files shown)


ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
+  const currentItems = models[modelId][fieldId];
+  const newItems = index ? currentItems.splice(index, 0, instance) : [...(currentItems ?? []), instance];
```
rart: Don't modify/mutate the existing array instance.

rart (changes requested):

ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
     }
   });
 
+  const currentItems = (models[modelId][fieldId] ?? []).slice();
```
rart:
```suggestion
  const currentItems = [...(models[modelId][fieldId] ?? [])];
```

rart (changes requested):

ui/guest/src/contentController.ts
```diff
@@ -385,6 +385,16 @@ export function insertItem(modelId: string, fieldId: string, index: number | str
...
 
+  const currentItems = [...(models[modelId][fieldId] ?? [])];
+  const newItems = index ? currentItems.splice(index, 0, instance) : [...currentItems, instance];
+  models$.next({
+    ...models,
+    [modelId]: {
+      ...models[modelId],
+      [fieldId]: newItems
+    }
+  });
```
rart: Create a nested fields/indexes example (e.g. repeat_o.itemSelector_o, 0.1) and see if this works. My guess if we need to accommodate the dotted notation by using the utility functions we have.

rart (changes requested):

ui/guest/src/utils/ice.ts
```diff
@@ -79,7 +79,7 @@ export function getCollectionWithoutItemAtIndex(collection: string[], index: str
   return collection.slice(0, parsedIndex).concat(collection.slice(parsedIndex + 1));
 }
 
-export function getCollection(model: ContentInstance, fieldId: string, index: string | number): string[] {
+export function getCollection(model: ContentInstance, fieldId: string, index: string | number) {
```
jvega190: @rart will ping you to ask a couple of things about this

ui/guest/src/contentController.ts
```diff
@@ -379,12 +380,37 @@ export function duplicateItem(modelId: string, fieldId: string, index: number |
 export function insertItem(modelId: string, fieldId: string, index: number | string, contentType: ContentType): void {
   const instance: InstanceRecord = {};
   const models = getCachedModels();
-  Object.entries(contentType.fields[fieldId].fields).forEach(([id, field]) => {
+  const fields = fieldId.split('.');
+  const field = fields.reduce((acc, field) => acc[field].fields, contentType.fields);
```
rart:
```suggestion
  const fieldPathArray = fieldId.split('.');
  const repeatFields = fields.reduce((acc, field) => acc[field].fields, contentType.fields);
```

ui/guest/src/contentController.ts
```diff
@@ -379,12 +380,37 @@ export function duplicateItem(modelId: string, fieldId: string, index: number |
...
+  Object.entries(field).forEach(([id, field]) => {
     if (!systemProps.includes(field.id)) {
       instance[id] = getDefaultValue(field);
     }
   });
 
+  const indexPieces = index?.toString().split('.');
+  const concatFieldId = !isSimple(fieldId)
+    ? mergeArraysAlternatively(fieldId.split('.'), indexPieces).join('.')
+    : fieldId;
```
rart:
```suggestion
  const fieldPathWithIndexes = isSimple(fieldId)
    ? fieldId
    : mergeArraysAlternatively(fieldId.split('.'), indexPieces).join('.');
```

ui/guest/src/contentController.ts
```diff
@@ -379,12 +380,37 @@ export function duplicateItem(modelId: string, fieldId: string, index: number |
...
+  const currentItems = getCollection(models[modelId], concatFieldId, index) ?? [];
```
rart: Strange to send the field with indexes and an index also. Pls check

ui/guest/src/contentController.ts
```diff
@@ -379,12 +380,37 @@ export function duplicateItem(modelId: string, fieldId: string, index: number |
...
+  if (nou(index) || indexPieces.length < fields.length) {
+    // Index doesn't exist for current field subsection (when not simple field), or index doesn't exist at all, insert at the end
+    newItems = [...currentItems, instance];
+  } else {
+    // If index is complex, we already have the set of items from the model, now we need the latter part of the index
+    // to insert the new instance at the desired position.
+    let targetIndex = isSimple(index) ? index : popPiece(index as string);
+    newItems = currentItems.splice(targetIndex as number, 0, instance);
+  }
+  const model = setCollection(models[modelId], fieldId, index, newItems);
```
rart: Do not mutate the existing instance inside of models.

rart (changes requested):

ui/guest/src/contentController.ts
rart: The additional complexity here is that shouldn't mutate the inner object target of the collection being set

rart (approved):

PR #3998, opened 2024-06-17
[5543] Improve the Add Components panel in XB

jvega190:
https://github.com/craftercms/craftercms/issues/5543

### DIFF
--- a/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
+++ b/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
@@ -85,6 +85,7 @@ export function PreviewBrowseComponentsPanelUI(props: PreviewBrowseComponentsPan
                 <DraggablePanelListItem
                   key={item.craftercms.id}
                   primaryText={item.craftercms.label}
+                  avatarColorBase={item.craftercms.contentTypeId}
                   onDragStart={() => onDragStart(item)}
                   onDragEnd={onDragEnd}
                 />
--- a/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
+++ b/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
@@ -64,48 +64,44 @@ export function PreviewBrowseComponentsPanelUI(props: PreviewBrowseComponentsPan
   } = props;
   const { formatMessage } = useIntl();
   const { classes } = useComponentsPanelUI();
-  return (
+  return awaitingGuestCheckIn ? (
+    <Alert severity="info" variant="outlined" icon={<HourglassEmptyRounded />} sx={{ border: 0 }}>
+      <FormattedMessage defaultMessage="Waiting for the preview application to load." />
+    </Alert>
+  ) : (
     <>
-      {awaitingGuestCheckIn ? (
-        <Alert severity="info" variant="outlined" icon={<HourglassEmptyRounded />} sx={{ border: 0 }}>
-          <FormattedMessage defaultMessage="Waiting for the preview application to load." />
-        </Alert>
-      ) : (
-        <>
-          <Pagination
-            count={count}
-            rowsPerPage={limit}
-            page={pageNumber}
-            onPageChange={(e, page: number) => onPageChanged(page * limit)}
-            onRowsPerPageChange={onRowsPerPageChange}
+      <Pagination
+        count={count}
+        rowsPerPage={limit}
+        page={pageNumber}
+        onPageChange={(e, page: number) => onPageChanged(page * limit)}
+        onRowsPerPageChange={onRowsPerPageChange}
+      />
+      <div className={classes.browsePanelWrapper}>
+        <List>
+          {items.map((item: ContentInstance) => (
+            <DraggablePanelListItem
+              key={item.craftercms.id}
+              primaryText={item.craftercms.label}
+              avatarColorBase={item.craftercms.contentTypeId}
+              onDragStart={() => onDragStart(item)}
+              onDragEnd={onDragEnd}
+            />
+          ))}
+        </List>
+        {count === 0 && (
+          <EmptyState
+            title={formatMessage(translations.noResults)}
+            classes={{ image: classes.noResultsImage, title: classes.noResultsTitle }}
           />
-          <div className={classes.browsePanelWrapper}>
-            <List>
-              {items.map((item: ContentInstance) => (
-                <DraggablePanelListItem
-                  key={item.craftercms.id}
-                  primaryText={item.craftercms.label}
-                  avatarColorBase={item.craftercms.contentTypeId}
-                  onDragStart={() => onDragStart(item)}
-                  onDragEnd={onDragEnd}
-                />
-              ))}
-            </List>
-            {count === 0 && (
-              <EmptyState
-                title={formatMessage(translations.noResults)}
-                classes={{ image: classes.noResultsImage, title: classes.noResultsTitle }}
-              />
-            )}
-            <FormHelperText className={classes.helperTextWrapper}>
-              <FormattedMessage
-                id="previewBrowseComponentsPanel.sharedComponentsHelperText"
-                defaultMessage="Only shared components are shown here"
-              />
-            </FormHelperText>
-          </div>
-        </>
-      )}
+        )}
+        <FormHelperText className={classes.helperTextWrapper}>
+          <FormattedMessage
+            id="previewBrowseComponentsPanel.sharedComponentsHelperText"
+            defaultMessage="Only shared components are shown here"
+          />
+        </FormHelperText>
+      </div>
     </>
   );
 }
--- a/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
+++ b/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
@@ -24,6 +24,8 @@ import EmptyState from '../EmptyState/EmptyState';
 import { useComponentsPanelUI } from './styles';
 import FormHelperText from '@mui/material/FormHelperText';
 import Pagination from '../Pagination';
+import HourglassEmptyRounded from '@mui/icons-material/HourglassEmptyRounded';
+import Alert from '@mui/material/Alert';
 
 export interface ComponentResource {
   count: number;
@@ -34,6 +36,7 @@ export interface ComponentResource {
 }
 
 export interface PreviewBrowseComponentsPanelUIProps {
+  awaitingGuestCheckIn: boolean;
   items: Array<ContentInstance>;
   count: number;
   pageNumber: number;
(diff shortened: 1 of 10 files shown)


ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanelUI.tsx
```diff
@@ -48,42 +51,61 @@ export interface PreviewBrowseComponentsPanelUIProps {
...
-              onDragEnd={onDragEnd}
-            />
-          ))}
-        </List>
-        {count === 0 && (
-          <EmptyState
-            title={formatMessage(translations.noResults)}
-            classes={{ image: classes.noResultsImage, title: classes.noResultsTitle }}
+      {awaitingGuestCheckIn ? (
+        <Alert severity="info" variant="outlined" icon={<HourglassEmptyRounded />} sx={{ border: 0 }}>
```
rart: Unnecessary fragment
```ts
return (
    awaitingGuestCheckIn ? (
      <Alert severity="info" variant="outlined" icon={<HourglassEmptyRounded />} sx={{ border: 0 }}>
```

ui/app/src/state/epics/contentTypes.ts
```diff
@@ -65,21 +65,21 @@ export default [
...
-            ? Object.values(state.contentTypes.byId)
-                .filter(
-                  (contentType: ContentType) =>
-                    contentType.type === 'component' && !contentType.id.includes('/level-descriptor')
-                )
-                .map((contentType) => contentType.id)
+          state.preview.components.contentTypeFilter === 'compatible'
+            ? allowedContentTypes?.length
+              ? allowedContentTypes
+              : [''] // Empty array is like sending no filters, so it returns items unfiltered.
```
rart: This is not an empty array so either the comment or the object needs to be fixed.

ui/guest/src/iceRegistry.ts
```diff
@@ -153,8 +153,10 @@ export function register(registration: ICERecordRegistration): number {
       }
       allowedContentTypesData.recordIdKeyLookup[record.id] = key;
       allowedContentTypesData.allowedLookup[key].recordIds.push(record.id);
-      collectAndEmitAllowedContentTypes();
     }
+    // If there are no allowed content types, we still need to call this for the redux state to be updated and not be
+    // null. This way the panels don't get stuck in a loading state ('Waiting for the preview application to load').
+    collectAndEmitAllowedContentTypes();
```
```suggestion
    // Invoking all the time so that the `allowedContentTypes$` subject emits either way.
    // Not emitting could cause UIs to be stuck in a loading state.
    collectAndEmitAllowedContentTypes();
```

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -139,52 +147,61 @@ export function PreviewSearchPanel() {
...
-                items: response.result.items,
-                count: response.result.total,
-                limit: options?.limit ?? initialSearchParameters.limit
+      if (allowedTypesData) {
+        // If there are no allowed types, an empty array causes the request to return everything.
+        const allowedTypes = Object.keys(allowedTypesData).length
+          ? Object.entries(allowedTypesData)
+              .filter(([, type]) => type.shared)
+              .map(([key]) => key)
+          : [''];
```
rart: Not very efficient or performant to make the engine parse the keys and also parse the entries
```suggestion
        // If there are no allowed types, an empty array causes the request to return everything.
        let allowedTypes = Object.entries(allowedTypesData).filter(([, type]) => type.shared)
              .map(([key]) => key);
       if (allowedTypes.length === 0) allowedTypes = ['']
```

rart: If there are no allowed types, don't we want to get assets only, instead of getting everything? Perhaps the comment needs fixing?

rart (changes requested):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
jvega190: Updated the comment to be more detailed. Also, `allowedTypes` is only for the `content-type` filter, `mime-type` stays the same so assets will be retrieved.

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -139,52 +147,61 @@ export function PreviewSearchPanel() {
...
-                isFetching: false,
-                items: response.result.items,
-                count: response.result.total,
-                limit: options?.limit ?? initialSearchParameters.limit
+      if (allowedTypesData) {
+        // If there are no allowed types, an empty array causes the request to return everything. So we send an array
+        // with an empty string to filter out all content-types.
+        let allowedTypes = Object.entries(allowedTypesData)
+          .filter(([, type]) => type.shared)
+          .map(([key]) => key);
```
rart: Could be done in a single loop with `flatMap`

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -139,52 +147,61 @@ export function PreviewSearchPanel() {
...
-                count: response.result.total,
-                limit: options?.limit ?? initialSearchParameters.limit
-              });
-            } else {
-              setState({
-                isFetching: false,
-                items: response.result.items,
-                count: response.result.total,
-                limit: options?.limit ?? initialSearchParameters.limit
+      if (allowedTypesData) {
```
rart: This puts the state in fetching and never clears it. Overall the approach doesn't make too much sense. Perhaps just search without types like when allowedTypes.length == 0. We should ensure cancellation as there's a chance of overlapping requests.

ui/app/src/state/epics/contentTypes.ts
```diff
@@ -65,21 +65,23 @@ export default [
     action$.pipe(
       ofType(fetchComponentsByContentType.type, setContentTypeFilter.type),
       withLatestFrom(state$),
-      switchMap(([, state]) =>
-        fetchItemsByContentType(
+      switchMap(([, state]) => {
+        const allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes)
+          .filter(([, value]) => value.shared)
+          .map(([key]) => key);
```
rart: - flatMap
- Object.entries will throw if allowedContentTypes is nullish

ui/app/src/state/epics/contentTypes.ts
```diff
@@ -65,21 +65,23 @@ export default [
...
-                  (contentType: ContentType) =>
-                    contentType.type === 'component' && !contentType.id.includes('/level-descriptor')
-                )
-                .map((contentType) => contentType.id)
+          state.preview.components.contentTypeFilter === 'compatible'
+            ? allowedContentTypes?.length
+              ? allowedContentTypes
+              : // When there are no allowed content types, sending an empty array is like sending no content-type filter,
+                // so it returns items unfiltered. By sending an array with an empty string, we filter out all content-types.
+                ['']
```
rart: Maybe rather than this _hack_, the `fetchItemsByContentType` should not send the `content-type` filter if it's null or empty. This would allow having these checks removed all-across. Or ultimately, move this to the service if it makes sense; it is a bit of a yellow flag that this is in 2 places in the code.

Also, is this call exclusively to get content items by their content type? If all content types are filtered out, what is this returning? If this thing's job is to search by content type and we know in advanced the result is empty, then this shouldn't even be done and/or the service should have that smarts.

rart (changes requested):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
rart: You still have the if statement?

rart (changes requested):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
jvega190: that's just to wait for the allowedTypesData to be loaded in the state

rart: Waiting is not necessary, you already adjusted to account for null. Going back to the original comment, this puts the state in `fetching` but never clears it. If waiting is desired then this shouldn't be called in the first place.

rart (changes requested):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
jvega190: Actually the fetching state is set back to false it both the next and error sections of the subscription
- success: https://github.com/craftercms/studio-ui/pull/3998/files?show-viewed-files=true&file-filters%5B%5D=#diff-0aa4d1c7d6765372b4344f4c540a5ff21fba5f157d2ec625517c46585a00fe65R182
https://github.com/craftercms/studio-ui/pull/3998/files?show-viewed-files=true&file-filters%5B%5D=#diff-0aa4d1c7d6765372b4344f4c540a5ff21fba5f157d2ec625517c46585a00fe65R190

- error:  https://github.com/craftercms/studio-ui/pull/3998/files?show-viewed-files=true&file-filters%5B%5D=#diff-0aa4d1c7d6765372b4344f4c540a5ff21fba5f157d2ec625517c46585a00fe65R198

jvega190: Removed the if condition as it is not needed anymore

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -137,13 +145,17 @@ export function PreviewSearchPanel() {
 
   const onSearch = useCallback(
     (keywords: string = '', options?: ComponentsContentTypeParams) => {
+      let sub;
```
rart: Missing type; but moreover, I don't see a point in declaring this here above instead of doing it inline with the call that would automatically give it its type.

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -137,13 +145,17 @@ export function PreviewSearchPanel() {
...
       setState({ isFetching: true });
       setError(null);
-      search(site, {
+      const allowedTypes = Object.entries(allowedTypesData ?? {}).flatMap(([key, type]) => (type.shared ? [key] : []));
+
+      sub?.unsubscribe();
```
rart: This will always be undefined here. Again, don't see why not declare it inline in the next line.

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -180,11 +192,16 @@ export function PreviewSearchPanel() {
             }
           },
           error: ({ response }) => {
+            setState({ isFetching: false });
             setError(response.response);
           }
         });
+
+      return () => {
+        sub.unsubscribe();
```
rart: This needs to be invoked by the consumers, this is not an effect.

rart (changes requested):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
```diff
@@ -124,13 +130,17 @@ export function PreviewSearchPanel() {
...
+  const awaitingGuestCheckIn = nou(allowedTypesData);
   const contentTypes = useContentTypeList(
     (contentType) => contentType.id !== '/component/level-descriptor' && contentType.type === 'component'
   );
   const contentTypesLookup = useMemo(
     () => (contentTypes ? createLookupTable(contentTypes, 'id') : null),
     [contentTypes]
   );
+  const onPageChangedSubscription = useRef<Subscription>();
+  const onRowsPerPageChangeSubscription = useRef<Subscription>();
```
rart: Why 2 subscriptions? Don't see why there would ever be two active. Also, the unsubscribe needs to be called before recreating the new subscription.

**Check if calling unsubscribe on completed subscriptions would throw**. If so, you need to clear the ref after unsubscribing.

You may even use the same ref for the useEffect.

rart (changes requested):

rart (approved):

ui/app/src/components/PreviewSearchPanel/PreviewSearchPanel.tsx
jvega190: After cheking, calling unsubscribe on completed subscriptions doesn't throw an error), it doesn't do anything.

PR #4013, opened 2024-06-25
[6988] Path selection not working with 31 folders

jvega190:
https://github.com/craftercms/craftercms/issues/6988

### DIFF
--- a/ui/app/src/services/content.ts
+++ b/ui/app/src/services/content.ts
@@ -1257,9 +1257,10 @@ export function fetchChildrenByPaths(
         map(({ response: { items } }) => {
           const data = {};
           items.forEach(({ children, levelDescriptor, total, offset, limit, path }) => {
+            const totalWithDescriptor = levelDescriptor ? total + 1 : total;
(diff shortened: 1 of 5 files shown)


jvega190:
There's a backend update needed for this to work. Ticket will be linked to this PR and its ticket.

jvega190:
https://github.com/craftercms/craftercms/issues/6992

ui/app/src/services/content.ts
```diff
@@ -1257,9 +1257,10 @@ export function fetchChildrenByPaths(
         map(({ response: { items } }) => {
           const data = {};
           items.forEach(({ children, levelDescriptor, total, offset, limit, path }) => {
+            const totalWithDescriptor = levelDescriptor ? total + 1 : total;
```
jvega190, line 1260: @rart the children API had a bug where the children count was considering the level-descriptor, now it has been fixed. But in the UI we include the descriptor (if exists) as part of the children (and considering it into the total). Since the API changed, we need to include the descriptor into the total for the UI to continue working properly (in PathNavs)

rart (approved):

PR #4017, opened 2024-06-26
[6991] XB - Inserting items in nested repeating groups ends up with incorrect content

jvega190:
https://github.com/craftercms/craftercms/issues/6991

### DIFF
--- a/ui/app/src/services/content.ts
+++ b/ui/app/src/services/content.ts
@@ -464,12 +464,12 @@ export function insertItem(
     site,
     path,
     (element) => {
-      let node = extractNode(element, removeLastPiece(fieldId) || fieldId, index);
+      const node = extractNode(element, fieldId, index);
       const newItem = createElement('item');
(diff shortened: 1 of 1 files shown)


ui/app/src/services/content.ts
```diff
@@ -464,12 +464,12 @@ export function insertItem(
...
     path,
     (element) => {
-      let node = extractNode(element, removeLastPiece(fieldId) || fieldId, index);
+      const node = extractNode(element, fieldId, index);
       const newItem = createElement('item');
       const serializedInstance = {};
       for (let key in instance) {
         if (key !== 'craftercms') {
-          let value = instance[key];
+          const value = instance[key];
```
rart: We need to do these updates their own ticket
```suggestion
          let value = instance[key];
```

rart (changes requested):

rart (approved):

PR #4018, opened 2024-06-27
[6983] Improve XB ease-of-use and resilience for JS apps programmatically registering elements for XB

jvega190:
https://github.com/craftercms/craftercms/issues/6983

### DIFF
--- a/ui/guest/src/elementRegistry.ts
+++ b/ui/guest/src/elementRegistry.ts
@@ -51,7 +51,7 @@ let registry: LookupTable<number[]> = {};
 // Lookup table of element record id, index by the element
 const recordIdByElementLookup = new Map<Element, number>();
 // Stream of ids being deregistered, used to cancel pending model fetch operations.
-const deregister$ = new Subject<number | string>();
+const deregister$ = new Subject<string>();
 
 export function get(id: number): ElementRecord {
   const record = db[id];
@@ -141,7 +141,7 @@ export function register(payload: ElementRecordRegistration): number {
       : Array.isArray(fieldId)
         ? fieldId
         : fieldId.split(',').map((str) => str.trim());
-  const terminator$ = deregister$.pipe(filter((_id) => _id === id));
+  const terminator$ = deregister$.pipe(filter((_id) => _id === String(id)));
 
   function create() {
     // Create/register the physical record
@@ -226,7 +226,7 @@ export function completeDeferredRegistration(id: number): void {
 
 export function deregister(id: string | number): ElementRecord {
   const record = db[id];
-  deregister$.next(id);
+  deregister$.next(String(id));
   if (notNullOrUndefined(record)) {
     const { iceIds, element } = record;
     recordIdByElementLookup.delete(element);
--- a/ui/guest/src/elementRegistry.ts
+++ b/ui/guest/src/elementRegistry.ts
@@ -181,6 +181,7 @@ export function register(payload: ElementRecordRegistration): number {
 
   // If the relevant model is loaded, complete its registration, otherwise,
   // request it and complete registration when it does load.
+  recordIdByElementLookup.set(element, id);
   if (hasCachedModel(modelId)) {
     completeRegistration(id);
   } else {
--- a/ui/guest/src/elementRegistry.ts
+++ b/ui/guest/src/elementRegistry.ts
@@ -49,9 +49,8 @@ let db: LookupTable<ElementRecord> = {};
 let registry: LookupTable<number[]> = {};
 // Lookup table of element record id, index by the element
 const recordIdByElementLookup = new Map<Element, number>();
-
-let registrationSubscription;
-let deferredRegistrationSubscription;
+// Subscription registry for deferred registrations
+let registrationSubscriptions = {};
 
 export function get(id: number): ElementRecord {
   const record = db[id];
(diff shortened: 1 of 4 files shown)


rart (changes requested):
Is this finished? I don't see for example the recordIdByElementLookup being cleaned when deregister is called?

Also not sure I'm seeing work to make sure deregister would would even when the deferred request hasn't completed.

ui/guest/src/store/subjects.ts
```diff
@@ -29,6 +29,8 @@ export const escape$ =
     ? new Subject<KeyboardEvent>()
     : fromEvent<KeyboardEvent>(document, 'keydown').pipe(filter((e) => e.key === 'Escape'));
 
+export const guestCheckIn$ = new ReplaySubject<boolean>(1);
```
jvega190, line 32: There's one detail here (in the usage). We still need to listen for the `xb:loaded` so were sure we have `craftercms.xb`, then subscribe to `guestCheckIn$`

ui/guest/src/elementRegistry.ts
```diff
@@ -154,7 +163,7 @@ export function register(payload: ElementRecordRegistration): number {
     // for the model to be loaded.
     if (isInheritedField(model.craftercms.id, fieldId)) {
       byPathFetchIfNotLoaded(model.craftercms.sourceMap?.[fieldId]).subscribe((response) => {
-        model$(response.craftercms.id)
+        deferredRegistrationSubscription = model$(response.craftercms.id)
```
rart: There are many calls to register done sequentially. A single subscription doesn't work. You need maybe a stream of deregistered ids to stop this one.

ui/guest/src/elementRegistry.ts
```diff
@@ -47,6 +47,11 @@ let seq = 0;
 let db: LookupTable<ElementRecord> = {};
 // Lookup table of element record id arrays, indexed by iceId
 let registry: LookupTable<number[]> = {};
+// Lookup table of element record id, index by the element
+const recordIdByElementLookup = new Map<Element, number>();
+
+let registrationSubscription;
+let deferredRegistrationSubscription;
```
rart: What's the need for 2 subjects?

rart (changes requested):
Let's add/move the `registerElements` and `deregisterElements` functions to the library (to the index)

ui/guest/src/elementRegistry.ts
jvega190: There are two places where we have subscriptions, in the main register funcion (if main model is not fetched) and in the `completeRegistration` function (if is an inherited field, to load the model). What I did was to add a lookup for registrations (with the id as the key), and the content is an array of the subscriptions, considering that there may be up to 2 subscriptions. Then, when deregistering, delete all of the subscriptions for the id

ui/guest/src/index.tsx
```diff
@@ -101,6 +103,36 @@ export function initExperienceBuilder(props: ExperienceBuilderProps) {
...
+        label: element.getAttribute('data-craftercms-label')
+      });
+    });
+  }
+}
+
+export function deregisterElements(container: HTMLElement) {
+  if (container) {
+    const articleElements = container.querySelectorAll('[data-craftercms-model-id]');
+    articleElements.forEach((element) => {
```
rart: These aren't articles here...
```suggestion
    const elements = container.querySelectorAll('[data-craftercms-model-id]');
    elements.forEach((element) => {
```

ui/guest/src/elementRegistry.ts
```diff
@@ -47,6 +47,10 @@ let seq = 0;
 let db: LookupTable<ElementRecord> = {};
 // Lookup table of element record id arrays, indexed by iceId
 let registry: LookupTable<number[]> = {};
+// Lookup table of element record id, index by the element
+const recordIdByElementLookup = new Map<Element, number>();
+// Subscription registry for deferred registrations
+let registrationSubscriptions = {};
```
rart: Pls apply discussed changes

rart (changes requested):

ui/guest/src/elementRegistry.ts
```diff
@@ -41,12 +41,17 @@ import { notNullOrUndefined, nou, nullOrUndefined } from '@craftercms/studio-ui/
...
 
 let seq = 0;
 // Element record registry
 let db: LookupTable<ElementRecord> = {};
 // Lookup table of element record id arrays, indexed by iceId
 let registry: LookupTable<number[]> = {};
+// Lookup table of element record id, index by the element
+const recordIdByElementLookup = new Map<Element, number>();
+// Stream of ids being deregistered, used to cancel pending model fetch operations.
+const deregister$ = new Subject<number | string>();
```
rart:
```suggestion
const deregister$ = new Subject<string>();
```

ui/guest/src/elementRegistry.ts
```diff
@@ -132,6 +141,7 @@ export function register(payload: ElementRecordRegistration): number {
       : Array.isArray(fieldId)
         ? fieldId
         : fieldId.split(',').map((str) => str.trim());
+  const terminator$ = deregister$.pipe(filter((_id) => _id === id));
```
rart:
```suggestion
  const terminator$ = deregister$.pipe(filter((_id) => _id === String(id)));
```

rart (changes requested):
Pls remember to port changes to 4.1 PR too

rart (approved):

PR #4028, opened 2024-07-02
[7000] Setting an incorrect value in time field controls shows a notification that blocks the form

jvega190:
https://github.com/craftercms/craftercms/issues/7000


jvega190:
The problem was that 2 events were being triggered at the same time (`change` and `blur`). Causing 2 dialogs two open and misbehave. Change event is enough to handle the input events needed for this control

rart (approved):

PR #4031, opened 2024-07-03
[7003] Content diff current version fallback error

jvega190:
https://github.com/craftercms/craftercms/issues/7003

### DIFF
--- a/templates/web/diff.ftl
+++ b/templates/web/diff.ftl
@@ -108,27 +108,30 @@
   </div>
 </#if>
 
+<#include "/static-assets/app/pages/legacy.html">
 <script>
-  $(document).ready(function(){
-    CStudioAuthoring.Service.getCurrentVersion(CStudioAuthoringContext.site, "${dir}", {
-      success: function(version) {
-        $('#current-version').append(version)
-      }
-    });
+  document.addEventListener("CrafterCMS.CodebaseBridgeReady", () => {
+    jQuery(function() {
+      CStudioAuthoring.Service.getCurrentVersion(CStudioAuthoringContext.site, "${dir}", {
+        success: function(version) {
+          $('#current-version').append(version)
+        }
+      });
 
-    $('#cancelBtn').on('click', function() {
-      parent.$('body').trigger('diff-end');
-    });
-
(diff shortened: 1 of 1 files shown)


templates/web/diff.ftl
```diff
@@ -108,27 +108,30 @@
...
+<#include "/static-assets/app/pages/legacy.html">
 <script>
-  $(document).ready(function(){
-    CStudioAuthoring.Service.getCurrentVersion(CStudioAuthoringContext.site, "${dir}", {
-      success: function(version) {
-        $('#current-version').append(version)
-      }
-    });
+  document.addEventListener("CrafterCMS.CodebaseBridgeReady", () => {
+    jQuery(function() {
```
rart: Probably not necessary to add the jquery ready function

rart (changes requested):

templates/web/diff.ftl
jvega190: Yeah it wasn't necessary, though it seems like it was delaying the script and causing the store to be ready, so I removed the jquery ready and now I subscribed to the store for it to be ready when calling the request (for it to have the needed tokens, etc)

templates/web/diff.ftl
```diff
@@ -108,12 +108,15 @@
...
 
+<#include "/static-assets/app/pages/legacy.html">
 <script>
-  $(document).ready(function(){
-    CStudioAuthoring.Service.getCurrentVersion(CStudioAuthoringContext.site, "${dir}", {
-      success: function(version) {
-        $('#current-version').append(version)
-      }
+  document.addEventListener("CrafterCMS.CodebaseBridgeReady", () => {
+    window.craftercms.store$().subscribe(() => {
```
rart: For better race condition avoidance...
```ts
function init() {
  window.craftercms.store$().subscribe(() => {
    ...
  });
}
if (window.craftercms?.store$)
  init()
else 
  document.addEventListener("CrafterCMS.CodebaseBridgeReady", init)
```

rart (changes requested):

rart (approved):
@jvega190 why draft?

PR #4043, opened 2024-07-11
[7009] Timing issue with Guest check in when switching projects 

jvega190:
https://github.com/craftercms/craftercms/issues/7009

### DIFF
--- a/ui/app/src/components/Preview/Preview.tsx
+++ b/ui/app/src/components/Preview/Preview.tsx
@@ -139,7 +139,7 @@ function Preview() {
       // Not the first render. Something changed and we're updating.
 
       // Check if either the QS or the state has changed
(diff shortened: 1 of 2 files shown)


ui/app/src/components/Preview/Preview.tsx
```diff
@@ -139,7 +139,7 @@ function Preview() {
       // Not the first render. Something changed and we're updating.
 
       // Check if either the QS or the state has changed
-      const qsSiteChanged = qs.site !== prev.qsSite || qs.site !== site;
+      const qsSiteChanged = qs.site !== prev.qsSite && qs.site !== site;
```
rart, line 142: This goes against what the comment is saying. Need you to walk through the logic of why this is the correct behaviour and how it doesn't break something else.

rart (approved):

PR #4048, opened 2024-07-15
[4.1.x][7017] Search feature issues on PathNavTree

jvega190:
https://github.com/craftercms/craftercms/issues/7017

### DIFF
--- a/ui/app/src/state/reducers/pathNavigatorTree.ts
+++ b/ui/app/src/state/reducers/pathNavigatorTree.ts
@@ -77,17 +77,11 @@ const restoreTree = (state, payload) => {
   const childrenByParentPath = chunk.childrenByParentPath;
   const totalByPath = chunk.totalByPath;
   const offsetByPath = chunk.offsetByPath;
-  // Set totalByPath of items that are not in the children object (items not expanded) so we can display the
-  // expand/collapse icon (if they have children). We filter out the items not in the children object to prevent displaying
-  // an incorrect total count (since the children object has the total considering filters).
-  items
-    .filter((item) => !children[item.path])
-    .forEach((item) => {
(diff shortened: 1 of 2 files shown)


ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -77,10 +77,17 @@ const restoreTree = (state, payload) => {
...
-    totalByPath[item.path] = item.childrenCount;
-  });
+  // Set totalByPath of items that are not in the children object (items not expanded) so we can display the
+  // expand/collapse icon (if they have children). We filter out the items not in the children object to prevent displaying
+  // an incorrect total count (since the children object has the total considering filters).
+  items
+    .filter((item) => !children[item.path])
+    .forEach((item) => {
+      totalByPath[item.path] = item.childrenCount;
+    });
```
rart: Get rid of the filter. Do conditional on the forEach to avoid the additional iteration.

ui/app/src/state/reducers/pathNavigatorTree.ts
```diff
@@ -77,10 +77,17 @@ const restoreTree = (state, payload) => {
...
   Object.keys(children).forEach((parentPath) => {
+    // Initialize the childrenByParentPath with an empty array.
+    childrenByParentPath[parentPath] = [];
```
rart: Why? Explain. The comment currently says what's evident by simply reading the code; it needs the rationale.
```suggestion
    // Initialize the childrenByParentPath with an empty array so that blah blah...
    childrenByParentPath[parentPath] = [];
```

rart (changes requested):

rart (approved):

PR #4058, opened 2024-07-17
[7032] Uploading an image or file with invalid filename using Image Uploaded to S3 Repository or File Uploaded to S3 Repository datasources a window will show up that can't be closed or cancelled

jvega190:
https://github.com/craftercms/craftercms/issues/7032

### DIFF
--- a/static-assets/components/cstudio-dialogs/uploadS3-dialog.js
+++ b/static-assets/components/cstudio-dialogs/uploadS3-dialog.js
@@ -192,33 +192,9 @@ CStudioAuthoring.Dialogs.UploadS3Dialog = CStudioAuthoring.Dialogs.UploadS3Dialo
         me.callback.success(uploaded);
         CStudioAuthoring.Dialogs.UploadS3Dialog.closeDialog();
       },
-      onError: function ({ response }) {
-        const res = JSON.parse(response.body.response).response,
-          errorMsg = `${res.message}. ${res.remedialAction}`;
-
+      onError: function () {
         me.uploadingFile = false;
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-dialogs/uploadS3-dialog.js
```diff
@@ -192,33 +192,9 @@ CStudioAuthoring.Dialogs.UploadS3Dialog = CStudioAuthoring.Dialogs.UploadS3Dialo
...
       },
-      onError: function ({ response }) {
-        const res = JSON.parse(response.body.response).response,
-          errorMsg = `${res.message}. ${res.remedialAction}`;
-
+      onError: function () {
         me.uploadingFile = false;
         $('#uploadCancelButton').attr('disabled', false);
-
-        CStudioAuthoring.Operations.showSimpleDialog(
```
jvega190, line 202: Since the SingleFileUpload component already does the error handling (shows an alert w/retry icon) this error dialog is no longer needed. The error was that the response was different so the parsing was failing and the state wasn't restored.

static-assets/themes/cstudioTheme/css/contextNav.css
```diff
@@ -1103,3 +1103,19 @@ body.yui-skin-cstudioTheme .mask {
 #crop-popup-inner .error {
   border-color: rgba(169, 68, 66, 0.47);
 }
+
+/* Upload alert override */
```
jvega190, line 1107: Need to discuss

rart (approved):

PR #4070, opened 2024-07-24
[7084] XB styles are affected by Guest app styles

jvega190:
https://github.com/craftercms/craftercms/issues/7084


ui/guest/src/react/GuestGlobalStyles.tsx
```diff
@@ -62,184 +63,187 @@ export function GuestGlobalStyles(props: GuestGlobalStylesProps) {
...
-            }
-          },
-          styles ?? {}
-        )}
-      />
+            },
+            styles ?? {}
+          )}
+        />
+        <Global styles={staticGlobalStyles} />
```
rart: These styles have more than just the typography. Let's split typography vs other things on the studio Side, so you can import `typograhyStyles` by itself. Also, on the Studio side, where ever those things are use, you'd now need to create 2 `<Global />` blocks to add both.

ui/guest/src/react/UltraStyledTooltip.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { Tooltip, TooltipProps } from '@mui/material';
+
+export const UltraStyledTooltip = ({ title, children, ...props }: TooltipProps) => {
+  return (
+    <Tooltip slotProps={{ tooltip: { sx: { fontSize: 12 } } }} {...props} title={title}>
+      {children}
+    </Tooltip>
+  );
+};
```
rart: Why not done through the styled API?

ui/guest/src/react/UltraStyledTypography.tsx
```diff
@@ -0,0 +1,29 @@
...
+// Since this typography runs guest-side, there may be all sorts of styles that affect Typography styles.
+// The idea of this component is to be as specific as possible to avoid guest site styles to break ours.
+
+import { styled } from '@mui/material/styles';
+import { Typography } from '@mui/material';
+
+export const UltraStyledTypography = styled(Typography)(({ theme }) => ({
+  color: theme.palette.text.primary,
+  letterSpacing: 'unset',
+  fontSize: 16
```
rart: Isn't the font size by default 14 px? Actually, we can access that from the theme object? (i.e. `theme.typography.fontSize`)

rart (changes requested):

ui/guest/src/react/UltraStyledTooltip.tsx
jvega190: For some reason the styles for the popper are not applying through sx styles. I tried on an example provided from mui.com and doesn't work. https://codesandbox.io/s/jovial-wescoff-23jhhj?file=/src/Demo.tsx
Tooltip classes doc: https://mui.com/material-ui/api/tooltip/#classes.

jvega190: Using slotProps to set the styles for the tooltip works

rart:
```suggestion
export const UltraStyledTooltip = (props: TooltipProps) => {
  return (
    <Tooltip {...props} slotProps={{ ...props?.slotProps, tooltip: { ...props.slotProps?.tooltip, sx: { ...props.slotProps?.tooltip?.sx, fontSize: 12 } } }} />
  );
};
```

rart (changes requested):

rart (approved):

PR #4073, opened 2024-07-26
[7069] Pressing the ? key in the preview no longer shows the dialog of keyboard shortcuts

jvega190:
https://github.com/craftercms/craftercms/issues/7069


jvega190:
@rart tested in chrome for ubuntu, macos and windows (in both XB and studio side). Btw I don't like how the lib we use is not clear at all in their docs for this particular hotkey

rart (approved):

PR #4074, opened 2024-07-29
[7063] Rail toolbars are hard to grab and drag in Chrome

jvega190:
https://github.com/craftercms/craftercms/issues/7063

### DIFF
--- a/ui/app/src/components/ICEToolsPanel/ICEToolsPanel.tsx
+++ b/ui/app/src/components/ICEToolsPanel/ICEToolsPanel.tsx
@@ -45,6 +45,7 @@ export function ICEToolsPanel() {
   const isOpen = editMode;
   const isLockedForMe = Boolean(item && isItemLockedForMe(item, username));
   const isStale = guest?.mainModelModifier ? guest.mainModelModifier.username !== username : false;
+  const previewIframe = document.getElementById('crafterCMSPreviewIframe');
 
   const onWidthChange = (width) => {
     dispatch(updateIcePanelWidth({ width }));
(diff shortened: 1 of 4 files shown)


ui/app/src/components/ICEToolsPanel/ICEToolsPanel.tsx
```diff
@@ -66,6 +67,16 @@ export function ICEToolsPanel() {
       width={width}
       maxWidth={windowSize}
       onWidthChange={onWidthChange}
+      onResizeStart={() => {
```
jvega190: @rart the issue is that in some browsers the preview iframe causes the mouse events not to trigger (while dragging the drawer). One one side I think this can be a preview util (to set/unset pointer events), but on the other side feels a bit weird to have a util to set pointer events. What do you think?
\* I tested this on chrome for macOS and the behavior is still fine

rart: Yes, mostly because the code is duplicated it needs refactoring.

rart (changes requested):
I think this is only part of the fix. It seemed like part of the issue is that the borders are harder to grab; I think it might be due to the shadows and positioning of the preview frame and possibly adding z-index and or positioning styles to the borders could help. Test it out comparing 4.1 with 4.2 pls.

ui/app/src/components/ResizeableDrawer/ResizeableDrawer.tsx
```diff
@@ -95,13 +97,13 @@ const useStyles = makeStyles<ResizeableDrawerStyles, ResizeableDrawerClassKey>()
       ...drawerPaperRight
     },
     resizeHandle: {
-      width: '1px',
+      width: '4px',
```
jvega190: 4px is the width of the handle when active, so by setting this at the same width we make it easier to grab with no visual difference (since this is not visible when inactive).

ui/app/src/components/ResizeableDrawer/ResizeableDrawer.tsx
```diff
@@ -95,7 +97,7 @@ const useStyles = makeStyles<ResizeableDrawerStyles, ResizeableDrawerClassKey>()
...
```
rart:
```suggestion
      width: '2px',
```

ui/app/src/components/ToolsPanel/ToolsPanel.tsx
```diff
@@ -68,6 +69,7 @@ export function ToolsPanel() {
   const uiConfig = useSiteUIConfig();
   const baseUrl = useSelection<string>((state) => state.env.authoringBase);
   const { username } = useActiveUser();
+  const previewIframe = document.getElementById('crafterCMSPreviewIframe');
```
rart: x

rart (changes requested):

rart (approved):

PR #4076, opened 2024-07-30
[7071] Publishing Dialog thows includes error related to permitted roles after certain actions

jvega190:
https://github.com/craftercms/craftercms/issues/7071

### DIFF
--- a/ui/app/src/components/PublishingStatusDialog/PublishingStatusDialogContainer.tsx
+++ b/ui/app/src/components/PublishingStatusDialog/PublishingStatusDialogContainer.tsx
@@ -60,7 +60,7 @@ export function PublishingStatusDialogContainer(props: PublishingStatusDialogCon
   const user = useActiveUser();
   const siteId = useActiveSiteId();
   const userRoles = user?.rolesBySite[siteId];
-  const allowedUser = permittedRoles?.some((role) => userRoles?.includes(role)) ?? false;
+  const allowedUser = (userRoles && permittedRoles?.some((role) => userRoles.includes(role))) ?? false;
 
   const handleClose = () => {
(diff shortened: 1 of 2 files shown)


ui/app/src/state/epics/users.ts
```diff
@@ -41,7 +41,7 @@ export default [
   // region changeSite
   (action$) =>
     action$.pipe(
-      ofType(changeSite.type),
+      ofType(changeSiteComplete.type),
```
jvega190, line 44: @rart the issue was happening because the user roles were not being fetched properly after changing a site (at `changeSite` the site hasn't been updated yet. This is because of the preview switch API addition (now we have a `changeSiteComplete`, so at that moment the site is updated in the state, and the roles are fetched for the new site.

ui/app/src/components/PublishingStatusDialog/PublishingStatusDialogContainer.tsx
```diff
@@ -60,7 +60,7 @@ export function PublishingStatusDialogContainer(props: PublishingStatusDialogCon
   const user = useActiveUser();
   const siteId = useActiveSiteId();
   const userRoles = user?.rolesBySite[siteId];
-  const allowedUser = permittedRoles?.some((role) => userRoles.includes(role)) ?? false;
+  const allowedUser = permittedRoles?.some((role) => userRoles?.includes(role)) ?? false;
```
rart: No point iterating if userRoles is nullish.
```suggestion
  const allowedUser = (userRoles && permittedRoles?.some((role) => userRoles.includes(role))) ?? false;
```

rart (changes requested):

rart (approved):

PR #4078, opened 2024-07-31
[7094] It's not possible to rename content twice in a row via the form engine

jvega190:
https://github.com/craftercms/craftercms/issues/7094


rart:
Keep branch for 4.3+ enhancements.

PR #4079, opened 2024-07-31
[7095] Rename dialog offers the ability to edit a broken reference but will not allow the user to open the content form

jvega190:
https://github.com/craftercms/craftercms/issues/7095

### DIFF
--- a/templates/web/legacy/embedded_legacy_form.ftl
+++ b/templates/web/legacy/embedded_legacy_form.ftl
@@ -91,6 +91,7 @@
     const newEmbedded = CStudioAuthoring.Utils.getQueryVariable(location.search, 'newEmbedded');
     const contentTypeId = CStudioAuthoring.Utils.getQueryVariable(location.search, 'contentTypeId');
     const isNewContent = CStudioAuthoring.Utils.getQueryVariable(location.search, 'isNewContent') === 'true';
+    const formId = CStudioAuthoring.Utils.getQueryVariable(location.search, 'formId');
     const LEGACY_FORM_DIALOG_CANCEL_REQUEST = 'LEGACY_FORM_DIALOG_CANCEL_REQUEST';
 
     CStudioAuthoring.OverlayRequiredResources.loadContextNavCss();
@@ -125,6 +126,7 @@
                       type: 'EMBEDDED_LEGACY_FORM_SUCCESS',
                       refresh: true,
                       tab: type,
+                      formId,
                       action
                     }, '*');
                   },
@@ -134,6 +136,7 @@
                       type: 'EMBEDDED_LEGACY_FORM_FAILURE',
                       refresh: false,
                       tab: type,
+                      formId,
                       action: 'failure',
                       message: formatMessage(embeddedLegacyFormMessages.contentFormFailedToLoadErrorMessage)
                     }, '*');
@@ -148,6 +151,7 @@
                       refresh: false,
                       close: close,
                       tab: type,
+                      formId,
                       action: 'cancelled'
                     }, '*');
                   },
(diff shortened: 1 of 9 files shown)


jvega190:
@rart the issue is that when on a form, if you open the renameDialog and the item has dependencies, they will show up in the dialog, and if possible, the Edit icon shows up. The problem is that due to our current limitation of only one open form, we can't open the dependency's form too.
I created this PR with a prop `allowEditDependencies` (defaults to true), and set it to false when opening from a form for you to analize this alternative

rart (changes requested):
@jvega190 let's add a separate form dialog where it is needed for this action and use that instead of dispatching the regular edit action

jvega190:
This is ready for review, for 4.3

ui/app/src/components/LegacyFormDialog/EmbeddedLegacyContainer.tsx
```diff
@@ -95,6 +96,7 @@ export const EmbeddedLegacyContainer = React.forwardRef(function EmbeddedLegacyE
       fieldsIndexes[id] = index;
     });
   }
+  const formId = useMemo(() => uuid(), []);
```
rart:
```suggestion
  const formId = useId();
```

ui/app/src/components/LegacyFormDialog/EmbeddedLegacyContainer.tsx
```diff
@@ -51,6 +51,7 @@ import { useStyles } from './styles';
 import { hasEditAction } from '../../utils/content';
 import { nnou } from '../../utils/object';
 import { useFetchItem } from '../../hooks/useFetchItem';
+import { nanoid as uuid } from 'nanoid';
```
rart: x

ui/app/src/components/RenameAssetDialog/RenameAssetDialogContainer.tsx
```diff
@@ -68,6 +71,16 @@ export function RenameAssetDialogContainer(props: RenameAssetContainerProps) {
...
 
+  useEffect(() => {
+    const hostToHost$ = getHostToHostBus();
+    const subscription = hostToHost$.pipe(filter((e) => e.type === contentEvent.type)).subscribe(() => {
+      dispatch(fetchRenameAssetDependants());
+    });
+    return () => {
+      subscription.unsubscribe();
+    };
+  }, [dispatch]);
```
rart: Let's combine this and the effect above into one.

ui/app/src/components/RenameContentDialog/RenameContentDialog.tsx
```diff
@@ -56,7 +59,21 @@ export function RenameContentDialog(props: RenameContentDialogProps) {
...
+
+  useEffect(() => {
+    const hostToHost$ = getHostToHostBus();
+    const subscription = hostToHost$.pipe(filter((e) => e.type === contentEvent.type)).subscribe(() => {
+      loadDependants(siteId, path, value);
+    });
+    return () => {
+      subscription.unsubscribe();
+    };
+  }, [siteId, path, value, loadDependants]);
```
rart: Same here...
```suggestion
  useEffect(() => {
    loadDependants(siteId, path, value);
    const hostToHost$ = getHostToHostBus();
    const subscription = hostToHost$.pipe(filter((e) => e.type === contentEvent.type)).subscribe(() => {
      loadDependants(siteId, path, value);
    });
    return () => {
      subscription.unsubscribe();
    };
  }, [siteId, path, value, loadDependants]);
```

templates/web/legacy/embedded_legacy_form.ftl
```diff
@@ -91,6 +91,7 @@
     const newEmbedded = CStudioAuthoring.Utils.getQueryVariable(location.search, 'newEmbedded');
     const contentTypeId = CStudioAuthoring.Utils.getQueryVariable(location.search, 'contentTypeId');
     const isNewContent = CStudioAuthoring.Utils.getQueryVariable(location.search, 'isNewContent') === 'true';
+    const formId = CStudioAuthoring.Utils.getQueryVariable(location.search, 'formId');
```
rart, line 94: Use id will likely have characters like `:`; URL-encode/decode where necessary here and when receiving the message from this frame.

ui/app/src/components/RenameDialogBody/RenameItemView.tsx
```diff
@@ -68,9 +70,20 @@ export function RenameItemView(props: RenameItemViewProps) {
...
   const siteId = useActiveSiteId();
   const dispatch = useDispatch();
+  const [legacyFormDialogState, setLegacyFormDialogState] = useSpreadState({
+    open: false,
+    path: null
+  });
 
   const handleEditorDisplay = (item: DetailedItem) => {
-    openItemEditor(item, authoringBase, siteId, dispatch, fetchRenameAssetDependants());
+    if (item.systemType === 'component' || item.systemType === 'page') {
```
rart: Why is the handling different based on system type?

rart (changes requested):

ui/app/src/components/RenameDialogBody/RenameItemView.tsx
jvega190: At this point we may have for example assets/templates/etc, and the else condition handle those that don't use a form

rart: Do we have a system type for taxonomy, or does it come as component? Check that the case is handled.

rart (commented):
See comments

ui/app/src/components/RenameDialogBody/RenameItemView.tsx
jvega190: There's a system type for taxonomy, updated the validation. Actually there are a couple of details related to taxonomies here, will ping you to see if a new ticket is required for it

jvega190: Updated to consider taxonomy type

jvega190:
Closing this PR now that we have a dialogs system that supports multiple dialogs.

PR #4086, opened 2024-08-05
[7018] Fix max condition validation

jvega190:
https://github.com/craftercms/craftercms/issues/7018


static-assets/components/cstudio-forms/controls/node-selector.js
```diff
@@ -312,7 +312,14 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField,
                 'message-dialog',
                 CStudioAuthoring.Operations.simpleDialogTypeINFO,
                 CMgs.format(langBundle, 'notification'),
-                CMgs.format(langBundle, 'addMoreItemsError'),
+                _self.formatMessage(
+                  {
+                    id: 'nodeSelector.maxItemsReached',
```
jvega190: @rart without id it breaks with the following error:
`Error: [@formatjs/intl] An `id` must be provided to format a message.`

rart: Needs to go to i18n-legacy for extraction (and auto-id) to work

rart (changes requested):

rart (approved):

PR #4089, opened 2024-08-05
[7070] Drop Target tool in Right Sidebar should be filtered to content types that apply to the screen or page the user is editing

jvega190:
https://github.com/craftercms/craftercms/issues/7070


jvega190:
@rart will ping you tomorrow to discuss the content-types update => update allowedTypes

jvega190:
There are 3 parts on this work: 
1 - Editing the content-type: That can cause the compatible types to change, so we added the alert on the Panels saying that Preview needs to be reloaded.
2 - Removing content using XB: If the content had content inside registered, the inner content was not being deregistered when deleting the content, causing the panels not to update the compatible types.
3 - The contentTypes update event was not being triggered from the configurationEvent (websocket). So it was updated to handle the websocket event and trigger contentTypeUpdated.

rart (approved):

PR #4090, opened 2024-08-06
[4.1.x][6988] Path selection not working with 31 folders

jvega190:
https://github.com/craftercms/craftercms/issues/6988


jvega190:
The PathNavigatorTree's showPaginationOptions doesn't apply in 4.1.x as there is no pagination for the PathNavTree

rart (approved):

PR #4091, opened 2024-08-06
[4.1.x][7020] Add and Browse Components sidebars should be sorted alphabeticaly

jvega190:
https://github.com/craftercms/craftercms/issues/7020

### DIFF
--- a/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
+++ b/ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
@@ -50,9 +50,9 @@ export function PreviewBrowseComponentsPanel() {
   const contentTypesBranch = useSelection((state) => state.contentTypes);
   const editMode = useSelection((state) => state.preview.editMode);
   const contentTypes = contentTypesBranch.byId
-    ? Object.values(contentTypesBranch.byId).filter(
-        (contentType) => contentType.type === 'component' && !contentType.id.includes('/level-descriptor')
-      )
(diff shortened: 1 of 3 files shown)


ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
```diff
@@ -66,7 +66,7 @@ export function PreviewBrowseComponentsPanel() {
 
   useEffect(() => {
     if (siteId && contentTypesBranch.isFetching === false) {
-      dispatch(fetchComponentsByContentType({}));
+      dispatch(fetchComponentsByContentType({ sortBy: 'internalName', sortOrder: 'asc' }));
```
rart, line 69: I don't see the types for the `fetchComponentsByContentType` action receiving sorting related stuff. Not sure if the service does but there seems to be a missing piece. 

Also, if the service does, what's the need of the sort above on line 55?

rart (changes requested):

ui/app/src/components/PreviewBrowseComponentsPanel/PreviewBrowseComponentsPanel.tsx
jvega190: Yeah, updated `ComponentsContentTypeParams`. Also, the [reducer](https://github.com/craftercms/studio-ui/blob/support/4.1.x/ui/app/src/state/reducers/preview.ts#L481) sets the query that is used for the search.

The sort on line 55 is for the content-types, the sort params are for the items of the content-type

rart (approved):

PR #4095, opened 2024-08-08
[7020] Add and Browse Components sidebars should be sorted alphabetically

jvega190:
https://github.com/craftercms/craftercms/issues/7020


rart (approved):

rart:
I'm ok with merging for 4.2.

PR #4102, opened 2024-08-14
[6697] Fix select file button not fully disabled when uploading

jvega190:
https://github.com/craftercms/craftercms/issues/6697


ui/app/src/components/SingleFileUpload/SingleFileUpload.tsx
```diff
@@ -392,6 +392,7 @@ export function SingleFileUpload(props: SingleFileUploadProps) {
             type="file"
             onChange={onChange}
             onClick={onInputClick}
+            disabled={disableInput}
```
jvega190, line 395: The button (see code below) was being disabled, but the input wasn't, so even though the button was visually disabled, the input's onClick action was still being triggered.

rart (approved):

PR #4108, opened 2024-08-20
[7184] XB - Node selector with files datasource issues

jvega190:
https://github.com/craftercms/craftercms/issues/7184

### DIFF
--- a/ui/guest/src/elementRegistry.ts
+++ b/ui/guest/src/elementRegistry.ts
@@ -79,7 +79,10 @@ export function setLabel(record: ElementRecord): void {
                   `${component.craftercms.label} (${getCachedContentType(component.craftercms.contentTypeId).name})`
                 );
               } else {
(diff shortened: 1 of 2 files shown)


ui/guest/src/elementRegistry.ts
```diff
@@ -79,7 +79,10 @@ export function setLabel(record: ElementRecord): void {
                   `${component.craftercms.label} (${getCachedContentType(component.craftercms.contentTypeId).name})`
                 );
               } else {
-                labels.push(`${field.name}`);
+                // When there's a symmetric combination (between fieldId and index), the label is the field name plus the index.
```
rart:
```suggestion
                // Cases where the item selector has files (not a model), we want the label to be similar to that of repeat group field items: the field name and the item position (e.g. "Features item # 1").
```

rart (approved):

PR #4112, opened 2024-08-21
[7188] Search - Too much spacing between items in line view mode

jvega190:
https://github.com/craftercms/craftercms/issues/7188


rart (changes requested):
Please confirm browse doesn't have the same issue?

jvega190:
Browse doesn't have the issue:
image

rart (approved):

PR #4116, opened 2024-08-22
[7178] XB - repeating group items with non-referenced items by parent don't show full set of actions

jvega190:
https://github.com/craftercms/craftercms/issues/7178

### DIFF
--- a/ui/guest/src/react/ZoneMenu.tsx
+++ b/ui/guest/src/react/ZoneMenu.tsx
@@ -141,7 +141,9 @@ export function ZoneMenu(props: ZoneMenuProps) {
 
   // Use componentId to find the container modelId. If not a component, is unlikely to have a container, but in
   // cases where a page is being referenced by another model, we fall back to modelId.
-  const containerModelId = findParentModelId(componentId ?? modelId, modelHierarchyMap, models);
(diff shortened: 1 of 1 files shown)


ui/guest/src/react/ZoneMenu.tsx
```diff
@@ -135,9 +135,13 @@ export function ZoneMenu(props: ZoneMenuProps) {
     // endregion
     [modelId, recordType, iceRecord]
   );
-
-  const isItemFile = collection ? Boolean(collection[elementIndex]?.hasOwnProperty('key')) : false;
-  const collectionContainsFiles = collection ? collection.some((item) => item.hasOwnProperty('key')) : false;
+  // File validations only applies to node-selector, not to repeating-group
```
jvega190, line 139: While testing, I saw that duplicate option was not showing in a repeat-group with taxonomies, since they have a 'key' property. In general, this `isItemFile` and `collectionContainsFiles` validation was done to handle files inside node-selector.

rart, line 139:
```suggestion
  // TODO: Revisit how we detect files. Checking it has a key property doesn't feel robust.
  // File validations only applies to node-selector, not to repeating-group
```

rart (approved):

PR #4137, opened 2024-08-30
[7225] TinyMCE: fix dark theme in Insert Template

jvega190:
https://github.com/craftercms/craftercms/issues/7225

### DIFF
--- a/static-assets/js/tinymce-plugins/template/plugin.js
+++ b/static-assets/js/tinymce-plugins/template/plugin.js
@@ -381,8 +381,6 @@
       const contentStyle = (_a = getContentStyle(editor)) !== null && _a !== void 0 ? _a : '';
       const cors = shouldUseContentCssCors(editor) ? ' crossorigin="anonymous"' : '';
       const templatesCSS = editor.options.get('templates_css') ?? [];
-      console.log('templatesCSS', templatesCSS);
-      console.log('contentCSS', editor.contentCSS);
       global$2.each([...editor.contentCSS, ...templatesCSS], (url) => {
         contentCssEntries +=
           '<link type="text/css" rel="stylesheet" href="' + editor.documentBaseURI.toAbsolute(url) + '"' + cors + '>';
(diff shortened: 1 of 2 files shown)


static-assets/js/tinymce-plugins/template/plugin.js
```diff
@@ -23,387 +23,422 @@
...
-        global$2.each(editor.contentCSS, url => {
-          contentCssEntries += '<link type="text/css" rel="stylesheet" href="' + editor.documentBaseURI.toAbsolute(url) + '"' + cors + '>';
+  const getPreviewContent = (editor, html) => {
+    var _a;
+    let previewHtml = parseAndSerialize(editor, html);
+    if (html.indexOf('<html>') === -1) {
+      let contentCssEntries = '';
+      const contentStyle = (_a = getContentStyle(editor)) !== null && _a !== void 0 ? _a : '';
+      const cors = shouldUseContentCssCors(editor) ? ' crossorigin="anonymous"' : '';
+      const templatesCSS = editor.options.get('templates_css') ?? [];
```
jvega190, line 383: @rart prettier ended up formatting the whole file, but the changes are lines 383 and 384. Please let me know if you want a non-prettified version of the file only with the changes

static-assets/js/tinymce-plugins/template/plugin.js
```diff
@@ -23,387 +23,422 @@
...
+      global$2.each([...editor.contentCSS, ...templatesCSS], (url) => {
```
rart, line 384: This would throw if contentCSS is null or undefined.
```suggestion
      global$2.each([...(editor.contentCSS??[]), ...templatesCSS], (url) => {
```

But also, do we still want the contentCSS here?

rart (changes requested):

static-assets/js/tinymce-plugins/template/plugin.js
jvega190: I checked the scenario and when there is no contentCSS set (from editor initialization) its value is an empty array. Also, if you set a value to contentCSS as we do in the forms engine control, it'll apply to the template content too. Do we want to keep that, or to manually set it on the forms engine control too?

rart (approved):

PR #4140, opened 2024-09-03
[7238] Save and minimize then save and close in a new embeded content cause an error

jvega190:
https://github.com/craftercms/craftercms/issues/7238

### DIFF
--- a/static-assets/components/cstudio-forms/data-sources/embedded-content.js
+++ b/static-assets/components/cstudio-forms/data-sources/embedded-content.js
@@ -68,7 +68,6 @@ YAHOO.extend(CStudioForms.Datasources.EmbeddedContent, CStudioForms.CStudioFormD
         false,
         {
           success: function (contentTO, editorId, name, value) {
-            debugger;
             if (!_self.inserted) {
               control.insertItem(name, value, null, null, _self.id);
               _self.inserted = true;
--- a/static-assets/components/cstudio-forms/data-sources/embedded-content.js
+++ b/static-assets/components/cstudio-forms/data-sources/embedded-content.js
@@ -48,7 +48,16 @@ YAHOO.extend(CStudioForms.Datasources.EmbeddedContent, CStudioForms.CStudioFormD
         false,
         {
           success: function (contentTO, editorId, name, value) {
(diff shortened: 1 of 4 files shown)


static-assets/components/cstudio-forms/data-sources/embedded-content.js
```diff
@@ -39,6 +40,7 @@ YAHOO.extend(CStudioForms.Datasources.EmbeddedContent, CStudioForms.CStudioFormD
   itemsAreContentReferences: true,
 
   createElementAction: function (control, _self) {
+    _self.inserted = false;
```
rart, line 43: Should this be cleared at some point or is the instantiation lifecycle enough to manage that state?

static-assets/components/cstudio-forms/data-sources/embedded-content.js
```diff
@@ -66,7 +68,16 @@ YAHOO.extend(CStudioForms.Datasources.EmbeddedContent, CStudioForms.CStudioFormD
...
+            if (!_self.inserted) {
+              control.insertItem(name, value, null, null, _self.id);
+              _self.inserted = true;
+            } else {
+              // Recently added item is in the last position
+              const itemIndex = control.items.length - 1;
+              // When adding a new embedded content, the form may be saved using the 'Save draft' or the 'Save & Minimize'
+              // options. When it happens, the next save operation will be an edition, not a creation.
+              control.updateEditedItem({ value }, _self.id, itemIndex);
+            }
```
rart, line 89: - What about the path above (the if case)?
- I'm not completely in context, but not sure the last position will always contain the item that's being created or edited?

rart (changes requested):

static-assets/components/cstudio-forms/data-sources/embedded-content.js
jvega190: In this case (when creating a new node-selector content) it'll always be at the last position. Also, the update call in here will refer only to the newly created item, that again, will be at the last position

jvega190:
Closing this PR as we opted for a different approach

PR #4146, opened 2024-09-05
[7253] No minimum width for Project Tools sidebar when scaling manually

jvega190:
https://github.com/craftercms/craftercms/issues/7253


rart (changes requested):
The same happens in Global app. Perhaps we should add the minimum to the ResizableDrawer; it already has a maxWidth prop, we should have minWidth too there with this default for all.

rart (approved):

PR #4158, opened 2024-09-18
[7287] S3 Image/Video datasource filtering by Content-Type not work

jvega190:
https://github.com/craftercms/craftercms/issues/7287

### DIFF
--- a/static-assets/components/cstudio-browse/browseS3.js
+++ b/static-assets/components/cstudio-browse/browseS3.js
@@ -429,7 +429,7 @@
       site = CStudioAuthoring.Utils.getQueryParameterByName('site'),
       filter = CStudioAuthoring.Utils.getQueryParameterByName('filter');
 
-    CrafterCMSNext.services.aws.list(site, profileId, { path, filter }).subscribe(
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-browse/browseS3.js
```diff
@@ -429,7 +429,7 @@
       site = CStudioAuthoring.Utils.getQueryParameterByName('site'),
       filter = CStudioAuthoring.Utils.getQueryParameterByName('filter');
 
-    CrafterCMSNext.services.aws.list(site, profileId, { path, filter }).subscribe(
+    CrafterCMSNext.services.aws.list(site, profileId, { path, type: filter }).subscribe(
```
jvega190, line 432: @rart the UI uses filter, but the service param is `type` (checked in the [API docs](https://docs.craftercms.org/en/4.1/_static/api/studio.html#tag/aws/operation/listItems)). I tested in all of the s3 browse datasources (image/video/file) and it's working properly. I'll add a comment on the ticket for all of those DS to be tested

rart (approved):

PR #4161, opened 2024-09-20
[7309] Video controller dialogue does not close after uploading video to S3

jvega190:
https://github.com/craftercms/craftercms/issues/7309

### DIFF
--- a/static-assets/components/cstudio-forms/data-sources/video-S3-upload.js
+++ b/static-assets/components/cstudio-forms/data-sources/video-S3-upload.js
@@ -57,8 +57,8 @@ YAHOO.extend(CStudioForms.Datasources.VideoS3Upload, CStudioForms.CStudioFormDat
 
     var callback = {
       success: function (fileData) {
-        var uri = fileData;
-        var fileExtension = fileData.split('.').pop();
+        var uri = fileData.url ? fileData.url : fileData;
+        var fileExtension = uri.split('.').pop();
 
         var videoData = {
           previewUrl: uri,


static-assets/components/cstudio-forms/data-sources/video-S3-upload.js
```diff
@@ -57,8 +57,8 @@ YAHOO.extend(CStudioForms.Datasources.VideoS3Upload, CStudioForms.CStudioFormDat
 
     var callback = {
       success: function (fileData) {
-        var uri = fileData;
-        var fileExtension = fileData.split('.').pop();
+        var uri = fileData.url ? fileData.url : fileData;
+        var fileExtension = uri.split('.').pop();
```
rart, line 61: What's the reason for the property being one thing or the other? If it is the API, it documented there?

rart (approved):
.

rart (changes requested):
@jvega190 confirm the second file on the 4.1.x PR is not missing from this PR.

jvega190:
@rart the second file is not missing in this PR, was an issue only happening in 4.1.x (was already fixed in here)

static-assets/components/cstudio-forms/data-sources/video-S3-upload.js
jvega190: According to the API, it's the `fileData.url`, though I got confused at it is that way in the image datasource. Should I update them as part of this PR?

rart (approved):

PR #4171, opened 2024-09-30
[7339] Enhance RTE paste to support basic formatting

jvega190:
https://github.com/craftercms/craftercms/issues/7339

### DIFF
--- a/static-assets/components/cstudio-forms/controls/rte.js
+++ b/static-assets/components/cstudio-forms/controls/rte.js
@@ -399,7 +399,7 @@ CStudioAuthoring.Module.requireModule(
                 ? 'dark'
                 : 'default',
             external_plugins: external,
-            paste_remove_styles_if_webkit: false,
+
             file_picker_callback: function (cb, value, meta) {
               // meta contains info about type (image, media, etc). Used to properly add DS to dialogs.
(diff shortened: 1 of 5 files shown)


static-assets/js/tinymce-plugins/craftercms_paste_extension/plugin.js
```diff
@@ -21,11 +21,17 @@
   var hookPluginName = 'craftercms_tinymce_hooks';
 
   function removeElAttributes(element) {
+    const textDecoration = element.style?.textDecoration;
```
jvega190: Here's where we remove attributes, but keep only text-decoration

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -398,7 +399,7 @@ CStudioAuthoring.Module.requireModule(
                 ? 'dark'
                 : 'default',
             external_plugins: external,
-
+            paste_remove_styles_if_webkit: false,
```
jvega190: By setting this prop to false, styles are allowed when pasted, but still filtered only to allow text-decoration (for strikethrough and underline).

rart (approved):

PR #4177, opened 2024-10-04
[7363] Search results not refreshing deleted items in view after delete action

jvega190:
https://github.com/craftercms/craftercms/issues/7363


rart (approved):
Safe for 4.2

PR #4182, opened 2024-10-10
[7315] Pressing the arrow keys in certain BP sections of components locks the content for the user

jvega190:
https://github.com/craftercms/craftercms/issues/7315


jvega190:
Left as draft as it is for 4.3

rart (approved):

PR #4183, opened 2024-10-11
[6872] XB drops request parameters on page view

jvega190:
https://github.com/craftercms/craftercms/issues/6872


jvega190:

rart (approved):

PR #4188, opened 2024-10-15
[7396] RTE Cut/paste action not working

jvega190:
https://github.com/craftercms/craftercms/issues/7396


static-assets/js/tinymce-plugins/craftercms_paste/plugin.js
```diff
@@ -1540,7 +1540,7 @@
     );
   };
   var setHtml5Clipboard = function (clipboardData, html, text) {
-    if (hasWorkingClipboardApi(clipboardData)) {
+    if (clipboardData) {
```
jvega190, line 1543: Updated to match v7 paste code (just check if clipboardData exists)

rart (approved):

PR #4205, opened 2024-10-23
[7433] Able to bypass character limit in Simple Text field through XB

jvega190:
https://github.com/craftercms/craftercms/issues/7433

### DIFF
--- a/static-assets/js/tinymce-plugins/craftercms_paste/plugin.js
+++ b/static-assets/js/tinymce-plugins/craftercms_paste/plugin.js
@@ -1739,18 +1739,16 @@
   };
 
   var setup$1 = function (editor) {
-    var plugin = editor.plugins.paste;
-    var preProcess = getPreProcess(editor);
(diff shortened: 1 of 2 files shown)


jvega190:

static-assets/js/tinymce-plugins/craftercms_paste/plugin.js
```diff
@@ -1739,18 +1739,16 @@
   };
 
   var setup$1 = function (editor) {
-    var plugin = editor.plugins.paste;
-    var preProcess = getPreProcess(editor);
-    if (preProcess) {
-      editor.on('PastePreProcess', function (e) {
-        preProcess.call(plugin, plugin, e);
-      });
+    const processEvent = (f) => (e) => {
```
jvega190, line 1742: Update for pre/post process to properly return editor in the callback

rart (approved):

PR #4215, opened 2024-10-30
[7465] Avoid calling dashboard APIs twice on load

jvega190:
https://github.com/craftercms/craftercms/issues/7465

### DIFF
--- a/ui/app/src/components/PendingApprovalDashlet/PendingApprovalDashlet.tsx
+++ b/ui/app/src/components/PendingApprovalDashlet/PendingApprovalDashlet.tsx
@@ -110,6 +110,7 @@ export function PendingApprovalDashlet(props: PendingApprovalDashletProps) {
   const dispatch = useDispatch();
   const filterState = useDashletFilterState('pendingApprovalDashlet');
   const refs = useUpdateRefs({
+    items,
     currentPage,
     filterState,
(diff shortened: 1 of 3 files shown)


ui/app/src/components/PendingApprovalDashlet/PendingApprovalDashlet.tsx
```diff
@@ -164,7 +165,10 @@ export function PendingApprovalDashlet(props: PendingApprovalDashletProps) {
   }, [loadPage, refs]);
 
   useEffect(() => {
-    refs.current.loadPagesUntil(refs.current.currentPage, filterState.selectedTypes);
+    // To avoid re-fetching when it first loads
+    if (refs.current.items) {
```
jvega190, line 169: @rart when loading the dashlet, there are two effects that call functions that end up calling the same API. 
loadPage is the one we want to call when loading the dashlet.
loadPagesUntil is used to retrieve all pages until page 'x', useful when reloading the current page, changing a filter in the dashlet (this is the reason why the effect is used) and reloading due to a websocket event. We load all of the pages because there may be changes (for example the number of items is now reduced, so current page is not valid. That's considered in loadPagesUntil).

rart (approved):

PR #4224, opened 2024-11-01
[6608] Setting a state for all selected items across pages will return an HTTP 500 error and no error message in the UI

jvega190:
https://github.com/craftercms/craftercms/issues/6608


jvega190:
Added error handling when setItemState requests fail. The actual error in the API seems to be a bug in the backend, I'll add a comment in the ticket

rart (approved):

PR #4231, opened 2024-11-05
[988E] Upgrade Cropper.js to the latest version

jvega190:
https://github.com/craftersoftware/craftercms/issues/988


rart:
Work should be destined to FE2

PR #4234, opened 2024-11-06
[7488] Audit Management - misaligned content in cells

jvega190:
https://github.com/craftercms/craftercms/issues/7488

### DIFF
--- a/ui/app/src/components/AuditGrid/styles.ts
+++ b/ui/app/src/components/AuditGrid/styles.ts
@@ -24,6 +24,8 @@ export const useStyles = makeStyles()((theme) => ({
     }
   },
   cellRoot: {
+    display: 'flex',
+    alignItems: 'center',
     '&:focus-within': {
       outline: 'none !important'
     }


ui/app/src/components/AuditGrid/styles.ts
```diff
@@ -24,6 +24,8 @@ export const useStyles = makeStyles()((theme) => ({
     }
   },
   cellRoot: {
+    display: 'flex',
```
jvega190, line 27: Added the styles here since that's already being used all over the Data Grid columns definition https://github.com/craftercms/studio-ui/blob/develop/ui/app/src/components/AuditGrid/AuditGridUI.tsx#L192, adding sx would imply more updates (thinking about the release)

rart (changes requested):
The approach is ok, but please double check and confirm all cells have single element or text node values. If there are multiple elements in a flex container, the alignment might not be what we want horizontally.

jvega190:
@rart confirming that all cells have a single typography or a tooltip > button

rart (approved):

PR #4235, opened 2024-11-06
[7440] Keep the size of Search & install plugin window the same when searching for plugins

jvega190:
https://github.com/craftercms/craftercms/issues/7440


ui/app/src/components/MarketplaceDialog/InstallPluginDialogContainer.tsx
```diff
@@ -214,7 +214,8 @@ export function InstallPluginDialogContainer(props: InstallPluginDialogProps) {
             styles={{
               root: {
                 flexGrow: 1,
-                justifyContent: 'center'
+                justifyContent: 'center',
+                minHeight: 'unset'
```
rart, line 218: What happens without this?

rart (approved):

ui/app/src/components/MarketplaceDialog/InstallPluginDialogContainer.tsx
jvega190: The LoadingState default minHeight is '100%', and in this view it causes a scroll bar

PR #4236, opened 2024-11-06
[7445] Eliminate Edit button when viewing past commit files from Item History

jvega190:
https://github.com/craftercms/craftercms/issues/7445


ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
```diff
@@ -307,6 +314,26 @@ export function HistoryDialogContainer(props: HistoryDialogContainerProps) {
...
+            item,
+            rootPath: getRootPath(item.path)
+          })
+        );
+      });
+    return () => {
+      subscription.unsubscribe();
+    };
+  }, [dispatch, item]);
+  // endregion
```
rart, line 342: There's something wrong with the `fetchItemVersions` and possibly with `versionsChangeItem` too. The epic checks properties that don't match the payload definition. Then, the reducer types the payload as something different.

Also, the action should be light on what it requires. If it just needs the path, don't expect the whole item. If the rootPath, etc is not needed, shouldn't be there. The `HistoryConfigProps` interface seems underused.

Check that family of actions, epics and reducers. If this becomes big, we might have to open a new ticket to fix.


```suggestion
  // region Item Updates Propagation
  const itemPath = item?.path;
  useEffect(() => {
    const events = [contentEvent.type];
    const hostToHost$ = getHostToHostBus();
    const subscription = hostToHost$
      .pipe(filter((e) => events.includes(e.type) && e.payload.targetPath === itemPath))
      .subscribe(() => {
        dispatch(
          fetchItemVersions({
            // add what's needed after fixing types
          })
        );
      });
    return () => {
      subscription.unsubscribe();
    };
  }, [dispatch, itemPath]);
  // endregion
```

ui/app/src/components/PreviewDialog/utils.ts
```diff
@@ -28,6 +28,7 @@ export interface PreviewDialogBaseProps {
   content?: string;
   mimeType?: string;
   backgroundModeIndex?: number;
+  hideEdit?: boolean;
```
rart: Default to true on the component.
```suggestion
  showEdit?: boolean;
```

rart (changes requested):
Check if there's anything for this ticket that should be carried to the new Diff UX branch

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
jvega190: I'd prefer to make those updates on the diff dialog updates branch, or after merging it since there can be multiple conflicts. What do you think?

rart (approved):

ui/app/src/components/HistoryDialog/HistoryDialogContainer.tsx
rart: ok

PR #4240, opened 2024-11-08
[7495] Entering a number from keyboard in Publish - Schedule date is causing the dialog to shown an error.

jvega190:
https://github.com/craftercms/craftercms/issues/7495

### DIFF
--- a/ui/app/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/ui/app/src/components/DateTimePicker/DateTimePicker.tsx
@@ -265,24 +265,13 @@ function DateTimePicker(props: DateTimePickerProps) {
                   : () => {
                       setTimePickerOpen(true);
                     },
-                inputProps: {
-                  onChange: handlePopupOnlyInputChange,
-                  value: asLocalizedDateTime(internalDate, localeCode, {
-                    hour12,
-                    hour: '2-digit',
-                    minute: '2-digit',
-                    // If the timezone control isn't displayed, the time displayed may
-                    // be misleading/unexpected to the user, so if timezone isn't displayed,
-                    // display timezone here.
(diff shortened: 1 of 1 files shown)


ui/app/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -264,16 +266,7 @@ function DateTimePicker(props: DateTimePickerProps) {
...
                     },
                 inputProps: {
-                  onChange: handlePopupOnlyInputChange,
-                  value: asLocalizedDateTime(internalDate, localeCode, {
-                    hour12,
-                    hour: dateTimeFormatOptions?.hour || '2-digit',
-                    minute: dateTimeFormatOptions?.minute || '2-digit',
-                    // If the timezone control isn't displayed, the time displayed may
-                    // be misleading/unexpected to the user, so if timezone isn't displayed,
-                    // display timezone here.
```
rart, line 274: We need to address this somehow.

rart (changes requested):

ui/app/src/components/DateTimePicker/DateTimePicker.tsx
jvega190: what if we go back to setting the value with the `asLocalizedDateTime` but with the default format values (`2-digit` for hour and minute), and that way we keep the displaying of the timezone when needed?

btw, the showTimeZoneSelector prop is not currently being set to false in any usage

jvega190: I pushed the change I mentioned in the comment above, please let me know if not what we want so I can revert.

rart (changes requested):
You didn't test the `!showTimeZoneSelector` case... The keyboard doesn't work well.

rart (approved):

PR #4333, opened 2025-01-23
[7688] Fix Project Tools not rendering when opening from the Launcher

jvega190:
https://github.com/craftercms/craftercms/issues/7688

Refactor site tools to use context for tool management

Introduced a SiteToolsContext to simplify tool management and navigation logic. Replaced direct navigation calls with context-based operations, ensuring more consistent state across embedded and URL-driven site tools. Adjusted translation utils and configuration recovery logic for better error handling and user notifications.

PR #4334, opened 2025-01-23
[4.2.x][7688] Import 'SiteToolsContext' directly from file

jvega190:
https://github.com/craftercms/studio-ui/pull/4333


rart (changes requested):
Let's remove the mountMode prop also from the components that render SiteTools.

rart (approved):

PR #4362, opened 2025-02-13
[5810] Restore Content Type Form Controllers

rart:
craftercms/craftercms#5810

### DIFF
--- a/static-assets/components/cstudio-admin/mods/content-types.js
+++ b/static-assets/components/cstudio-admin/mods/content-types.js
@@ -3501,6 +3501,9 @@
           CStudioForms.Util.escapeXml(formDef.quickCreatePath) +
           '</quickCreatePath>\r\n';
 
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-admin/mods/content-types.js
```diff
@@ -3501,6 +3501,9 @@
           CStudioForms.Util.escapeXml(formDef.quickCreatePath) +
           '</quickCreatePath>\r\n';
 
+        xml += `\t<!-- Specifies whether a form-controller.js is present in the type definition and should be loaded by the Forms Engine -->\n`
```
rart, line 3504: FYI There's an XML comment _documenting_ the property which Github doesn't seem to be showing in this view. See [here](https://github.com/craftercms/studio-ui/blob/f4813711a10533e39aaf72551670e0f5c18f1438/static-assets/components/cstudio-admin/mods/content-types.js#L3504).

image

sumerjabri (approved):

PR #4376, opened 2025-02-24
[7769] Request confirmation before acting on "Publish All" on the "Publish On Demand" widget

jvega190:
https://github.com/craftercms/craftercms/issues/7769


rart (changes requested):
There are missing commits from 4.x here

rart (approved):

PR #4380, opened 2025-02-25
[7122] Workflow is not cancelled when renaming an item's parent

jvega190:
https://github.com/craftercms/craftercms/issues/7122

## Summary by CodeRabbit

- **New Features**
	- Enhanced the folder renaming process with additional validations to ensure active operations are properly managed.
	- Introduced a warning alert with a confirmation checkbox that notifies you of pending operations before proceeding with a rename.


ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -132,7 +165,17 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
 										setConfirm({ body: message });
 									} else {
 										if (rename) {
-											onRenameFolder(site, path, name);
+											if (containsItemsInWorkflow) {
+												const packageIds: number[] = packagesInWorkflow.map((pkg) => pkg.id);
+												cancelPackages(site, {
+													packageIds,
+													comment: `Cancel packages to rename folder "${path}"`
```
jvega190: @rart right now the comment is 'generated'. Should it have i18n?

rart: Comment and package name generation in general is something we need to discuss in detail and adjust in all the recent publishing work. I've added a task on our tracker ticket. And let's leave a TODO here.
```suggestion
													// TODO: Correct comment generation
													comment: `Cancel packages to rename folder "${path}"`
```

ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -68,16 +73,44 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
...
+					setFetchingAffectedPackages(false);
+					setPackagesInWorkflow(packages);
+				},
+				error: () => {
+					setFetchingAffectedPackages(false);
+				}
+			});
+		} else {
+			setPackagesInWorkflow(undefined);
+		}
```
rart: Let's make the else into a guard statement.
```suggestion
		if (!rename) {
		setPackagesInWorkflow(undefined);
		return;
		}
			// If renaming, check if the folder contains items in workflow
			setFetchingAffectedPackages(true);
			fetchAffectedPackages(site, path, true).subscribe({
				next: (packages) => {
					setFetchingAffectedPackages(false);
					setPackagesInWorkflow(packages);
				},
				error: () => {
					setFetchingAffectedPackages(false);
				}
			});
		
```

rart (changes requested):
- Review, understand (ask questions) and test the changes I've pushed
- One scenario I'm not too clear on how it works in practice is the suggested modified value being a path that exists. Please test and lets review that scenario

jvega190:
Reviewed changes, tested the different scenarios.

jvega190:
@rart I updated the workflow warning message to be the same as in ViewPackagesDialog (`The item is part of one or more publishing packages. Editing it will cancel the packages.`). Though since this scenario is a bit different (the parent has children in workflow, instead of being the item in workflow), I',m not sure it applies. What do you thing?

ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -242,14 +272,31 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
...
 						onChange={(event) => onInputChanges(applyFolderNameRules(event.target.value, { allowBraces }))}
 					/>
+					{rename && containsItemsInWorkflow && (
+						<Alert severity="warning" icon={false} sx={{ mb: 1 }}>
+							<FormControlLabel
+								disableTypography
+								control={<Checkbox onChange={onCancelPackagesAckChange} />}
+								label={
+									<Typography>
+										<FormattedMessage defaultMessage="The item is part of one or more publishing packages. Editing it will cancel the packages." />
```
rart:
```suggestion
										<FormattedMessage defaultMessage="The folder contains items which take part in one or more publishing packages. Renaming it will cancel the packages." />
```

ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -68,25 +74,44 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
...
-	const isValid = !isBlank(name) && !folderExists && (!rename || (name !== value && moveFolderAck));
+	const [fetchingAffectedPackages, setFetchingAffectedPackages] = useState(false);
+	const [packagesInWorkflow, setPackagesInWorkflow] = useState(undefined);
+	const containsItemsInWorkflow = packagesInWorkflow?.length > 0;
+	const [cancelPackagesAck, setCancelPackagesAck] = useState(false);
+	const isValid =
+		!isBlank(name) &&
+		!folderExists &&
+		(!rename || (name !== value && moveFolderAck)) &&
+		(!rename || !containsItemsInWorkflow || (containsItemsInWorkflow && cancelPackagesAck));
```
rart:
```suggestion
		(!rename || (name !== value && moveFolderAck && !fetchingAffectedPackages &&(!containsItemsInWorkflow || cancelPackagesAck)));
```

ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
```diff
@@ -242,14 +272,31 @@ export function CreateFolderContainer(props: CreateFolderContainerProps) {
...
 				</form>
 			</DialogBody>
 			<DialogFooter>
 				<SecondaryButton onClick={onCloseButtonClick} disabled={isSubmitting}>
 					<FormattedMessage id="words.cancel" defaultMessage="Cancel" />
 				</SecondaryButton>
-				<PrimaryButton onClick={onSubmit} disabled={isSubmitting || !isValid} loading={isSubmitting}>
+				<PrimaryButton
+					onClick={onSubmit}
+					disabled={isSubmitting || !isValid || fetchingAffectedPackages}
```
rart:
```suggestion
					disabled={!isValid}
```

rart (changes requested):

ui/app/src/components/CreateFolderDialog/CreateFolderContainer.tsx
jvega190: We would lose the `isSubmitting` validation, should I keep it here or add it to the `isValid` var?

rart: isSubmitting is considered by the loading prop of the button

jvega190:
@rart after merging with develop I found an issue that may be with the validateActionPolicy API not returning what's expected, I'll check with Jonathan tomorrow. I'll leave this as draft for now.

rart (approved):

PR #4434, opened 2025-03-18
[4.x][7852] Able to save mandatory RTE field when it is empty

jvega190:
https://github.com/craftercms/craftercms/issues/7852


jvega190:
@rart I updated the forms-engine control to handle the `external_change` event triggered by the ace code view plugin, that is already handled in the XB control.
I also moved the `external_change` event trigger in the plugin for it to trigger in both inline and dialog modes, instead of only in dialog mode.

static-assets/js/tinymce-plugins/ace/plugin.js
```diff
@@ -131,6 +131,7 @@
             let value = aceEditor.getValue();
             editorTextareaEl.value = value;
             editor.setContent(value);
+            editor.fire('external_change');
```
rart, line 134: This seems a bit redundant. Also, it was removed from one onChange and added here. Pls explain the issue and the PR.

rart: ?

jvega190: The issue was that when updating the content using the code editor, the model was being updated (`editor.setContent(value)`)  but the field validation was not being executed. By handling the `external_change` event in the control (rte.js), whenever there's an external change it runs the validation of the field.
So, since the model was being updated but the validation was not ran (in code editor mode), the empty content was being saved when it shouldn't.

The `external_change` was moved from the onChange event to the other one because the first event is only set when in dialog mode. The event where it is being triggered now is for both inline and dialog mode (ace editor has 2 mode, inline and dialog).

rart (approved):

PR #4500, opened 2025-05-13
[7505] Improve the error message displayed when creating access tokens with an expiry time that is already past

jvega190:
https://github.com/craftercms/craftercms/issues/7505

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling in the token creation dialog to display more accurate and capitalized validation error messages.
  - Enhanced error dialogs to show detailed validation error information for better clarity.

### DIFF
--- a/ui/app/src/components/ErrorDialog/ErrorDialog.tsx
+++ b/ui/app/src/components/ErrorDialog/ErrorDialog.tsx
@@ -20,7 +20,7 @@ import CloseIcon from '@mui/icons-material/CloseRounded';
 import Dialog from '@mui/material/Dialog';
 import StandardAction from '../../models/StandardAction';
 import { ApiResponse } from '../../models/ApiResponse';
-import ApiResponseErrorState, { ApiResponseErrorStateProps } from '../ApiResponseErrorState';
+import ApiResponseErrorState, { type ApiResponseErrorStateProps } from '../ApiResponseErrorState';
 import { useUnmount } from '../../hooks/useUnmount';
 import Box from '@mui/material/Box';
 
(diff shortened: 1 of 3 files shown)


jvega190:
@rart to clarify, the date/time selector doesn't allow to select a previous date/time. But if you select a date/time and wait some minutes for it to be past, then submit, it is allowed, and it ends up with the error.

One detail here. This is the response we got from the error:
```
"response": {
        "code": 1001,
        "message": "Invalid parameter(s)",
        "remedialAction": "Check API and make sure you're sending the correct parameters",
        "documentationUrl": ""
    },
    "validationErrors": [
        {
            "field": "expiresAt",
            "message": "must be a future date"
        }
    ]
```
So by sending the error response, the dialog will show this message:
image

Should we process the validation errors so if they are non nullish we display a message using those messages?
Something like:
image

ui/app/src/components/ErrorDialog/ErrorDialog.tsx
```diff
@@ -20,13 +20,14 @@ import CloseIcon from '@mui/icons-material/CloseRounded';
 import Dialog from '@mui/material/Dialog';
 import StandardAction from '../../models/StandardAction';
 import { ApiResponse } from '../../models/ApiResponse';
-import ApiResponseErrorState from '../ApiResponseErrorState';
+import ApiResponseErrorState, { ApiResponseErrorStateProps } from '../ApiResponseErrorState';
```
rart: Please configure your IDE to add the `type` annotation on imports.
```suggestion
import ApiResponseErrorState, { type ApiResponseErrorStateProps } from '../ApiResponseErrorState';
```

rart (changes requested):

ui/app/src/components/ErrorDialog/ErrorDialog.tsx
jvega190: Done, IDE is now configured to add the type annotations, thanks!

rart (approved):

PR #4512, opened 2025-05-16
[4.x][7987] Inserting links to pages in RTE via XB does not work

jvega190:
https://github.com/craftercms/craftercms/issues/7987


ui/guest/src/controls/rte.ts
```diff
@@ -177,7 +179,8 @@ export function initTinyMCE(
         )
         .subscribe(({ payload }) => {
           if (payload) {
-            cb(payload.path, { alt: payload.name });
+            const path = meta.filetype === 'file' ? getPreviewURLFromPath(payload.path) : payload.path;
```
rart, line 184: Why alter the path for "files"?

jvega190: File type may be 'image', 'media' or 'file'. 'image' and 'media' are the ones used when adding images or videos. File type is the one returned when selecting on the link picker.
When selecting a link (using the file browse datasource) we get an absolute path, and just like in legacy RTE, we need to return the previewUrl.

rart:
@coderabbitai review

rart (approved):

PR #4513, opened 2025-05-19
[8073] Allowed types not showing on Preview Panels

jvega190:
https://github.com/craftercms/craftercms/issues/8073

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured datasource properties are consistently updated and reflect parsed values when referenced directly.  
- **New Features**
  - Expanded content type permissions to include a global wildcard, allowing broader access to components.  
  - Improved handling of wildcard content types for better classification and fetching of allowed components.  
  - Added support for a global wildcard representation in content type displays for clearer identification.

### DIFF
--- a/ui/app/src/services/contentTypes.ts
+++ b/ui/app/src/services/contentTypes.ts
@@ -408,6 +408,7 @@ function parseLegacyFormDefinition(definition: LegacyFormDefinition): ContentTyp
 	const fields: LookupTable<ContentTypeField> = {};
 	const sections: Array<ContentTypeSection> = [];
 	const dataSources: LookupTable<DataSource> = {};
+	// TODO: update type to be LookupTable<DataSource>. https://github.com/craftercms/craftercms/issues/8216
 	const dropTargetsLookup: LookupTable<LegacyDataSource> = {};
 
 	const legacyDataSourceArray = asArray(definition.datasources?.datasource);
--- a/ui/app/src/services/contentTypes.ts
+++ b/ui/app/src/services/contentTypes.ts
@@ -431,6 +431,8 @@ function parseLegacyFormDefinition(definition: LegacyFormDefinition): ContentTyp
 				//   break;
 			}
 			dataSources[datasource.id].properties[property.name] = value;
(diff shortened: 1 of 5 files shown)


ui/app/src/services/contentTypes.ts
```diff
@@ -431,6 +431,8 @@ function parseLegacyFormDefinition(definition: LegacyFormDefinition): ContentTyp
 				//   break;
 			}
 			dataSources[datasource.id].properties[property.name] = value;
+			// datasource also needs to be directly edited since it's set to dropTargetsLookup when datasource type is 'components'
+			datasource.properties[property.name] = value;
 		});
 		if (datasource.type === 'components') {
 			dropTargetsLookup[datasource.id] = datasource;
```
rart: We should avoid editing the item that was fetched or in general from the collection that's getting looped. Use the copy (see line 438ish).

Check if there's any implications of doing it this way.

Also, the comment is not clear. Improve pls.

```suggestion
			dropTargetsLookup[datasource.id] = dataSources[datasource.id];
```

rart (changes requested):

ui/app/src/services/contentTypes.ts
jvega190: @rart The thing is that `datasources` if of type `DataSource`, and dropTargetsLookup is a lookup of LegacyDataSource with is used later to parse the validations. I created a copy of the legacy datasource and set it to dropTargetsLookup. Should we consider updating the dropTargetsLookup to be of type DataSource? (maybe as part of FE2 work)

rart: Yes. Pls file ticket.

rart (approved):

ui/app/src/services/contentTypes.ts
rart: Add link to this ticket.

jvega190: https://github.com/craftercms/craftercms/issues/8216. Added TODO and link in code too.

jvega190:
The commit that was done after approval has only a comment, no code updates

rart (approved):

rart (approved):

PR #4519, opened 2025-05-29
[4.x][8073] Allowed types not showing on Preview Panels

jvega190:
https://github.com/craftercms/craftercms/issues/8073

### DIFF
--- a/ui/app/src/state/epics/contentTypes.ts
+++ b/ui/app/src/state/epics/contentTypes.ts
@@ -65,9 +65,15 @@ export default [
       ofType(fetchComponentsByContentType.type, setContentTypeFilter.type),
       withLatestFrom(state$),
       switchMap(([, state]) => {
-        const allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
+        let allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
           ([key, type]) => (type.shared ? [key] : [])
         );
+        // If '*' is included in allowedContentTypes, it means that all content types are allowed.
+        if (allowedContentTypes.includes('*')) {
+          allowedContentTypes = Object.values(state.contentTypes.byId)
+            .filter((contentType) => contentType.type === 'component' && !contentType.id.includes('/level-descriptor'))
+            .map((contentType) => contentType.id);
+        }
         return fetchItemsByContentType(
           state.sites.active,
           state.preview.components.contentTypeFilter === 'compatible'
--- a/ui/app/src/state/epics/contentTypes.ts
+++ b/ui/app/src/state/epics/contentTypes.ts
@@ -66,7 +66,7 @@ export default [
       withLatestFrom(state$),
       switchMap(([, state]) => {
         // allowedContentTypes is an array of content type IDs that are 'compatible' with preview.
-        // For a content type to be compatible, the type should have the 'shareExisting' property set to true.
+        // For a content type to be compatible (in the context of the browse components panel), the type should have the 'shareExisting' property set to true.
         // *Note that this is different from the 'shared' property, which means that new content of the type can be created.
         let allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
           ([key, type]) => (type.sharedExisting ? [key] : [])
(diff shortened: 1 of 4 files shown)


rart:
@coderabbitai review

ui/app/src/state/epics/contentTypes.ts
```diff
@@ -65,9 +65,15 @@ export default [
...
-        const allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
+        let allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
           ([key, type]) => (type.shared ? [key] : [])
         );
+        // If '*' is included in allowedContentTypes, it means that all content types are allowed.
+        if (allowedContentTypes.includes('*')) {
+          allowedContentTypes = Object.values(state.contentTypes.byId)
+            .filter((contentType) => contentType.type === 'component' && !contentType.id.includes('/level-descriptor'))
+            .map((contentType) => contentType.id);
+        }
```
rart: Above, you can see the flatMap checks for `.shared`. I'm not sure that's considered here. Please double check.

ui/guest/src/react/ZoneMarker.tsx
```diff
@@ -179,7 +179,9 @@ export function ZoneMarker(props: ZoneMarkerProps) {
                   onClick={(e) => e.stopPropagation()}
                 >
                   {Object.entries(allowedTypesMeta).map(([id, modes]) => {
-                    const type = contentTypes[id];
+                    // 'id' can be '*' (meaning all content types), so we need to handle that scenario
+                    // TODO: i18n
+                    const type = id === '*' ? { id: 'all', name: 'all types' } : contentTypes[id];
```
rart: Where does the `all` id of this on-the-fly object ends up being used? Don't see any other change consuming this ID.

What's the plan for this TODO?
At a minimum this should have proper capitalisation...
```suggestion
                    // 'id' can be '*' (meaning all content types), so we need to handle that scenario
                    // TODO: i18n
                    const type = id === '*' ? { id: 'all', name: 'All types' } : contentTypes[id];
```

rart (changes requested):

ui/guest/src/react/ZoneMarker.tsx
jvega190: The 'all' id object is only for the `getAvatarWithIconColors` to get a proper avatar.
We have some TODOS all over guest for when we include i18n to it. Updated for it to be capitalized

ui/app/src/state/epics/contentTypes.ts
jvega190: There was an error, since the property we use for allowing existing components is 'sharedExisting'. So I updated it in here and in PreviewBrowseComponentsPanel.

This epic here handles two actions:
- fetchComponentsByContentType: Used only by PreviewBrowseComponentsPanel. This one has a scenario where we fetch only compatible components. For this one we define `allowedContentTypes`.
- setContentTypeFilter: Used by PreviewBrowseComponents, PreviewComponentsPanel and PreviewInPageInstancePanel. For this action `allowedContentTypes` is not used since the fetch is done against a specific type, and we fetch components no matter if they're not compatible.

So, the update in `allowedContentTypes` won't affect usages outside PreviewBrowseComponentsPanel.

rart (approved):

ui/app/src/state/epics/contentTypes.ts
```diff
@@ -65,9 +65,18 @@ export default [
       ofType(fetchComponentsByContentType.type, setContentTypeFilter.type),
       withLatestFrom(state$),
       switchMap(([, state]) => {
-        const allowedContentTypes = Object.entries(state.preview.guest?.allowedContentTypes ?? {}).flatMap(
-          ([key, type]) => (type.shared ? [key] : [])
+        // allowedContentTypes is an array of content type IDs that are 'compatible' with preview.
+        // For a content type to be compatible, the type should have the 'shareExisting' property set to true.
```
rart:
```suggestion
        // For a content type to be compatible (in the context of the browse components panel), the type should have the 'shareExisting' property set to true.
```

rart (changes requested):

rart (approved):

PR #4522, opened 2025-05-30
[4.x][8154] Site breaks when we try to duplicate an item

jvega190:
https://github.com/craftercms/craftercms/issues/8154

### DIFF
--- a/ui/guest/src/contentController.ts
+++ b/ui/guest/src/contentController.ts
@@ -787,7 +787,7 @@ fromTopic(fetchGuestModelComplete.type).subscribe((action: StandardAction<FetchG
       mhm[id].parentContainerFieldPath = mhm[id].parentContainerFieldPath ?? hierarchyMap[id].parentContainerFieldPath;
       mhm[id].parentContainerFieldIndex =
         mhm[id].parentContainerFieldIndex ?? hierarchyMap[id].parentContainerFieldIndex;
-      // Flipping around from the fix on ticket #6433 for ticket #8154. The children coming from Host must be the latest.Add commentMore actions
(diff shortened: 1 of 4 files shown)


ui/guest/src/contentController.ts
```diff
@@ -787,7 +787,9 @@ fromTopic(fetchGuestModelComplete.type).subscribe((action: StandardAction<FetchG
       mhm[id].parentContainerFieldPath = mhm[id].parentContainerFieldPath ?? hierarchyMap[id].parentContainerFieldPath;
       mhm[id].parentContainerFieldIndex =
         mhm[id].parentContainerFieldIndex ?? hierarchyMap[id].parentContainerFieldIndex;
-      mhm[id].children = mhm[id].children ?? hierarchyMap[id].children;
+      // Flipping around from the fix on ticket #6433 for ticket #8154. The children coming from Host must be the latest.Add commentMore actions
+      // TODO: Observed the local hierarchyMap[id].children containing duplicates. Where is that coming from? Are children computed locally?
```
jvega190, line 791: @rart a couple of questions:
- In the previous comment at the end, it says 'Add commentMore actions'. What's the idea there?
- Any particular scenario where you saw the duplicate children? I wasn't able to see duplicates while duplicating and or removing content

rart: - Ignore the "add commentmore actions" looks like a typo. Don't carry that piece of the comment.
- It was seeing in a specific site. Not to worry at this time.

rart (changes requested):
Pls fix the comment

rart (approved):

PR #4525, opened 2025-06-06
[4.x][8120] User can bypass the required field checking in Content Form for the control

jvega190:
https://github.com/craftercms/craftercms/issues/8120

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -2787,6 +2787,7 @@ const initializeCStudioForms = () => {
                 }
               });
           });
+          form.sections.push(formSection);
         },
 
         /**


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -2787,6 +2787,7 @@ const initializeCStudioForms = () => {
                 }
               });
           });
+          form.sections.push(formSection);
```
jvega190, line 2790: The issue happens when rendering only some fields in the form (not the full form)
When validating the form, what the forms-engine does is to check against the list of sections rendered. In the path where selected fields are rendered, we were not setting the form section into the sections variable of the form, so when validating, sections was empty and no validations were performed.

jvega190:
Left as draft to finish testing

rart (approved):

PR #4528, opened 2025-06-12
[8202] Forms engine error opening node selector item when selecting control's data sources in certain order

jvega190:
https://github.com/craftercms/craftercms/issues/8202

## Summary by CodeRabbit

- **New Features**
	- Items now track and display the datasource from which they originate when added through selection controls.

- **Refactor**
	- Improved consistency in how datasource information is passed and managed during item insertion across selection dialogs and forms.
	- Optimized rendering by deferring item list updates until after multiple insertions complete.

### DIFF
--- a/static-assets/components/cstudio-forms/controls/node-selector.js
+++ b/static-assets/components/cstudio-forms/controls/node-selector.js
@@ -534,12 +534,12 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField, 
 		return validation;
 	},
 
-	newInsertItem: function (key, value, type) {
+	newInsertItem: function (key, value, type, datasource) {
 		const validation = this.checkValidations(key, value);
 
 		if (validation.successful) {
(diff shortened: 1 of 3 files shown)


static-assets/components/cstudio-forms/controls/node-selector.js
```diff
@@ -534,12 +534,12 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField,
 		return validation;
 	},
 
-	newInsertItem: function (key, value, type) {
+	newInsertItem: function (key, value, type, datasource) {
```
jvega190, line 537: When inserting a new item to the node-selector, there's this logic where we set the datasourceId to the item, so when editing an item of the datasource we read the datasourceId to know what data source will be used for edition.
The data source prop got lost in `newInsertItem`, causing the default datasource (first in list of datasources) to be used for edition, ending up in errors in some scenarios

rart (approved):

PR #4537, opened 2025-06-16
[4.x][8027] Opening a template editor form the content type or an item's options menu forces lowercase template names

jvega190:
https://github.com/craftercms/craftercms/issues/8027

### DIFF
--- a/ui/app/src/components/ContentTypeManagement/ContentTypeManagement.tsx
+++ b/ui/app/src/components/ContentTypeManagement/ContentTypeManagement.tsx
@@ -31,6 +31,7 @@ import {
   emitSystemEvent
 } from '../../state/actions/system';
 import { ProjectToolsRoutes } from '../../env/routes';
+import { fetchContentTypes } from '../../state/actions/preview';
 
 export interface ContentTypeManagementProps {
   embedded?: boolean;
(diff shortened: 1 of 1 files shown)


ui/app/src/components/ContentTypeManagement/ContentTypeManagement.tsx
```diff
@@ -61,6 +62,7 @@ export function ContentTypeManagement(props: ContentTypeManagementProps) {
       .subscribe((e: any) => {
         switch (e.data.type) {
           case 'CONTENT_TYPES_ON_SAVED': {
+            dispatch(fetchContentTypes());
```
jvega190, line 65: The issue was that when adding a template to the content type and saving, the list of content types were not being updated in the store. So if you wanted to edit the template for the created content it was not going to 'find' any template and then it was going to create a new one based on the content-type id (in this scenario it was the same template name but in lowercase).

jvega190:
@coderabbitai review

ui/app/src/components/ContentTypeManagement/ContentTypeManagement.tsx
rart: Make sure TB2 doesn't have this problem.

rart (approved):

PR #4538, opened 2025-06-17
[8052] Item selectors should allow navigating to the item when the form is in view mode

jvega190:
https://github.com/craftercms/craftercms/issues/8052

## Summary by CodeRabbit

- **Refactor**
  - Simplified the logic for displaying the edit button in the node selector, making it visible for all editable items regardless of user permissions.

- **Documentation**
  - Updated comments to clarify that items without write permission will open in view mode when the edit button is clicked.

### DIFF
--- a/static-assets/components/cstudio-forms/controls/node-selector.js
+++ b/static-assets/components/cstudio-forms/controls/node-selector.js
@@ -408,7 +408,9 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField, 
 					// isEditable: studio-ui has mechanisms to edit the item (e.g. a component or a text file)
 					// allowEdit: the datasource has edit capabilities (datasource.edit exists).
 					const isEditable = this.allowEdit && (isComponent || craftercms.utils.content.isEditableAsset(item.key));
-					if (isEditable && hasEditAction) {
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/controls/node-selector.js
```diff
@@ -408,7 +408,9 @@ YAHOO.extend(CStudioForms.Controls.NodeSelector, CStudioForms.CStudioFormField,
 					// isEditable: studio-ui has mechanisms to edit the item (e.g. a component or a text file)
 					// allowEdit: the datasource has edit capabilities (datasource.edit exists).
 					const isEditable = this.allowEdit && (isComponent || craftercms.utils.content.isEditableAsset(item.key));
-					if (isEditable && hasEditAction) {
+					// At this point, we only need to check if the item is editable (see definition above). If the user doesn't
+					// have write permission, the datasource.edit method will open the item in view mode.
+					if (isEditable) {
```
jvega190, line 413: Edit/view icon has only being rendered if there were edit permissions. In previous lines, there's a piece of logic for the icon (to be edit or view). So I updated it for it to render no matter the edit permissions of the item, since the datasources already handle the edit or view form opening.

rart (approved):

PR #4549, opened 2025-06-18
[8223] Preview not updating after form update

jvega190:
https://github.com/craftercms/craftercms/issues/8223

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved form save behavior to only redirect when the model path changes; otherwise, the current preview page reloads for a smoother user experience.


jvega190:
@rart this branch doesn't need the filter in the lockContentEvent, because the new fetchContentItem service returns the correct 404 if it doesn't exist. I'll create the ticket for the lockContentEvent having the outdated path

rart (approved):

PR #4558, opened 2025-06-19
[8055] /site/components/{parentPath[0]} doesn't work after two layers of "embedded" components

jvega190:
https://github.com/craftercms/craftercms/issues/8055

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of embedded forms by correctly determining the parent path from the URL if it is not set, ensuring forms function as expected in more scenarios.


jvega190:
Holding PR to discuss parentPath logic when multiple levels of embedded

rart (approved):

PR #4562, opened 2025-06-23
[4.x][7954] Page become unresponsive with form with multiple RTEs

jvega190:
https://github.com/craftercms/craftercms/issues/7954

### DIFF
--- a/static-assets/components/cstudio-forms/controls/rte.js
+++ b/static-assets/components/cstudio-forms/controls/rte.js
@@ -515,10 +515,13 @@ CStudioAuthoring.Module.requireModule(
             })
           });
 
+          const isInRepeatGroup = Boolean(config.repeatContainer);
           // Update all content before saving the form (all content is automatically updated on focusOut)
-          callback = {};
-          callback.beforeSave = function () {
-            _thisControl.save();
+          callback = {
+            beforeSave: function () {
+              _thisControl.save();
+            },
+            ...(isInRepeatGroup ? { repeatGroupId: config.repeatContainer.id } : {})
           };
           _thisControl.form.registerBeforeSaveCallback(callback);
 
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -1465,6 +1465,7 @@ const initializeCStudioForms = () => {
                 CStudioAuthoring.Utils.showConfirmDialog({
                   body: formatMessage(formEngineMessages.formNotReadyForSaving)
                 });
+                setButtonsEnabled(true);
                 return;
               }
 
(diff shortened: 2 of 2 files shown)


static-assets/themes/cstudioTheme/css/forms-default.css
rart, line 1: Where did all these style come from?
Can they be factored to not be duped?

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -279,248 +283,263 @@ CStudioAuthoring.Module.requireModule(
...
+              ...(isInRepeatGroup ? { repeatGroupId: config.repeatContainer.id } : {})
+            };
+            _thisControl.form.registerBeforeSaveCallback(callback);
           };
-          _thisControl.form.registerBeforeSaveCallback(callback);
+
+          if (config.onDemandEditorInitialization) {
+            onDemandEditorInitializationEl.addEventListener('click', () => {
+              initTiny();
+              onDemandEditorInitializationEl.style.display = 'none';
```
rart: Is there a reason not to remove it?
Could we maybe use the same element and avoid this function returning two elements?

static-assets/components/cstudio-forms/controls/rte.js
```diff
@@ -279,248 +283,263 @@ CStudioAuthoring.Module.requireModule(
...
-              case 'autoGrow':
-                this.autoGrow = prop.value == 'false' ? false : true;
-                break;
-              case 'maxlength':
-                inputEl.maxlength = prop.value;
-                break;
-              case 'enableSpellCheck':
-                this.enableSpellCheck = !prop.value || prop.value === 'true';
-                break;
+          const { inputEl, onDemandEditorInitializationEl } = this._renderInputMarkup(config, rteId);
```
rart: Not ideal to change the return signature of this function. 
Could we: 
- use the same element for the preview (inputEl)
- create another function to do this work when the property is true?
- Set the new element to a `this` property and get it there from instead of from this return value. Looks like you're already setting this element to this somewhere else to `this.onDemandEditorInitializationEl` so perhaps set it at once on the renderInputMarkup?

rart (changes requested):
- See code rabbit comments above
- Add detailed change log
- Check the content type tool works with any value or absence of the new `onDemandEditorInitialization` tag

static-assets/themes/cstudioTheme/css/forms-default.css
jvega190: The styles are from the editor's css files used to style the editor content plus some customization. The problem is that since those are styles from the lib we can't refactor them, and if we use them in the forms engine they'd end up affecting other form styles.

static-assets/components/cstudio-forms/controls/rte.js
jvega190: We can remove it, there is no problem with that.

static-assets/components/cstudio-forms/controls/rte.js
jvega190: The detail with `inputEl` is that it is a textArea which is the element used by the editor to initialize (the editor creates a new div (doesn't reuse `inputEl`).
I'm being able to avoid changing the return signature of the function by using `this.onDemandEditorInitializationEl`.

jvega190:
@rart the PR now contains only the fix for the 'form not ready' issue

rart (approved):

PR #4587, opened 2025-07-31
[8295] Error in item tree sidebar when deleting a page 

jvega190:
https://github.com/craftercms/craftercms/issues/8295

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved item deletion in the navigation tree to handle cases where the parent path may include an index, ensuring more consistent removal of items.


jvega190:
This issue is happening because when checking against the parentPathOfTargetPath we're not considering the path `withIndex`, so the removal from state never happens.
It used to work before because after doing a deletion the websocket event `PUBLISH_EVENT` was being triggered and now it is not. (@jmendeza can you confirm that the removal of the event is expected?)
@rart This lack of `withIndex` verification is also present in 4.x, but the issue doesn't happen because we refresh the tree after `PUBLISH_EVENT`.

jmendeza:
@jvega190 this issue seems to happen when deleting a `NEW` item. If the item is `NEW`, then the deletion does not need to be published. When deleting a published item, the sidebar works as expected.
cc @rart

rart (approved):

PR #4592, opened 2025-08-11
[4.x][8308] Item Selector should show view option for file types with preview support

jvega190:
https://github.com/craftercms/craftercms/issues/8308

### DIFF
--- a/static-assets/components/cstudio-common/common-api.js
+++ b/static-assets/components/cstudio-common/common-api.js
@@ -2222,7 +2222,7 @@ var nodeOpen = false,
                 } else {
                   CStudioAuthoring.SelectedContent.init();
                 }
-                callback.success && callback.success(nodeRef);
+                callback?.success?.(nodeRef);
               }
             });
           }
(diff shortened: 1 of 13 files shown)


jvega190:
@rart I moved `isVideo`, `isAudio` and `isPdfDocument` utils from PathNavigator to `utils/content` for them to be usable in the legacy code (using `craftercms.utils.content...`. That's the reason whey there are many changes in the imports of those utils

rart (changes requested):
Missing CHANGELOG transcription

rart (commented):
Why was it necessary to update file-browse-repo and file-desktop-upload for the item selector to show the preview? What's the relationship in those changes?

jvega190:
> Why was it necessary to update file-browse-repo and file-desktop-upload for the item selector to show the preview? What's the relationship in those changes?

@rart that's how the node-selector currently works with for example the `components`, `shared-content`, etc datasources. The node selector shows the action button (view/edit) and when clicking calls the `edit` function of the datasource. The datasource validates the permissions/etc and previews or opens the content in edit mode. That way the DS is the one handling the edit/preview action

rart (approved):

PR #4593, opened 2025-08-11
[4.x][8319] Publish Approval dialog should default to the NOW option when approving a request that was subitted with NOW as the preferred publish time

jvega190:
https://github.com/craftercms/craftercms/issues/8319

### DIFF
--- a/ui/app/src/utils/itemActions.ts
+++ b/ui/app/src/utils/itemActions.ts
@@ -965,7 +965,7 @@ export const itemActionDispatcher = ({
     case 'schedulePublish':
     case 'requestPublish': {
       const schedulingMap = {
-        approvePublish: null,
+        approvePublish: undefined,
         schedulePublish: 'custom',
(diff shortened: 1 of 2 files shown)


ui/app/src/utils/itemActions.ts
```diff
@@ -965,7 +965,7 @@ export const itemActionDispatcher = ({
     case 'schedulePublish':
     case 'requestPublish': {
       const schedulingMap = {
-        approvePublish: null,
+        approvePublish: undefined,
```
jvega190, line 968: When setting the `approvePublish` value to null, the default value in the component was being overwritten, causing the scheduling calculation to be wrong. If `undefined`, the scheduling it defaulted to 'now' (custom schedules are still working properly).

rart (approved):

jvega190:
I committed an update to the CHANGELOG.md file to this PR by mistake. Already reverted the update

PR #4595, opened 2025-08-12
[8319] Publish Approval dialog should default to the NOW option when approving a request that was subitted with NOW as the preferred publish time

jvega190:
https://github.com/craftercms/craftercms/issues/8319

## Summary by CodeRabbit

* **Bug Fixes**
  * Approving a publish now preserves any previously scheduled time instead of clearing it, preventing invalid/empty schedule values.
  * Scheduling options (Keep/Now/Later) reliably reflect your current selection in the dialog.
  * Improved consistency between the Keep/Now/Later controls and the date-time picker so saved schedules match what you selected.

### DIFF
--- a/ui/app/src/components/PublishPackageReviewDialog/PublishingPackageReviewDialogContainer.tsx
+++ b/ui/app/src/components/PublishPackageReviewDialog/PublishingPackageReviewDialogContainer.tsx
@@ -193,7 +193,7 @@ export function PublishingPackageReviewDialogContainer(props: PublishingPackageR
 		if (state.action === 'approve') {
 			const data: PublishingPackageApproveParams = {
 				comment: state.approverComment,
-				schedule: state.scheduling === 'custom' ? state.schedule.toISOString() : null,
+				schedule: state.scheduling === 'custom' || state.scheduling === 'keep' ? state.schedule.toISOString() : null,
 				updateSchedule: true
 			};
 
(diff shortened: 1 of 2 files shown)


jvega190:
Fixed a couple of issues in the package review:
- Pre-selected `now` or `scheduled` selection
- Schedule not sent when selecting `keep`

rart (approved):

sumerjabri (changes requested):
@jvega190 @rart 

In `ui/app/src/components/PublishPackageReviewDialog/PublishingPackageReviewDialogContainer.tsx`

There is an issue reported by the rabbit -- START


191-199: Preserving schedule on 'keep' is correct; fix potential 'Invalid Date' source for custom scheduling

The conditional serialization looks good: sending null for "now" and ISO for "custom"/"keep" aligns with the desired behavior.

However, the initial schedule derivation uses:

schedule: new Date(publishingPackage?.schedule) ?? createAtLeastHalfHourInFutureDate()
If publishingPackage?.schedule is undefined, new Date(undefined) returns an Invalid Date object, which is not null/undefined, so the right-hand side of ?? is never used. This can surface as an invalid value when the user switches to "Later", or cause .toISOString() to throw on submit.

Recommend using a ternary to ensure a valid Date is always set:

```
useEffect(() => {
  setState({
    scheduling: publishingPackage?.schedule ? 'keep' : 'now',
    schedule: publishingPackage?.schedule
      ? new Date(publishingPackage.schedule)
      : createAtLeastHalfHourInFutureDate()
  });
}, [publishingPackage, setState]);
```

--END

jvega190:
Reviewed and fixed `Outside diff range comments`

rart (approved):

rart (approved):

sumerjabri (approved):

PR #4634, opened 2025-09-16
[4.x][8352] Focus search input when clicking the search button

jvega190:
https://github.com/craftercms/craftercms/issues/8352


ui/app/src/components/SearchBar/SearchBar.tsx
```diff
@@ -131,7 +131,7 @@ export function SearchBar(props: SearchBarProps) {
       className={cx(classes.search, focus && 'focus', showActionButton && 'noPadded', props.classes?.root)}
     >
       {showDecoratorIcon && onDecoratorButtonClick ? (
-        <IconButton onClick={onDecoratorButtonClick} size="large">
+        <IconButton onClick={() => onDecoratorButtonClick(inputRef.current)} size="large">
           <DecoratorIcon className={classes.searchIcon} />
         </IconButton>
       ) : (
```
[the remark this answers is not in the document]
jvega190: I works both ways

ui/app/src/components/SearchBar/SearchBar.tsx
```diff
@@ -131,7 +131,7 @@ export function SearchBar(props: SearchBarProps) {
       className={cx(classes.search, focus && 'focus', showActionButton && 'noPadded', props.classes?.root)}
     >
       {showDecoratorIcon && onDecoratorButtonClick ? (
-        <IconButton onClick={onDecoratorButtonClick} size="large">
+        <IconButton onClick={() => onDecoratorButtonClick(inputRef.current)} size="large">
```
rart: This is changing the signature (shouldn't), not documenting the change (CHANGELOG), and since we made the search bar visible don't see why it is still required?

rart (changes requested):

ui/app/src/components/SearchBar/SearchBar.tsx
jvega190: Per conversation, reverted and added the onClick action to select the input when clicking the decorator button

jvega190:
Outside diff range comment addresed in [here](https://github.com/craftercms/craftercms/issues/8358)

rart (commented):
Why are we adding autoFocus to the various panels as part of this work?

ui/app/src/components/SearchBar/SearchBar.tsx
```diff
@@ -122,7 +122,7 @@ export function SearchBar(props: SearchBarProps) {
   const [focus, setFocus] = useState(false);
   const { formatMessage } = useIntl();
   const finalPlaceholder = placeholder || formatMessage(messages.placeholder);
-  const inputRef = useRef<HTMLInputElement>(undefined);
+  const inputRef = useRef<HTMLInputElement | null>(undefined);
```
rart: There's no type consistency here. 

Either the type should be `Element | undefined`...
```suggestion
  const inputRef = useRef<HTMLInputElement | undefined>(undefined);
```

...or the ref should be initialised to `null` instead of `undefined`.
```suggestion
  const inputRef = useRef<HTMLInputElement | null>(null);
```

rart (changes requested):

jvega190:
> Why are we adding autoFocus to the various panels as part of this work?

Since this ticket is related to focus on input, I added the autoFocus on those panels for consistency with the others that already have it

rart (approved):

PR #4635, opened 2025-09-16
[4.x][8351] User search with [SA] breaks Studio UI

jvega190:
https://github.com/craftercms/craftercms/issues/8351


jvega190:

rart (approved):

PR #4638, opened 2025-09-16
[8352] Focus search input when clicking the search button

jvega190:
https://github.com/craftercms/craftercms/issues/8352

## Summary by CodeRabbit

* **New Features**
  - Search inputs now auto-focus when asset, page explorer, and search panels render so you can begin typing immediately.

* **Refactor**
  - Search button now directly focuses the input instead of toggling a collapsed state, providing a more consistent, streamlined search experience across the app.
  - Search bars are no longer collapsed and use consistent styling/width for predictable behavior.


russdanner (approved):

ui/app/src/components/SearchBar/SearchBar.tsx
```diff
@@ -55,7 +55,7 @@ export interface SearchBarProps {
 	onKeyPress?(key: string): void;
 	onKeyDown?: InputBaseProps['onKeyDown'];
 	onActionButtonClick?(e: React.MouseEvent<HTMLButtonElement, MouseEvent>, input: HTMLInputElement): void;
-	onDecoratorButtonClick?(): void;
+	onDecoratorButtonClick?(input: HTMLInputElement): void;
```
rart: Don't think this makes a great deal of sense from an API standpoint. Perhaps SearchBar exposes the input ref — if it doesn't already — and the consumer uses it that way.

rart (changes requested):

rart:
@coderabbitai review

ui/app/src/components/SearchBar/SearchBar.tsx
```diff
@@ -111,7 +110,7 @@ export function SearchBar(props: SearchBarProps) {
 		>
 			{showDecoratorIcon && onDecoratorButtonClick ? (
 				<IconButton
-					onClick={onDecoratorButtonClick}
+					onClick={() => onDecoratorButtonClick()}
```
rart: Why this change?

rart (changes requested):

ui/app/src/components/SearchBar/SearchBar.tsx
jvega190: Not needed, that was because I added a param that was later removed. Rolled back the change

jvega190:
@coderabbitai about the duplicate comment in here https://github.com/craftercms/studio-ui/pull/4638#pullrequestreview-4155921044, we're already handling the nullish input in `onActionButtonClick`. That's expected. Though we'll add the `null` in the onActionButtonClick callback type

rart (approved):

PR #4729, opened 2025-12-18
[4.x][8462] Changing page type retains original components, causing unintended component deletion

jvega190:
https://github.com/craftercms/craftercms/issues/8462

### DIFF
--- a/static-assets/components/cstudio-forms/forms-engine.js
+++ b/static-assets/components/cstudio-forms/forms-engine.js
@@ -3505,14 +3505,12 @@ const initializeCStudioForms = () => {
               'createdDate_dt',
               'lastModifiedDate',
               'lastModifiedDate_dt',
-              'components',
               'orderDefault_f',
               'placeInNav',
               'rteComponents'
             ],
             output = '',
             validFieldsStr,
-            fieldRe,
             section;
 
           // Add valid fields from the ones created dynamically by controls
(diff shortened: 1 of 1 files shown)


static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -3505,14 +3505,12 @@ const initializeCStudioForms = () => {
               'createdDate_dt',
               'lastModifiedDate',
               'lastModifiedDate_dt',
-              'components',
```
jvega190, line 3508: Per conversation with Russ, this `components` field is not part of the initial valid fields, since it's a legacy field we used for components inside RTEs. This was causing the stale `components` field to stay in the xml.

static-assets/components/cstudio-forms/forms-engine.js
```diff
@@ -3615,11 +3613,8 @@ const initializeCStudioForms = () => {
               attributes.pop();
             }
 
-            // Because we added start and end elements, we can be sure that any field names will
-            // be delimited by the delimiter token (ie. comma)
-            fieldRe = new RegExp(',' + key + '(?:,||d+|)', 'g');
-
-            if (fieldRe.test(validFieldsStr)) {
+            // Check if the current form field key (id) is in the validFields array
+            if (validFields.includes(key)) {
```
jvega190, line 3615: When checking for the valid field, we were validating with a regex but since we're checking against the form model, we can only check if the current field is in the validFields array.
There was an error in the regex causing that for example
`components` was valid because the regex watch matching a field like `components_o`, causing the field to end up incorrectly in the xml

jvega190:
Outside diff range comments addressed.

russdanner (commented):
I am trying to followe the comments on the individual changes  -- and I dont get it.

Why can't we just remove components from the list of fields to keep?

I don't want to make logic changes that will result in unknown behavior changes unless there's a great reason.

If this change is not fixing a major issue, ignore code rabbit and tell it you are fixing a specific issue.

rart (approved):

PR #4742, opened 2026-01-07
[8496] Implement archetypes from configuration

jvega190:
https://github.com/craftercms/craftercms/issues/8496

## Summary by CodeRabbit

* **New Features**
  * Archetypes now load dynamically from site config, offering translated archetype labels, grouped listings (including an "Other" group), and a skeleton while loading.

* **Refactor**
  * Content type creation and ID/validation flow unified to consistently derive prefixes and merge archetype descriptors when initializing types.

* **Chores**
  * New internal hooks and metadata enhancements surface archetypes across the UI and augment legacy config with name data.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


jvega190:
https://github.com/craftercms/craftercms/issues/8484 needs to be merged first

jvega190:

sumerjabri:
@rart testing

ui/app/src/models/GlobalState.ts
```diff
@@ -74,6 +74,7 @@ import { PackageDetailsDialogStateProps } from '../components';
 import { ViewPackagesDialogStateProps } from '../components/ViewPackagesDialog';
 import type { FolderMoveAlertDialogStateProps } from '../components/FolderMoveAlertDialog/FolderMoveAlertDialog';
 import type { PublishingStatus } from './Publishing';
+import type { ArcheType } from '../components/ContentTypeManagement/descriptors/archetypes';
```
rart:
```suggestion
import type { Archetype } from '../components/ContentTypeManagement/descriptors/archetypes';
```

ui/app/src/hooks/useArcheTypesList.ts
```diff
@@ -0,0 +1,24 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import useArcheTypes from './useArcheTypes';
+
+export function useArcheTypesList() {
+	const archeTypes = useArcheTypes();
```
rart:
```suggestion
	const archeTypes = useArchetypes();
```

ui/app/src/hooks/useArcheTypesList.ts
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import useArcheTypes from './useArcheTypes';
+
+export function useArcheTypesList() {
```
rart:
```suggestion
export function useArchetypesList() {
```

ui/app/src/components/ContentTypeManagement/components/CreateTypeDialog.tsx
```diff
@@ -229,4 +229,8 @@ function suggestTypeId(label: string): string {
 	return transformId(camelized);
 }
 
+function getPrefixForType(type: string): string {
```
rart:
```suggestion
function getTypeIdPathPrefixForType(type: string): string {
```

rart (changes requested):
See pending rabbit comments

ui/app/src/state/reducers/uiConfig.ts
```diff
@@ -104,14 +108,48 @@ const reducer = createReducer<GlobalState['uiConfig']>(initialState, (builder) =
...
+								dataSources: asArray(deserializedDescriptor.dataSources)
+							};
+
+							archetypes[tag.id] = {
+								...archetypes[tag.id],
+								name: deserializedDescriptor.name,
+								descriptor: extendArchetypeDescriptor(parentArchetype?.descriptor, deserializedDescriptor)
+							};
+						}
+					});
```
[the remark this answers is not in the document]
jvega190: It'll be documented that parent archetypes must be defined before children, and XML example will be like that

jvega190:

ui/app/src/components/ContentTypeManagement/components/CreateTypeDialog.tsx
```diff
@@ -148,13 +152,11 @@ function CreateTypeDialogBody(props: CreateTypeDialogBaseProps) {
...
-						</MenuItem>
-						<MenuItem value="component">
-							<FormattedMessage defaultMessage="Component" />
-						</MenuItem>
-						{/* Post v5 TODO: List archetypes from config */}
+						{archetypes?.map((archetype) => (
+							<MenuItem key={archetype.id} value={archetype.id}>
+								{getPossibleTranslation(archetype.name, formatMessage)}
+							</MenuItem>
+						))}
```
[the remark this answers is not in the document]
jvega190: archetypes are always loaded before the dialog is open

ui/app/src/state/reducers/uiConfig.ts
```diff
@@ -104,14 +108,49 @@ const reducer = createReducer<GlobalState['uiConfig']>(initialState, (builder) =
...
 
+				configDOM
+					.querySelectorAll('[id="craftercms.components.ContentTypeManagement"] > configuration > objectTypes')
+					.forEach((tag) => {
+						const descriptor = tag.querySelector('descriptor');
+						// Parent archetypes must be defined before children so that they can be extended properly
+						const extendsFrom = tag.getAttribute('extends');
+						let parentArchetype = null;
+						if (extendsFrom) {
+							parentArchetype = archetypes[extendsFrom] ?? {};
```
rart: Don't think we need to initialise this this early. Should be done for typebuiler and possibly for FE.

rart (changes requested):

ui/app/src/state/reducers/uiConfig.ts
jvega190: I'm moving this to a separate action to retrieve the archetypes from the xml when using the `useArchetypes` hook. Though when the ui.xml is edited, `fetchUiConfig` gets dispatched, so if there was an update to the archetypes they get updated in the state.
By calling a separate action then I have to call the new action when the ui.xml is updated. Is that ok?

jvega190: There's a detail here, the uiConfig updates when the ui.xml is saved in configuration (also with websocket), so by having the archetypes calculation in the fetchUiConfigComplete action keeps it up to date. I created an action to get the archetypes from the ui config and I thought about a couple of options.
- Dispatch the action whenever is needed using on mount so it gets the latest every time, but it won't update when the websocket event gets triggered.
- Dispatch the action in the `useArchetypes` hook. But once the archetypes are loaded they won't update unless I dispatch it also when the ui config websocket event triggers.

What do you think we should do to keep the archetypes up to date?

rart: Would it work to initialise where it is needed (not many places), and re-init on the events that update the config if that piece of state is already initialised? I'm not sure we even need it in redux, could be component state if it makes sense.

jvega190: I removed the archetypes from redux, calculating them in the useArchetypes hook, so it'll only be initialized when needed.

ui/app/src/hooks/useArchetypes.ts
```diff
@@ -0,0 +1,74 @@
...
+import { extendArchetypeDescriptor } from '../utils/object';
+import { useEffect, useState } from 'react';
+import { Archetype } from '../components/ContentTypeManagement/descriptors/archetypes';
+
+export function useArchetypes() {
+	const [archetypes, setArchetypes] = useState<LookupTable<Archetype>>({});
+	const uiConfigXml = useSelection((state) => state.uiConfig.xml);
+
+	useEffect(() => {
+		if (uiConfigXml) {
```
rart, line 30: If not loaded, and not loading, it should dispatch a fetch request.

rart (approved):
Approved for merge. 

@jvega190 please send enhancements separately.

PR #4750, opened 2026-01-15
[8507] FE2 - change content type functionality

jvega190:
Ref https://github.com/craftercms/craftercms/issues/8507

## Summary by CodeRabbit

* **Improvements**
  * You can now specify an alternative content type/template when updating existing items, allowing flexible reassignment during edits.
  * Update flows accept an optional override to resolve the form against a different content type/template.
  * When an alternative content type/template is applied during an edit, the form flow will mark the edit as having pending changes for clearer dialog behavior.


jvega190:

ui/app/src/components/FormsEngine/lib/formUtils.tsx
```diff
@@ -381,13 +381,15 @@ export function fetchUpdateRequirements({
 	path,
 	modelId,
 	readonly,
-	contentTypesById
+	contentTypesById,
+	changeTypeId
```
rart, line 385: What's the flow here? Why would the changeTypeId come at this time?

rart (commented):
@russdanner should we have a UX for changing type where the user reviews all the data that would be discarded before proceeding to open the form?

Change Type => Change Type Review Dialog => Accept? => Execute type change => Open form with type changed

It would also clean up the FE implementation by removing the need to handle a type change prop and all the logic this involves.

russdanner:
> @russdanner should we have a UX for changing type where the user reviews all the data that would be discarded before proceeding to open the form?
> 
> Change Type => Change Type Review Dialog => Accept? => Execute type change => Open form with type changed
> 
> It would also clean up the FE implementation by removing the need to handle a type change prop and all the logic this involves.

I love it, can we connect on the effort required?  If it help make things easier (shortens other tasls) I am open to doing it in 5.0, otherwise, for the sake of scope may need to move it out a bit.

ui/app/src/components/FormsEngine/FormsEngine.tsx
```diff
@@ -154,6 +155,7 @@ export interface UpdateModeProps {
 		path: string;
 		modelId?: string;
 		values?: LookupTable<unknown>;
+		changeTypeId?: string;
```
rart: Add nice and descriptive comment to this prop

rart (changes requested):

ui/app/src/components/FormsEngine/lib/formUtils.tsx
jvega190: Here in `fetchUpdateRequirements` we fetch the item based on the path, and get the contentType from the `contentTypesById` lookup (`contentTypesById[item.contentTypeId]`). So the `changeTypeId` prop is needed to get the new content type, instead of the item one.

ui/app/src/components/FormsEngine/FormsEngine.tsx
```diff
@@ -614,6 +617,12 @@ function FormOrchestrator(props: FormsEngineProps) {
 	const tableOfContents = <TableOfContents fieldsToRender={fieldsToRender} containerRef={containerRef} />;
 	const effectRefs = useUpdateRefs({ fieldsToRender, versionCommentAtom: stableFormContext.atoms.versionComment });
 
+	useMount(() => {
+		if (update?.changeTypeId) {
+			setHasPendingChanges(true);
+		}
+	});
```
rart, line 640: Explain the reason for doing this (add comment)

rart (changes requested):

ui/app/src/components/FormsEngine/FormsEngine.tsx
```diff
@@ -614,6 +617,14 @@ function FormOrchestrator(props: FormsEngineProps) {
 	const tableOfContents = <TableOfContents fieldsToRender={fieldsToRender} containerRef={containerRef} />;
 	const effectRefs = useUpdateRefs({ fieldsToRender, versionCommentAtom: stableFormContext.atoms.versionComment });
 
+	useMount(() => {
+		// If 'update.changeTypeId' has content, it means the content type has changed, so we set pending changes to true
+		// to be able to save the content (enable save button).
```
rart:
```suggestion
		// to be able to enable the save button and allow users to save immediately if that's all they want to do.
```

rart (approved):

sumerjabri:
@jvega190 please fix merge conflicts.

jvega190:
@sumerjabri conflicts fixed

PR #4766, opened 2026-01-27
[8483] Avoid collapsing header if resizing is bigger than scroll height

jvega190:
There is this issue when if after collapsing the header, if the scroll goes away (because the form content gets bigger) it causes the intersectionObserver to recalculate and make the header expand again, entering a loop. So this fix checks if the scrollHeight is 100px larger than the clientHeight. If not, then header collapse is not allowed.

https://github.com/craftercms/craftercms/issues/8483

## Summary by CodeRabbit

* **Bug Fixes**
  * Header collapse now only activates after initial content size is measured, preventing premature collapsing in compact layouts.
  * Collapse behavior is now guarded so it only toggles when appropriate, avoiding unexpected header changes.
  * Improved observer lifecycle and cleanup to prevent duplicate observers and ensure stable, predictable header behavior.


ui/app/src/components/FormsEngine/FormsEngine.tsx
```diff
@@ -778,20 +779,30 @@ function FormOrchestrator(props: FormsEngineProps) {
 	};
 
 	const [mainContent, setMainContent] = useState(null);
+	const [collapseHeaderAllowed, setCollapseHeaderAllowed] = useState(false);
```
rart: Is this something the calls for a rerender? Smells like it could simply be a ref.

rart (approved):

PR #4804, opened 2026-02-16
[8423] Add repeat group validator to check properties and fields of items

jvega190:
https://github.com/craftercms/craftercms/issues/8423

## Summary by CodeRabbit

* **New Features**
  * Added validation for repeating form fields with enforcement of minimum/maximum occurrence constraints.
  * Per-item validation across repeat group entries, aggregating results and surfacing localized messages for occurrence violations to improve form feedback and data integrity.


rart (commented):
The validators can return multiple messages. Are we generally stopping after the first error? Are the validation errors able to render multiple errors for a single control?

jvega190:
@rart multiple error can render for a single control (is supported and other validators do return multiple errors). Updated the validator to avoid returning after first error.

rart (approved):

PR #4824, opened 2026-03-13
[8569] Update staging publishing target indicator in UI

jvega190:
https://github.com/craftercms/craftercms/issues/8569

## Summary by CodeRabbit

* **Bug Fixes**
  * Item display logic updated so items in "staged" state correctly show the staged icon instead of the workflow icon, and staged items that are live-submitted or scheduled no longer incorrectly display staging.
* **Documentation**
  * Added internal helper documentation clarifying staged-display priority and exclusions.

### DIFF
--- a/ui/app/src/components/ItemDisplay/ItemDisplay.tsx
+++ b/ui/app/src/components/ItemDisplay/ItemDisplay.tsx
@@ -101,7 +101,7 @@ const ItemDisplay = forwardRef<HTMLSpanElement, ItemDisplayProps>((props, ref) =
 			}}
 		>
 			{/* @see https://github.com/craftercms/craftercms/issues/5442 */}
-			{inWorkflow && !shouldItemShowAsStaged(item as ContentItem)
+			{inWorkflow && !shouldItemShowAsStaged(item)
 				? showWorkflowState && (
 						<ItemStateIcon
 							{...stateIconProps}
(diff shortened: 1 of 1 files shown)


ui/app/src/components/ItemDisplay/ItemDisplay.tsx
```diff
@@ -159,4 +160,23 @@ const ItemDisplay = forwardRef<HTMLSpanElement, ItemDisplayProps>((props, ref) =
...
+/**
+ * Determines if the item's icon should be displayed as staged.
+ *
+ * @param item - The content item to check.
+ * @returns True if the item should be displayed as staged, false otherwise.
+ *
+ * Staging has priority over modified and null. Additionally, if an item is submitted to live, submitted to staging, or
+ * scheduled, it should not be shown as staged.
+ */
+function shouldItemShowAsStaged(item: ContentItem): boolean {
```
rart: Type the argument so that you don't have to cast the various complying interfaces...
```suggestion
function shouldItemShowAsStaged(item: { stateMap?: Record<'staged' | 'new' | ..., boolean> }): boolean {
```

rart (changes requested):

ui/app/src/components/ItemDisplay/ItemDisplay.tsx
jvega190: since `LightItem` doesn't have stateMap, it still complains that `LightItem doesn't have an optional param stateMap`  so I end up needing the cast.
I set `item: ContentItem | LightItem` and validated `stateMap` prop inside the util. That way I don't have to do any casting. What do you think?

rart (approved):

PR #4833, opened 2026-03-25
[8407] set 'includeChildren' to true if mainItems are all folders

jvega190:
https://github.com/craftercms/craftercms/issues/8407

## Summary by CodeRabbit

* **Bug Fixes**
  * **Publish Dialog**: The "include children" option now auto-enables when all selected items are folders and its state is correctly synchronized with the publish flow, ensuring consistent behavior and more intuitive defaults during publishing.


ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -138,11 +137,21 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
 	const arePublishingItemsFolders = useMemo(() => {
 		const allItems = [...mainItems, ...childrenItems];
 		return allItems.length > 0 && allItems.every((item) => item.systemType === 'folder');
 	}, [mainItems, childrenItems]);
 
+	useEffect(() => {
+		if (areMainItemsFolders && !effectRefs.current.includeChildren) {
+			setIncludeChildren(true);
+		}
+	}, [areMainItemsFolders, effectRefs]);
```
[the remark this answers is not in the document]
jvega190: That's correct, the idea is for the user to be able to set includeChildren back to false

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
```diff
@@ -138,11 +137,21 @@ export function PublishDialogContainer(props: PublishDialogContainerProps) {
...
+		return mainItems.length > 0 && mainItems.every((item) => item.systemType === 'folder');
+	}, [mainItems]);
 	const arePublishingItemsFolders = useMemo(() => {
 		const allItems = [...mainItems, ...childrenItems];
 		return allItems.length > 0 && allItems.every((item) => item.systemType === 'folder');
 	}, [mainItems, childrenItems]);
 
+	useEffect(() => {
+		if (areMainItemsFolders && !effectRefs.current.includeChildren) {
+			setIncludeChildren(true);
```
rart: Please double-check if this really needs an effect or even a reactful state? Could it be a computed const instead? e.g.
```ts
const includeChildren = areMainItemsFolders && !effectRefs.current.includeChildren
```
or similar...

rart (changes requested):

ui/app/src/components/PublishDialog/PublishDialogContainer.tsx
jvega190: I removed the effect that calls `setIncludeChildren(true)` and set the initial `includeChildren` state to true if all the mainItems are folders (the logic I was setting was for includeChildren to be true if all initial items are folders). I still need `includeChildren` to be a state since there's a checkbox to change the state

rart (approved):

PR #4834, opened 2026-03-25
[8419] Improve Change Type dialog/flow

jvega190:
https://github.com/craftercms/craftercms/issues/8419

- Update to show change type confirmation after selecting the new type, instead of when opening the dialog
- Add cancel button to dialog
- Do not show current type (show empty state if needed)

## Summary by CodeRabbit

* **New Features**
  * Selecting a new content type now opens a confirmation dialog before the change is applied.
  * Added a Cancel button to the Change Content Type dialog to dismiss it.

* **Improvements**
  * The item's current content type is excluded from the selectable list.
  * Improved dialog text localization and better handling of loading/empty states.


jvega190:
@coderabbitai review

rart (changes requested):
See outside diff range comment

jvega190:
@rart the outside diff range comment doesn't apply, since the dialogs system handles the onClose if not provided.

rart (approved):

PR #4837, opened 2026-03-27
[8193] Limit min/max size in nodeselector descriptor

jvega190:
https://github.com/craftercms/craftercms/issues/8193

## Summary by CodeRabbit

* **New Features**
  * Implemented stricter validation rules for node selector configuration, adding minimum value constraints to size fields to ensure valid specifications and prevent configuration errors.

* **Refactor**
  * Extended the validation framework to support additional validation key sources, enabling more flexible and comprehensive constraint definitions for field validations across the platform.

### DIFF
--- a/ui/app/src/components/ContentTypeManagement/descriptors/controls/nodeSelector.ts
+++ b/ui/app/src/components/ContentTypeManagement/descriptors/controls/nodeSelector.ts
@@ -40,14 +40,18 @@ export const nodeSelectorDescriptor: DescriptorContentType = {
 			type: 'int',
 			name: defineMessage({ defaultMessage: 'Minimum Size' }),
 			defaultValue: undefined,
-			validations: immutableEmptyObject
+			validations: {
+				minValue: createValidation('minValue', 0)
+			}
 		},
 		maxSize: {
 			id: 'maxSize',
 			type: 'int',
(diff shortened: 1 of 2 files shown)


ui/app/src/components/ContentTypeManagement/descriptors/controls/nodeSelector.ts
```diff
@@ -40,14 +40,18 @@ export const nodeSelectorDescriptor: DescriptorContentType = {
...
+			}
 		},
 		maxSize: {
 			id: 'maxSize',
 			type: 'int',
 			name: defineMessage({ defaultMessage: 'Maximum Size' }),
 			defaultValue: undefined,
-			validations: immutableEmptyObject
+			validations: {
+				minValue: createValidation('minValue', 1)
```
rart, line 53: Why would we start with a max of 1? Max should be null, I'd think.

rart (changes requested):

ui/app/src/components/ContentTypeManagement/descriptors/controls/nodeSelector.ts
jvega190: The default value is null, this is just the validation so if setting a value, the min allowed would be 1 (to prevent setting 0 or negative)

rart (approved):

PR #4839, opened 2026-04-08
[7830] "Allow Duplicates" option for node-selector not properly honored by Experience Builder drag and drop

jvega190:
https://github.com/craftercms/craftercms/issues/7830

## Bug Fixes
- Enforced duplicate-item validation across selection workflows so items cannot be duplicated when disallowed.
- Prevented errors when updating item hierarchies by guarding against missing or undefined entries.
- Adjusted preselection behavior so initial selections respect the allow-duplicates setting, avoiding unintended preselected items.


jvega190:

rart (approved):

PR #4846, opened 2026-04-21
[8564] Update Empty Dropzone Text

jvega190:
https://github.com/craftercms/craftercms/issues/8564

## Summary by CodeRabbit

* **Improvements**
  * Empty collection areas now show clearer, localized instructional prompts for dragging or adding content.

* **Localization**
  * Overlay messages (hidden content indicator, turn-off-edit prompt, empty-target instruction, click-to-add prompt) are localized and update when the active language changes.

### DIFF
--- a/ui/guest/src/react/GuestGlobalStyles.tsx
+++ b/ui/guest/src/react/GuestGlobalStyles.tsx
@@ -35,6 +35,7 @@ import { Theme } from '@mui/material/styles';
 import { deepmerge } from '@mui/utils';
 import { DeepPartial } from '@craftercms/studio-ui/models/DeepPartial';
 import { typographyStyles } from '@craftercms/studio-ui/components/GlobalStyles/typography';
+import { useIntl } from 'react-intl';
 
 export type GlobalStyleKeys =
 	| 'craftercms-asset-uploader-mask-container'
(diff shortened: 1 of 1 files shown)


ui/guest/src/react/GuestGlobalStyles.tsx
```diff
@@ -181,7 +181,7 @@ export function GuestGlobalStyles(props: GuestGlobalStylesProps) {
 										color: '#fff',
 										display: 'inline-block',
 										padding: '10px',
-										content: '"No items on this area."',
+										content: '"Empty component dropzone. Drag new or existing content items here to add them"',
```
rart: @russdanner this text is a bit of an issue when the drop zone doesn't support existing or new content.

We also use the term "drop targets", so don't think we should be using dropzone.

rart (changes requested):

ui/guest/src/react/GuestGlobalStyles.tsx
rart:
```suggestion										

content: '"Empty components target. Drag items here to add."',
```

rart: Also, translations?

jvega190:
Outside diff range comment addressed

russdanner:
@roy the text may be over perscritive in some scenarios but it tells them where to look on the right bar (create or existing) I'm ok with that. I think finding nothing in the right panel to drop or a rejection from the target area is ok.

Adjust language to target instead of "zone"

rart (approved):

PR #4854, opened 2026-05-07
[8667] Handle activity with type PUBLISH in the dashboard activities dashlets

jvega190:
https://github.com/craftercms/craftercms/issues/8667

### DIFF
--- a/ui/app/src/components/ActivityDashlet/utils.tsx
+++ b/ui/app/src/components/ActivityDashlet/utils.tsx
@@ -131,7 +131,7 @@ export function renderActivity(
 					values={{ render_package_link }}
 				/>
 			);
-		case 'APPROVE':
+		case 'PUBLISH':
 			return (
 				<FormattedMessage
 					id="activityDashlet.publishedActivityMessage"
@@ -204,7 +204,7 @@ export const activityNameLookup: Record<Activities | 'ALL', any> = {
 	DELETE: <FormattedMessage id="words.delete" defaultMessage="Delete" />,
 	INITIAL_PUBLISH: <FormattedMessage id="operations.initialPublish" defaultMessage="Initial Publish" />,
 	MOVE: <FormattedMessage id="words.move" defaultMessage="Move" />,
-	APPROVE: <FormattedMessage id="words.approve" defaultMessage="Approve" />,
+	PUBLISH: <FormattedMessage id="words.approve" defaultMessage="Approve" />,
 	PUBLISHED: <FormattedMessage id="words.publish" defaultMessage="Publish" />,
 	REJECT_PUBLISH_PACKAGE: <FormattedMessage id="words.reject" defaultMessage="Reject" />,
 	REQUEST_PUBLISH: <FormattedMessage id="operations.requestPublish" defaultMessage="Request Publish" />,
--- a/ui/app/src/components/ActivityDashlet/utils.tsx
+++ b/ui/app/src/components/ActivityDashlet/utils.tsx
@@ -134,7 +134,7 @@ export function renderActivity(
 		case 'PUBLISH':
 			return (
 				<FormattedMessage
-					id="activityDashlet.publishedActivityMessage"
+					id="activityDashlet.publishActivityMessage"
 					defaultMessage="Approved <render_package_link>a package</render_package_link>"
 					values={{ render_package_link }}
 				/>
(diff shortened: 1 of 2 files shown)


jvega190:

jvega190:
@coderabbitai review

sumerjabri:
@coderabbitai this PR is now ready to send to a human, before we do that, I think you should have a final look.

jvega190:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> Let me pull up the actual code to do a thorough final look.
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Get the full diff of the PR
> gh pr diff 4854 --repo craftercms/studio-ui
> 
> echo "---FILE CONTENT: utils.tsx---"
> fd "utils.tsx" --full-path | xargs grep -l "ActivityDashlet" | head -5
> 
> echo "---FILE CONTENT: Activity.ts---"
> fd "Activity.ts" | head -5
> ```
> 
> Length of output: 2152
> 
> Good, let me dig deeper into the i18n definitions and any remaining `APPROVE` references.
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> echo "=== Search for old message ID ==="
> rg "publishedActivityMessage" --type tsx --type ts --type js -n
> 
> echo "=== Search for new message ID ==="
> rg "publishActivityMessage" -n
> 
> echo "=== Search for APPROVE activity references (non-test) ==="
> rg "'APPROVE'" -n -g '!*.test.*' -g '!*.spec.*'
> 
> echo "=== Full renderActivity switch context ==="
> fd "utils.tsx" | xargs grep -l "renderActivity" | while read f; do
>   echo "FILE: $f"
>   cat "$f"
> done
> 
> echo "=== Check activityNameLookup full map ==="
> rg "activityNameLookup" -n -A5 -B5
> 
> echo "=== Check i18n/messages files for activityDashlet keys ==="
> rg "activityDashlet\." -n --type ts --type tsx --type js -l
> 
> echo "=== Find messages/intl definition files ==="
> fd -e ts -e tsx -e js "messages" | head -20
> ```
> 
> Length of output: 14038
> 
> `@sumerjabri` Here is my final look at PR `#4854` before it goes to a human reviewer. The changes are internally consistent and correctly address the stated goal. A few observations:
> 
> ### ✅ What looks good
> 1. **Cleanup is complete** — no remaining references to `'APPROVE'` anywhere in the codebase after this PR.
> 2. **Type consistency** — `Activity.ts` union type and `activityNameLookup` are now in sync; both drop `APPROVE` and add `PUBLISH`.
> 3. **New message ID is unique** — `activityDashlet.publishActivityMessage` is scoped only to the `PUBLISH` case, avoiding the ID collision that previously existed on the old `APPROVE` case.
> 
> ### ⚠️ Points to verify before human review
> **1. `defaultMessage` wording vs. action type name (minor)**
> 
> In `renderActivity`, the `PUBLISH` case reads:
> 
> ```tsx
> case 'PUBLISH':
>   return (
>     <FormattedMessage
>       id="activityDashlet.publishActivityMessage"
>       defaultMessage="Approved <render_package_link>a package</render_package_link>"
>       ...
>     />
>   );
> ```
> 
> The `defaultMessage` says **"Approved"** even though the action type is now `PUBLISH`. This appears intentional based on the removed TODO comment (the backend sends `PUBLISH` for what was previously `APPROVE`/`APPROVE_SCHEDULED`), but the reviewer should confirm the UX copy is still the right wording.
> 
> **2. Filter label for `PUBLISH` in the dropdown**
> 
> ```tsx
> PUBLISH: <FormattedMessage id="words.approve" defaultMessage="Approve" />,
> ```
> 
> Users will see **"Approve"** in the activity-type filter dropdown for the `PUBLISH` event type. This is consistent with point 1 above, but worth a UX sign-off to ensure "Approve" is the desired label for this filter option.
> 
> ### 🐛 Pre-existing bug (out of scope, but worth tracking)
> `REJECT_PUBLISH_PACKAGE` and `PUBLISHED` both share the same message ID:
> 
> ```tsx
> // REJECT_PUBLISH_PACKAGE case:
> id="activityDashlet.publishedActivityMessage"  // defaultMessage: "Rejected a package"
> 
> // PUBLISHED case:
> id="activityDashlet.publishedActivityMessage"  // defaultMessage: "Published a package"
> ```
> 
> If this ID is ever translated in an external i18n system, both cases would render the same localized string. This was not introduced by this PR — it's pre-existing — but it's worth a follow-up issue.
> 
> Would you like me to open a GitHub issue to track that pre-existing message ID collision?

@coderabbit the points 1 and 2 are expected, this is the behavior we want. I fixed the duplicate messages ids

rart (approved):

PR #4870, opened 2026-05-28
[4.x][8257] Bypassing required fields with Save and Minimize or Save without Closing

jvega190:
Per conversation, since `Save & Minimize` action does a save as draft and then minimizes, we updated the label to be `Save Draft & Minimize` 

https://github.com/craftercms/craftercms/issues/8257


jvega190:

rart (approved):

PR #4874, opened 2026-06-02
[7830] Enhance validation messaging and handle duplicate instances in drag-and-drop functionality

jvega190:
https://github.com/craftercms/craftercms/issues/7830

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate detection to prevent items from being dropped into zones where they already exist, with localized error messaging.

* **Improvements**
  * Enhanced validation message handling to display more contextual feedback when operations are restricted.

### DIFF
--- a/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
+++ b/ui/app/src/components/PreviewConcierge/PreviewConcierge.tsx
@@ -1038,7 +1038,9 @@ export function PreviewConcierge(props: PropsWithChildren<{}>) {
 				}
 				case snackGuestMessage.type: {
 					enqueueSnackbar(
-						payload.id in guestMessages ? formatMessage(guestMessages[payload.id], payload.values ?? {}) : payload.id,
+						payload.id in guestMessages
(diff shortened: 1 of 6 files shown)


rart (commented):
How are you handling duplicates in DnD/XB?

jvega190:
> How are you handling duplicates in DnD/XB?

The actual logic in XB of checking if zone supports duplicates or not was already done in [here](https://github.com/craftercms/studio-ui/pull/4839). Where we get the `allowDuplicates` validation for the field, and consider it when calculating the drop targets

For this PR, what I'm doing is when calculating the dragContext, I check if the instanceId already exists in any of the dropzones, if so, then I label it as duplicate in a zone (isInstanceDuplicateInZone).
Then, the idea in this PR is to enhance the message when trying to add a duplicate item and no dropzone allows duplicate items, so in the epic, when checking for the dragContext dropzones, if there are no dropzones availables, and the `isInstanceDuplicateInZone` is true, show the `no duplicates allowed` snack, instead of the generic `no drop targets found...`

rart (approved):

PR #4875, opened 2026-06-02
[8564] Add translations

jvega190:
Per conversation with Russ, I added the translations for this texts

https://github.com/craftercms/craftercms/issues/8564


jvega190:

PR #4876, opened 2026-06-02
[8708] Improve Insert Control dialog usability

jvega190:
- Move system types under the other fields types.
- Don't show system fields that are already used (you can't have multiples of these)
- Move UUID (a generic control) to the generic list of controls.

https://github.com/craftercms/craftercms/issues/8708

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate field type assignments and conflicting file-name/auto-filename combinations in the field picker.
  * Adjusted the field selection layout so filtered fields consistently appear before system fields for clearer choices.
  * Removed UUID from the list of available control types.


jvega190:
Outside diff range comment addressed

jvega190:

jvega190:

rart (approved):

PR #4886, opened 2026-06-08
[4.x][8220] Accessibility: Consider keyboard shortcut disabling

jvega190:
https://github.com/craftercms/craftercms/issues/8220


jvega190:

rart (commented):
Same comment as develop PR. Please confirm.

jvega190:
@rart checked in here and I didn't see any input being affected by the shortcuts dialog event. Per discussion dialogs like WidgetDialog are not affected so we'll not include them

rart (approved):

PR #4890, opened 2026-06-12
[8723] Handle errors when trying to paste an item in an invalid location for its content type

jvega190:
The error handler was correct, but the block/unblock UI actions were added to the new dialogs system, so the showing/removal wasn't working as expected in this scenario (and probably others).

Added error handling for this scenario, for the error message to be clear to the user.

This PR https://github.com/craftercms/studio/pull/3956 is needed for the error code handling to work properly.

https://github.com/craftercms/craftercms/issues/8723

## Summary by CodeRabbit

* **Style**
  * Improved how error messages wrap by adjusting text-breaking behavior for longer or unbroken content.

* **Bug Fixes**
  * Enhanced clipboard error handling by displaying a more specific, destination-clarifying message for a particular error code, with an improved fallback for other errors.

* **Refactor**
  * Removed the UI-blocking dialog flow and its related dialog-mapping/state handling to streamline dialog management.

### DIFF
--- a/ui/app/src/components/ErrorState/ErrorState.tsx
+++ b/ui/app/src/components/ErrorState/ErrorState.tsx
@@ -93,7 +93,7 @@ export function ErrorState(props: ErrorStateProps) {
 					sx={{
 						textAlign: 'center',
 						marginBottom: (theme) => theme.spacing(1),
-						wordBreak: 'break-all',
+						wordBreak: 'break-word',
 						...sxs?.message
 					}}
 					children={message}
--- a/ui/app/src/state/epics/mappedDialogs.ts
+++ b/ui/app/src/state/epics/mappedDialogs.ts
@@ -131,7 +131,6 @@ const dialogsMap = {
 	[showRenameAssetDialog.type]: 'craftercms.components.RenameAssetDialog',
 	[showDeleteDialog.type]: 'craftercms.components.DeleteDialog',
 	[showEditDialog.type]: 'craftercms.components.LegacyFormDialog',
-	[blockUI.type]: 'craftercms.components.UIBlocker',
 	[showFolderMoveAlertDialog.type]: 'craftercms.components.FolderMoveAlertDialog'
 };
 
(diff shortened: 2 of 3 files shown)


jvega190:

jvega190:

ui/app/src/state/epics/mappedDialogs.ts
```diff
@@ -274,29 +272,6 @@ const showDialogsEpics: CrafterCMSEpic[] = [
...
-		),
-	(action$, state$) =>
-		action$.pipe(
-			ofType(unblockUI.type),
-			withLatestFrom(state$),
-			map(() => {
-				return popDialog({ id: blockUI.type });
-			})
-		)
-	// end region
```
rart, line 299: What's this got to do with paste?
Why is this being removed?
Where is it handled if not here?

rart (changes requested):

ui/app/src/state/epics/mappedDialogs.ts
jvega190: The actions were added to the dialogs system, but we left some existing dialogs in the GlobalDialogManager (ItemMegaMenu, AuthMonitor, Laucher, etc) including the UI Blocker. So it was duplicated

jvega190:

ui/app/src/components/ErrorState/ErrorState.tsx
```diff
@@ -93,7 +93,7 @@ export function ErrorState(props: ErrorStateProps) {
 					sx={{
 						textAlign: 'center',
 						marginBottom: (theme) => theme.spacing(1),
-						wordBreak: 'break-all',
+						wordBreak: 'break-word',
```
rart, line 96: This style might have been in place for super long exceptions. Why change?

ui/app/src/state/epics/mappedDialogs.ts
rart: Haven't gotten an answer to where the UI blocker via dispatch is handled

rart (changes requested):

ui/app/src/state/epics/mappedDialogs.ts
jvega190: The UIBlocker is still in the GlobalDialogManager. When dispatching the action there's a uiBlocker reducer that updates the state (opens the blocker), and the component in the GlobalDialogManager gets the updated state and shows/hides according to the action dispatched

ui/app/src/components/ErrorState/ErrorState.tsx
jvega190: `break-all` causes it to break at any point of the text, causing it to look like this:
Screenshot from 2026-06-25 12-06-49
by setting it to `break-word`, it won't split words, like this:
Screenshot from 2026-06-25 12-05-59

rart (approved):

PR #4908, opened 2026-07-06
[8397] Remove viewPackages action from packageItems default set of actions

jvega190:
Validate view packages action against actionsToInclude (actionsToInclude is set to 'all actions' if no override is set)

https://github.com/craftercms/craftercms/issues/8397

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the “View packages” context-menu option could appear even when it was explicitly excluded by the available action filter.
  * The menu now respects the selected action set more consistently.


jvega190:

rart (approved):

PR #4909, opened 2026-07-06
[8454] Improve the "More Items" alignment in the side bar 

jvega190:
Updated the 'N more items' entry of the tree to be a `TreeItem` instead of a `Box` so it gets indented properly with the rest of items.

https://github.com/craftercms/craftercms/issues/8454

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the “more” entry in the path navigator tree so it behaves like a proper tree item, making it easier to interact with and navigate.
  * Updated click handling to avoid unintended tree selection or expansion when loading additional items.
  * Refined tree item spacing, hover, focus, and label styling for a more consistent and polished sidebar experience.


jvega190:

jvega190:
@coderabbit you're right, since we have disableSelection then it is a pre-existing limitation, that's how we're handling it right now. Considering that, please give it a final review before I hand over for human review

rart (approved):

PR #4917, opened 2026-07-10
[8755] Fix double scroll bar on Publishing Package Details

jvega190:
https://github.com/craftercms/craftercms/issues/8755

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the sizing behavior of the virtualized package items list container so it lays out correctly and avoids incorrect scrollbar height.
  - Updated the package items view’s scroll handling to prevent vertical scrolling in that region, reducing duplicate/undesired scrollbars.
  - Enabled vertical scrolling for the package items tree within its bounded area while preserving consistent scrolling for the surrounding layout.

### DIFF
--- a/ui/app/src/components/PackageItems/PackageItems.tsx
+++ b/ui/app/src/components/PackageItems/PackageItems.tsx
@@ -174,7 +174,7 @@ export function PackageItems(props: PackageItemsProps) {
 						minHeight: 420,
 						height: 'calc(60vh)',
 						maxHeight: 600,
-						overflowY: !disableTreeView && isTreeView ? 'auto' : 'hidden'
+						overflowY: 'hidden'
 					}}
 				>
 					{state.items &&
--- a/ui/app/src/components/PackageItems/PackageItems.tsx
+++ b/ui/app/src/components/PackageItems/PackageItems.tsx
@@ -174,7 +174,7 @@ export function PackageItems(props: PackageItemsProps) {
 						minHeight: 420,
 						height: 'calc(60vh)',
 						maxHeight: 600,
-						overflowY: 'auto'
+						overflowY: !disableTreeView && isTreeView ? 'auto' : 'hidden'
 					}}
 				>
 					{state.items &&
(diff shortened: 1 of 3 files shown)


jvega190:

ui/app/src/components/PackageItems/PackageItems.tsx
```diff
@@ -174,7 +174,7 @@ export function PackageItems(props: PackageItemsProps) {
 						minHeight: 420,
 						height: 'calc(60vh)',
 						maxHeight: 600,
-						overflowY: 'auto'
+						overflowY: !disableTreeView && isTreeView ? 'auto' : 'hidden'
```
jvega190: I updated this for the scroll to be handled by this element when in tree view mode. For the list view mode, the scroll is handled by the react-window/infinite-loader library.

ui/app/src/components/PackageItems/PackageItemsList.tsx
```diff
@@ -53,7 +53,7 @@ export function PackageItemsList(props: PackageItemsListProps) {
 	});
 
 	return (
-		<Box sx={{ flex: 1 }}>
+		<Box sx={{ flex: 1, minHeight: 0 }}>
```
jvega190, line 56: This is needed for virtualization in a flex column, that way the scrollbar for the react-window list works properly.

ui/app/src/components/PackageItems/PackageItems.tsx
rart: Why not always? Why the duality?

jvega190: The tree view container is this paper, so that's why the scroll is being handled in here. For the list view, react-window sets the containers/etc, so that's why I left the scroll handling in here for the tree. Would you like me to add a container + scroll handling in the PackageItemsTree component, so this component gets cleaned (of scroll handling)?

jvega190: Updated the tree view component to handle the scroll, the container is not handling it anymore

ui/app/src/components/PackageItems/PackageItemsTree.tsx
```diff
@@ -21,6 +21,7 @@ import React, { useMemo } from 'react';
 import { treeItemClasses } from '@mui/x-tree-view/TreeItem';
 import { SimpleTreeView } from '@mui/x-tree-view/SimpleTreeView';
 import { renderTreeNode } from './utils';
+import { Box } from '@mui/material';
```
rart:
```suggestion
import Box from '@mui/material/Box';
```

rart (changes requested):

rart (approved):

PR #4923, opened 2026-07-15
[4.x][8221] Accesibility - snack duration, animations enabling/disabling

jvega190:
https://github.com/craftercms/craftercms/issues/8221


jvega190:

jvega190:
@coderabbitai about the minor nit, that's expected, the field changes to the default value on empty on focus out. So if somebody clears the input, after focusing out it'll be set to the default

rart (approved):

